PR #12, opened 2020-11-19
feat: 🎸 add SQFormCheckboxGroup and SQFormCheckboxGroupItem

lucas-homer:
SQFormCheckboxGroup and -Item comps enable an 'inclusion list' UX to
add/remove items from a larger list of options via checkbox, with an
optional 'select all' functionality built in.

[Loom demo](https://www.loom.com/share/021e1f7fed42443181b4285aa81cdf3f)

### DIFF
--- a/src/components/SQForm/SQFormCheckboxGroup.js
+++ b/src/components/SQForm/SQFormCheckboxGroup.js
@@ -17,6 +17,15 @@ export default function SQFormCheckboxGroup({
 }) {
   const {values, setFieldValue} = useSQFormContext();
 
+  const handleSelectAllChange = event => {
+    if (event.target.checked) {
+      setFieldValue(name, selectAllData);
+    } else {
+      setFieldValue(name, []);
+    }
+    setFieldValue('selectAll', !values.selectAll);
+  };
+
   if (!useSelectAll) {
     return <SQFieldArray name={name}>{children}</SQFieldArray>;
   }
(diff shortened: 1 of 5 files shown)


src/components/SQForm/SQFormCheckboxGroup.js
```diff
@@ -0,0 +1,56 @@
...
+          label="Select All"
+          isChecked={values.selectAll}
+          onChange={e => {
+            if (e.target.checked) {
+              setFieldValue(name, selectAllData);
+            } else {
+              setFieldValue(name, []);
+            }
+            setFieldValue('selectAll', !values.selectAll);
+          }}
```
20BBrown14, line 38: For readability sake could we pull this anon function out to a const?

src/components/SQForm/SQFormCheckboxGroup.js
```diff
@@ -0,0 +1,56 @@
...
+          {...selectAllProps}
+        />
+      </Grid>
+      <SQFieldArray name={name}>{children}</SQFieldArray>
+    </>
+  );
+}
+
+SQFormCheckboxGroup.propTypes = {
```
lucas-homer, line 47: forgot to add the other proptypes! standby for that...

src/components/SQForm/SQFormCheckboxGroup.js
lucas-homer: sure! good idea.

lucas-homer: done

SeanGroff (approved):
This looks solid to me!
As we start using this more in the wild we may notice common patterns we can abstract into this component.
Since we don't know those yet, I think the current solution of going full-on inversion is solid

20BBrown14 (approved):

PR #13, opened 2020-11-20
fix: 🐛 reset form state during onclose event

lucas-homer:
We need to reset the form state when we 'cancel' out of the dialog, so I moved most of the dialog into another custom component, enabling access to `resetForm` from `useSQFormContext`, and put that inside a handler. Additionally, I noticed a small issue when consuming this component re: the DialogActions styles. We weren't getting the `space-between` override we set in CSS, so I moved that to a `classes` object to directly change the MuiDialogActions-root class.

[Loom](https://www.loom.com/share/54bc957125b64fdf8bafbea15ac3f7bd)

### DIFF
--- a/src/components/SQFormDialog/SQFormDialog.js
+++ b/src/components/SQFormDialog/SQFormDialog.js
@@ -45,8 +45,6 @@ function SQFormDialog({
         onSave={onSave}
         saveButtonText={saveButtonText}
         title={title}
-        enableReinitialize={false}
-        initialValues={initialValues}
         muiGridProps={muiGridProps}
       />
     </Formik>
--- a/src/components/SQFormDialog/SQFormDialog.js
+++ b/src/components/SQFormDialog/SQFormDialog.js
@@ -1,21 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Dialog from '@material-ui/core/Dialog';
-import DialogActions from '@material-ui/core/DialogActions';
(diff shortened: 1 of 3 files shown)


src/components/SQFormDialog/SQFormDialog.js
```diff
@@ -47,43 +34,21 @@ function SQFormDialog({
...
-                {saveButtonText}
-              </SQFormButton>
-            )}
-          </DialogActions>
-        </Form>
-      </Dialog>
+        onSave={onSave}
+        saveButtonText={saveButtonText}
+        title={title}
+        enableReinitialize={false}
```
SeanGroff, line 48: Can we expose `enableReinitialize` as a prop to the consumer? 
This prop has been needed more and more in SQForm (It's exposed there)

SeanGroff (changes requested):
Outside of exposing [enableInitialize](https://formik.org/docs/api/formik#enablereinitialize-boolean) this looks good to me!

src/components/SQFormDialog/SQFormDialog.js
lucas-homer: oh I think this is my mistake... good catch

lucas-homer: yeah I messed up a couple props getting passed down to the inner dialog component that only Formik needs.

lucas-homer: done!

PR #14, opened 2020-11-23
feat: 🎸 SQForm MultiSelect

atash-sq:
A new multiselect field for SQForm, equipped with Select All capability

https://www.loom.com/share/f1c1a91da7a745acbf25c646ae35018c

✅ Closes: SSC-82

### DIFF
--- a/src/components/SQForm/SQFormMultiSelect.js
+++ b/src/components/SQForm/SQFormMultiSelect.js
@@ -8,6 +8,7 @@ import FormHelperText from '@material-ui/core/FormHelperText';
 import Grid from '@material-ui/core/Grid';
 import Checkbox from '@material-ui/core/Checkbox';
 import ListItemText from '@material-ui/core/ListItemText';
+import Tooltip from '@material-ui/core/Tooltip';
 import {useSQFormContext} from '../../../src';
 import {useForm} from './useForm';
 
@@ -36,9 +37,11 @@ function SQFormMultiSelect({
   label,
   name,
   size = 'auto',
-  useSelectAll = true
+  useSelectAll = true,
+  toolTipPlacement = 'bottom'
 }) {
   const {setFieldValue} = useSQFormContext();
+  const [toolTipEnabled, setToolTipEnabled] = React.useState(true);
   const {
     formikField: {field},
     fieldState: {isFieldError},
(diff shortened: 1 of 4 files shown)


src/components/SQForm/SQFormMultiSelect.js
```diff
@@ -0,0 +1,107 @@
...
+    name,
+    isRequired
+  });
+  const labelID = label.toLowerCase();
+
+  const handleMultiSelectChange = event => {
+    const selectAllWasChecked = event.target.value.includes('ALL');
+    const selectNoneWasChecked = event.target.value.includes('NONE');
+
+    const values = selectAllWasChecked
```
atash-sq, line 39: I'm all for a better approach as to how we handle what gets sent to the `setFieldValue` function -- but I do like the fact that I can just use the children prop passed in to generate the list of things to select when I hit "Select All"

lucas-homer: can we refactor this a bit to avoid the nested ternary? I like the idea, generally, but I think if we wrote a little helper function or two we could make it a bit more readable.

stories/SQForm.stories.js
```diff
@@ -333,6 +348,40 @@ export const formWithCheckboxGroup = () => {
   );
 };
 
+export const basicFormWithMultiSelect = () => {
```
lucas-homer, line 351: Any chance you could add a few more fields or something to make the card a little bigger? I say that only because of the way the options move up and down in the loom video. It's kind of janky, and I wonder if it's a product of showing it in storybook, or if it's behave that way in the wild. Is there any way to address that? I'd lay money that Nemeth will balk at that.

lucas-homer (commented):
just a few comments, but overall looks good!

src/components/SQForm/SQFormMultiSelect.js
atash-sq: Yeah if the nested ternary is a no-go, I guess I could just write that out in if statements instead? It's a simple enough conditional that I don't know if a helper function would do much other than abstract the logic away from the function.

stories/SQForm.stories.js
atash-sq: I wondered the exact same thing while poking around 💯 . I believe it's a problem with storybook tbh. I'll spend a few minutes poking. But, I could take the same code and place it in Life CRM and wasn't getting that jank?

src/components/SQForm/SQFormMultiSelect.js
lucas-homer: yeah that works for me

SeanGroff: I think a functional approach would alleviate the need for the nested logic

```javascript
const value = event.target.value;
const isSelectAllChecked = getIsSelectAllChecked(value); // return a boolean
const values = getValues(children, isSelectAllChecked, value) // return children.map(), [], or value

setFieldValue(name, values);
```

SeanGroff (changes requested):
This looks great!
I think once we flatten the logic we've discussed in the comments we can ship it 💯

lucas-homer (approved):

SeanGroff (approved):
Looks great nice work 🚀

PR #15, opened 2020-11-24
fix: 🐛 Fixed a list item label

atash-sq:
Fixed a list item in multi select to use the label instead of value

✅ Closes: SSC-82


src/components/SQForm/SQFormMultiSelect.js
```diff
@@ -127,7 +127,7 @@ function SQFormMultiSelect({
             return (
               <MenuItem key={option.value} value={option.value}>
                 <Checkbox checked={field.value.includes(option.value)} />
-                <ListItemText primary={option.value} />
+                <ListItemText primary={option.label} />
```
atash-sq, line 130: image

lucas-homer: heh. nice

lucas-homer (approved):

PR #24, opened 2020-12-16
SSC-105 update stepper styling and submit button logic

chris-boyle-selectquote:
Several UI updates are being requested in our first use case of the stepper dialog
* making the primary button dynamically show “Next” or “Submit” based on what step it is on.
* making the arrows at the breadcrumb level bigger
* larger step icons
* remove some of the padding top and bottom in the breadcrumb area
* static breadcrumbs section that is not affected by scrolling.
https://www.loom.com/share/fbcb610dd9cf406f83c5539220f7a966


lucas-homer (commented):
in SQFormDialogStepper.js -- isn't `children` going to be an array of things? Are they actually dom nodes? or something else? I know your proptypes aren't part of the diff, but in wrapping my head around all this, I noticed that was not what I expected it'd be

lucas-homer (commented):
couple questions about the demos in the loom -- the first example you ended up on the last step, the field appears unrequired, the submit button is enabled, and the "required" error validation only showed up after trying to click submit. That seems like a bug, no? Second, in the second example, you clicked the first step as if it were a button, but it didn't take you back to step 1. Is that the desired behavior?

src/components/SQFormDialogStepper/SQFormDialogStepper.js
```diff
@@ -58,8 +74,9 @@ export function SQFormDialogStepper({
   }, [activeStep, totalSteps]);
 
   // Our last step doesn't get marked complete
-  const allStepsCompleted = () =>
-    Object.keys(completed).length === totalSteps - 1;
+  const allStepsCompleted = () => {
```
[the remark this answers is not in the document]
chris-boyle-selectquote: `isAllStepsComplete` sound good?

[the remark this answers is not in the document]
lucas-homer: not quite -- https://selectquote.atlassian.net/wiki/spaces/SE/pages/796590204/Internal+Code+Standards#Naming

stories/SQFormDialogStepper.stories.js
```diff
@@ -25,77 +30,175 @@ const handleSubmit = (values, actions) => {
...
-          label="Account Info"
-          validationSchema={object({
-            accountID: mixed().when('newAccount', {
-              is: true,
-              then: number()
-                .required()
-                .min(100, 'Since this is a new account we need the number')
-            })
+          label="Dependents"
+          validationSchema={Yup.object({
```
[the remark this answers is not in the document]
chris-boyle-selectquote: Screen Shot 2020-12-16 at 10 06 17 AM
That is the Yup.object.  I can convert it to a normal object and then refactor the places the validation schema is used.

chris-boyle-selectquote:
> couple questions about the demos in the loom -- the first example you ended up on the last step, the field appears unrequired, the submit button is enabled, and the "required" error validation only showed up after trying to click submit. That seems like a bug, no? Second, in the second example, you clicked the first step as if it were a button, but it didn't take you back to step 1. Is that the desired behavior?

In the first example the submit button being enabled and then disabled once an error is thrown is what I brought up with the group yesterday we didn't come up with any ideas/solutions.  I don't love how it works but UX signed off on how it functions in that demo.
In the second demo I clicked the first step after the form was submitted.  That is why it was in a weird state where you can't click to a previous step.

chris-boyle-selectquote:
> in SQFormDialogStepper.js -- isn't `children` going to be an array of things? Are they actually dom nodes? or something else? I know your proptypes aren't part of the diff, but in wrapping my head around all this, I noticed that was not what I expected it'd be

@lucas-homer it is an array of react elements.

stories/SQFormDialogStepper.stories.js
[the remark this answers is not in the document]
chris-boyle-selectquote: Using a plain JS Object isn't working or recommended in the [Formik docs](https://formik.org/docs/guides/validation#validationschema).  I am going to keep it as a Yup.object unless there is some objection or reason not to.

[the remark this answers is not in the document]
lucas-homer: ok cool! I think this should be added to our SQForm documentation because I've definitely not been doing it this way, but as you point out the Formik docs are clear on it

lucas-homer:
> > couple questions about the demos in the loom -- the first example you ended up on the last step, the field appears unrequired, the submit button is enabled, and the "required" error validation only showed up after trying to click submit. That seems like a bug, no? Second, in the second example, you clicked the first step as if it were a button, but it didn't take you back to step 1. Is that the desired behavior?
> 
> In the first example the submit button being enabled and then disabled once an error is thrown is what I brought up with the group yesterday we didn't come up with any ideas/solutions. I don't love how it works but UX signed off on how it functions in that demo.
> In the second demo I clicked the first step after the form was submitted. That is why it was in a weird state where you can't click to a previous step. It is also not a real scenario where the form would be open after submitting.

Could you make an Issue for that detail? Worth further investigation, but I don't want to hold this PR back if you're trying to consume the update sooner rather than later.

lucas-homer:
> > in SQFormDialogStepper.js -- isn't `children` going to be an array of things? Are they actually dom nodes? or something else? I know your proptypes aren't part of the diff, but in wrapping my head around all this, I noticed that was not what I expected it'd be
> 
> @lucas-homer it is an array of react elements.

could we update that in this PR and sneak in extra value with a small fix 😬

lucas-homer (approved):
Could we copy/paste over the docs from SSC for this component? I'm realizing we don't have any docs setup but we're already consuming this library, death by a thousand cuts, etc. I don't want to hold you up with this, but it would be great to get a little documentation going in here for this fairly complex component api

src/components/SQFormDialogStepper/SQFormDialogStepper.js
```diff
@@ -121,6 +140,8 @@ export function SQFormDialogStepper({
 
     return (
       <IconButton
+        height={'80px'}
```
20BBrown14, line 143: minor: can just do `height="80px"` and `width="80px"` applies throughout

src/components/SQFormDialogStepper/SQFormDialogStepper.js
```diff
@@ -184,35 +205,41 @@ export function SQFormDialogStepper({
...
+                    </Step>
+                  ))}
+                </Stepper>
                 <NextButton />
-              </div>
-              <Divider className="SQFormDialogStepper__divider" />
+              </Grid>
+            </div>
+            <DialogContent
+              dividers={true}
```
20BBrown14, line 236: Can just do shorthand `dividers`

stories/SQFormDialogStepper.stories.js
```diff
@@ -1,15 +1,20 @@
 import React from 'react';
-import {mixed, number, object, string} from 'yup';
+import * as Yup from 'yup';
+import Grid from '@material-ui/core/Grid';
 import {withKnobs, boolean} from '@storybook/addon-knobs';
 import {withInfo} from '@storybook/addon-info';
 import {action} from '@storybook/addon-actions';
+import {Typography} from '@material-ui/core';
```
20BBrown14, line 7: super nit: could we put all MUI imports together? And could we decide if we going to import from `/core` or not.
```
import Grid from '@material-ui/core/grid;
import {Typography} from `@material-ui/core`;
```

src/components/SQFormDialogStepper/SQFormDialogStepper.js
```diff
@@ -184,35 +205,41 @@ export function SQFormDialogStepper({
...
-                        </StepButton>
-                      </Step>
-                    ))}
-                  </Stepper>
-                </div>
+                <Stepper nonLinear activeStep={activeStep} classes={classes}>
+                  {steps.map((child, index) => (
+                    <Step key={child.props.label}>
+                      <StepButton
+                        onClick={handleStep(index)}
```
SeanGroff, line 224: This function is going to be executed on render right?
If i'm reading the code right this should be `onClick={() => handleStep(index)}`

SeanGroff (approved):
I have one question but other than that I approve.
@chris-boyle-selectquote  Can you ensure you are the dev that handles the initial consumption of this shared component in SQAH so the kinks can be ironed out and the other devs will have a real example to reference going forward?

src/components/SQFormDialogStepper/SQFormDialogStepper.js
chris-boyle-selectquote: @SeanGroff No it only fires when you click on one of the steps in the breadcrumb.  I just tested it with some console logs.

chris-boyle-selectquote:
> I have one question but other than that I approve.
> @chris-boyle-selectquote Can you ensure you are the dev that handles the initial consumption of this shared component in SQAH so the kinks can be ironed out and the other devs will have a real example to reference going forward?

That is the plan my man!

atash-sq (approved):

src/components/SQFormDialogStepper/SQFormDialogStepper.js
20BBrown14: @SeanGroff I thought the same thing, handleStep returns a function

SeanGroff: > @SeanGroff I thought the same thing, handleStep returns a function

Ohhhhhh ok, so we are partially applying the function with the index. That makes sense then!

chris-boyle-selectquote:
Going to go ahead and merge this and work on getting additional docs created in  a separate ticket.

PR #28, opened 2020-12-18
SSC-105 Updated styles from css file, simplified booleans

chris-boyle-selectquote:
When consuming this component none of the css file was packaged up into it.  I have changed to use makeStyles to have the css applied to the component. 

Also removed the stepper component if we only have one step.


SeanGroff (approved):
LGTM!
Bundling CSS is something I want to get away from when shipping library code.

Issue #38, opened 2020-12-23
SQFormAutoComplete resets form state on blur


chris-boyle-sq:
After selecting a value in the autocomplete field the rest of the fields that were validated are reset.  This causes required fields that were filled in to now show as needing to be touched again to be valid.   This is affecting all consumers of the SQFormAutoComplete component (dialog forms and forms). 

Steps to reproduce
Update the SQForm basic form example to have `isRequired={true}` on every field. 
Fill out the fields and when you get to the autocomplete field take not of all the validated fields with the green check mark.  
After leaving the auto complete field all fields that were validated will have their state reset to their initial state value.

Loom: https://www.loom.com/share/d7f51ceb423b45ed802851ca693f64af

chris-boyle-sq:
I am going to start digging into this since it is currently blocking me.

SeanGroff:
I've noticed this as well. Certainly a bug.

PR #42, opened 2020-12-29
fix: 🐛 Fixed onBlur event for autocomplete and touched fields SQAH-1268

chris-boyle-sq:
✅ Closes: #38

Spreading the touched fields with the autocomplete touched field so that we don't overwrite the existing touched fields.

Loom shows the existing bug and fix.
Loom: https://www.loom.com/share/38621828a71246e9889085446241b58a


atash-sq (approved):

SeanGroff (approved):
Thanks for the fix!

Issue #43, opened 2020-12-30
SQFormAutocomplete should be able to have a default value


20BBrown14:
Currently it doesn't seem like SQFormAutocomplete fields allow for a default value. Like other types of inputs, this input should use the initial values passed to the form, via formik, and show a default value. If no value is provided or selected a `- -` should be shown instead.

chris-boyle-sq:
I have some time to pick this up.

PR #46, opened 2020-12-31
fix: 🐛 Allow default value to be used in autocomplete component SSC-106

chris-boyle-sq:
✅ Closes: #43

We can now pass a default value to the autocomplete component.  It also sets the touched property to true so the field shows as valid and not still required.

Loom: https://www.loom.com/share/6c16a4ca15364c6f9871b953d0a0a994

### DIFF
--- a/src/components/SQForm/SQFormAutocomplete.js
+++ b/src/components/SQForm/SQFormAutocomplete.js
@@ -105,9 +105,6 @@ function SQFormAutocomplete({
 
   const initialValue = children.find(option => {
     if (option.value === value) {
-      if (!touched.hasOwnProperty(name)) {
-        setTouched({...touched, ...{[name]: true}});
-      }
       return option;
     }
     return null;
--- a/src/components/SQForm/SQFormAutocomplete.js
+++ b/src/components/SQForm/SQFormAutocomplete.js
@@ -102,6 +102,17 @@ function SQFormAutocomplete({
     name,
     isRequired
   });
+
+  const initialValue = children.find(option => {
+    if (option.value === value) {
+      if (!touched.hasOwnProperty(name)) {
+        setTouched({...touched, ...{[name]: true}});
+      }
+      return option;
+    }
+    return null;
+  });
+
   const [inputValue, setInputValue] = React.useState('');
   const prevValue = usePrevious(value);
 
@@ -150,6 +161,7 @@ function SQFormAutocomplete({
         onChange={handleAutocompleteChange}
         onInputChange={handleInputChange}
         inputValue={inputValue}
+        value={initialValue}
         getOptionLabel={option => option.label}
         renderInput={params => {
           return (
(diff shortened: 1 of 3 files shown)


lucas-homer (approved):
lgtm! thanks for taking care of this

20BBrown14 (approved):
🔥

SeanGroff (commented):
Making the field dirty when it hasn't actually been touched to make the required field fulfilled is going to make the form submit button enabled even though no real form changes have been made. See the useFormButton.js file. 

I think we might have to address the required fulfillment in the useSQForm.js file instead.

SeanGroff:
Instead of marking the field as touched can we try changing this line in useForm.js

```javascript
 const isFieldRequired = !isTouched && isRequired;
```

to

```javascript
 const isFieldRequired = !isTouched && isRequired && !value;
```

Sorry I'm on mobile or I'd be less lazy and give this a try myself.

chris-boyle-sq:
@SeanGroff that will work and not show the Required message but also not set the field as touched so the Submit button is not enabled.  
image

SeanGroff:
Admittedly this is tricky to get right. 

To me, this should be the expected behavior. 

Do we have a prod example where we need to populate an auto complete field with a default value then submit that to the database?

The general pattern is default to - - unless this field was previously submitted, then default it to the previously submitted value. 
 
If we populate an auto complete field with a value. That means that is the value previously submitted. Since there's no new changes to submit,the enable button should be disabled because there's no new changes to submit.

SeanGroff:
To be clear, we don't want to lie to the form and mark a field as touched when the user hasn't touched it. 

This issue IMO is a separate issue not specific to just the autocomplete field where we need to change how we are calculating the required, error and fulfilled states in useForm.js. 

Without deeply testing this, my initial assumption is to use `value` instead of `touched` to calculate these values.

I thought there's an issue filed for this but I need to double check. I can focus on this Monday to unblock this PR. 

I think this PR is solid if we remove the logic to mark the field as touched and handle the fulfilled state in a separate issue.

chris-boyle-sq:
> Admittedly this is tricky to get right.
> 
> To me, this should be the expected behavior.
> 
> Do we have a prod example where we need to populate an auto complete field with a default value then submit that to the database?
> 
> The general pattern is default to - - unless this field was previously submitted, then default it to the previously submitted value.
> 
> If we populate an auto complete field with a value. That means that is the value previously submitted. Since there's no new changes to submit,the enable button should be disabled because there's no new changes to submit.

Yes I agree with you.  The scenario we ran into is when we Pull up a form that was previously submitted.  The form is an "Edit" form but I think with the changes you suggested we get the right state for this autocomplete component.  I believe this is good to go now.  We can talk more on Monday if you want.

SeanGroff (approved):
All work to fix the bug for this issue is fulfilled.
The behavior of the isRequired indicator and submit button is outside the scope of this PR and will be addressed in issue #47

Issue #45, opened 2020-12-30
SQFormTextField and SQFormTextArea have optional type Prop


20BBrown14:
Curious on what people's thoughts are on having a `type` prop for [SQFormTextField](https://github.com/SelectQuoteLabs/SQForm/blob/master/src/components/SQForm/SQFormTextField.js#L50) and [SQFormTextArea](https://github.com/SelectQuoteLabs/SQForm/blob/master/src/components/SQForm/SQFormTextarea.js) so the consumer can define their own type on the [MUI TextField](https://material-ui.com/api/text-field/).

Valid html types: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Form_%3Cinput%3E_types

I'm thinking that mainly this would be good so we can provide a `number` type and get some of the HTML5 validation stuff. We could limit it to certain types if we wanted and of course we could default to Text.

PR #49, opened 2021-01-04
feat: 🎸 Allow text fields HTML5 types

20BBrown14:
Allow a HTML5 types to be supplied to SQForm text fields

You can now pass a string for prop `type` to give the the input under SQFormTextField a HTML5 input type

✅ Closes: #45

Loom: https://www.loom.com/share/cc3f1322e3e44ea385f1b978f1bcc935

### DIFF
--- a/src/components/SQForm/SQFormTextField.js
+++ b/src/components/SQForm/SQFormTextField.js
@@ -16,7 +16,8 @@ function SQFormTextField({
   onBlur,
   onChange,
   startAdornment,
-  endAdornment
+  endAdornment,
+  type = 'text'
 }) {
   const {
     formikField: {field},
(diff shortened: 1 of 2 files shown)


SeanGroff (changes requested):
Love it! Can you remove the `package-lock.json` file from this PR since no changes occurred to the package.json dependencies?

lucas-homer (approved):

20BBrown14:
> Love it! Can you remove the `package-lock.json` file from this PR since no changes occurred to the package.json dependencies?

Done and dusted

SeanGroff (approved):

chris-boyle-sq:
We should really advertise this one better in the documentation.  Otherwise people won't know about it and they will create some custom schema validation tests like I did.  I think this is the better way though!

Issue #50, opened 2021-01-04
Create a .nvmrc to specify the Node version for this project


SeanGroff:
The version should be version 14.15.4 (Node LTS)

SeanGroff:
I'll grab this and knock it out real quick

PR #51, opened 2021-01-04
chore: 🤖 Specifies a Node version using a .nvmrc file

SeanGroff:
✅ Closes: #50

Since we aren't using Docker to control the Node version we now use the next best thing. NVM

Updated the Readme to include Local Dev initial setup steps.


chris-boyle-sq (approved):

.nvmrc
```diff
@@ -0,0 +1 @@
+14.15.4
```
20BBrown14, line 1: Should we do `lts/fermium` to ensure we're getting the latest versions?

https://nodejs.org/en/about/releases/

SeanGroff: We should!

20BBrown14 (approved):

PR #55, opened 2021-01-06
feat: 🎸 allow passthrough of props to mui field children

caleboleary:
✅ Closes: 54

I omitted buttons and non-mui field types. My immediate use case for this will be autofocus on a text input in modal.

### DIFF
--- a/src/components/SQForm/SQFormCheckbox.js
+++ b/src/components/SQForm/SQFormCheckbox.js
@@ -11,7 +11,8 @@ function SQFormCheckbox({
   label,
   name,
   onChange,
-  size = 'auto'
+  size = 'auto',
+  muiFieldProps = {}
 }) {
   const {
     formikField: {field},
@@ -32,6 +33,7 @@ function SQFormCheckbox({
             disabled={isDisabled}
             name={name}
             onChange={handleChange}
+            {...muiFieldProps}
           />
         }
         label={label}
@@ -50,7 +52,9 @@ SQFormCheckbox.propTypes = {
   /** Custom onChange event callback */
   onChange: PropTypes.func,
   /** Size of the input given full-width is 12. */
-  size: PropTypes.oneOf(['auto', 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12])
+  size: PropTypes.oneOf(['auto', 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
+  /** Any valid prop for material ui child component - https://material-ui.com/  */
+  muiFieldProps: PropTypes.object
 };
 
 export default SQFormCheckbox;
--- a/src/components/SQForm/SQFormCheckbox.js
+++ b/src/components/SQForm/SQFormCheckbox.js
@@ -53,7 +53,7 @@ SQFormCheckbox.propTypes = {
   onChange: PropTypes.func,
   /** Size of the input given full-width is 12. */
   size: PropTypes.oneOf(['auto', 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
-  /** Any valid prop for material ui child component - https://material-ui.com/  */
+  /** Any valid prop for material ui checkbox child component - https://material-ui.com/api/checkbox/#props */
   muiFieldProps: PropTypes.object
 };
 
--- a/src/components/SQForm/SQFormDatePicker.js
+++ b/src/components/SQForm/SQFormDatePicker.js
@@ -24,7 +24,8 @@ function SQFormDatePicker({
   isRequired = false,
   placeholder = '',
   onBlur,
-  onChange
+  onChange,
+  muiFieldProps = {}
 }) {
   const {
     formikField: {field, helpers},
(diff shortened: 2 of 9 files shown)


SeanGroff:
@caleboleary Can you edit one of the form stories in Storybook to show autofocus working?
That will not only prove the new prop `muiFieldProps` works but also give a great example to devs who need an example of what should be a common UI need.

Code itself LGTM! 🚀

Chris-Boyle:
With using a MUI specific prop like this we are increasing the cognitive load for consumers and now expecting them to know all about muiFieldProps.  The alternative would be just adding a simple autoFocus prop that almost needs no explanation.  We want to keep our APIs as simple as possible to make them fail proof. I just wonder if we are doing that here.  The link in the proptypes should also be more specific.

caleboleary:
> With using a MUI specific prop like this we are increasing the cognitive load for consumers and now expecting them to know all about muiFieldProps. The alternative would be just adding a simple autoFocus prop that almost needs no explanation. We want to keep our APIs as simple as possible to make them fail proof and easy to quickly consume. I just wonder if we are doing that here. The link in the proptypes should also be more specific.

I think an autofocus option would be good too, but as a consumer of this lib, I also want more control when I need it, and would prefer not to have to do another PR if I find I need another MUI prop. 

What about exposing both `autofocus` and `muiFieldProps`? That caters to both kinds of consumers (those who want simple and those who want more control).

Also, yeah I can add more relevant links to the different field types per file.

Chris-Boyle:
@caleboleary  If we can add documentation and some examples of using the muiFieldProps that may be best.  I am OK with just using muiFieldProps and not autoFocus too.  I just don't think the idea of this library is to have people dig through documentation in another library.  We should have the necessary documentation for the components in a concise manner to make consumption quick and painless.

caleboleary:
@Chris-Boyle 

Sounds good to me. I'll update the proptype comments to go to relevant component mui pages, and will add a storybook story that shows off autofocus. Update coming shortly.

SeanGroff:
> With using a MUI specific prop like this we are increasing the cognitive load for consumers and now expecting them to know all about muiFieldProps. The alternative would be just adding a simple autoFocus prop that almost needs no explanation. We want to keep our APIs as simple as possible to make them fail proof and easy to quickly consume. I just wonder if we are doing that here. The link in the proptypes should also be more specific.

I'm glad this was mentioned. 
This is the very reason `{...props}` should be avoided at all costs. It provides zero intent of what it and how the component should be used.

What we have to balance in our component props API is API simplicity vs total functionality.
I believe our current props API for let's say the Textfield strikes a good balance of declaring the props for the 80% use case of that component. Maybe the need never comes up for the other 20%, which is why we shouldn't preemptively add a prop like `muiFieldProps` initially. 

In most (probably all but won't go that far) of our apps we don't use the autoFocus property or call for it to be used in the UI. 

Since autoFocus then becomes part of the 20% (the rarely or much less used props) we don't want to add another prop to our props API but we do want to provide an escape hatch for the less commonly used functionality. 

By including `muiFieldProps` instead of `{...props}` we are at least lowering the scope of what we do with this prop to the documented props from MUI.

We could go a step further and specify the accepted props in our prop-type but then we have to manage this with any prop additions to the MUI library.

caleboleary:
https://www.loom.com/share/78a6ccd7e0f74b11a70b27d66b50b085 autofocus story loom

SeanGroff:
> > With using a MUI specific prop like this we are increasing the cognitive load for consumers and now expecting them to know all about muiFieldProps. The alternative would be just adding a simple autoFocus prop that almost needs no explanation. We want to keep our APIs as simple as possible to make them fail proof and easy to quickly consume. I just wonder if we are doing that here. The link in the proptypes should also be more specific.
> 
> I'm glad this was mentioned.
> This is the very reason `{...props}` should be avoided at all costs. It provides zero intent of what it and how the component should be used.
> 
> What we have to balance in our component props API is API simplicity vs total functionality.
> I believe our current props API for let's say the Textfield strikes a good balance of declaring the props for the 80% use case of that component. Maybe the need never comes up for the other 20%, which is why we shouldn't preemptively add a prop like `muiFieldProps` initially.
> 
> In most (probably all but won't go that far) of our apps we don't use the autoFocus property or call for it to be used in the UI.
> 
> Since autoFocus then becomes part of the 20% (the rarely or much less used props) we don't want to add another prop to our props API but we do want to provide an escape hatch for the less commonly used functionality.
> 
> By including `muiFieldProps` instead of `{...props}` we are at least lowering the scope of what we do with this prop to the documented props from MUI.
> 
> We could go a step further and specify the accepted props in our prop-type but then we have to manage this with any prop additions to the MUI library.

One more point to make. If we become better about utilizing autoFocus in our UI's and it becomes part of the 80% commonly used props, we should add autoFocus as its own prop at that time. (While still keeping the escape hatch)

SeanGroff (approved):
Thanks for going to extra mile to do this across all components @caleboleary 🥇

Chris-Boyle (approved):

SeanGroff (approved):

Issue #53, opened 2021-01-04
update multi select to accept onChange handler


ghost:
Currently most of our components have an onChange prop that can be passed but the multiselect component does not.  We would like to have an onChange handler so we can conditionally show/hide other fields in the form based on what options are selected in the multi-select component.

Chris-Boyle:
I can start looking at this.

PR #62, opened 2021-01-11
feat: 🎸 Added onChange handler to multiselect component SSC-107

Chris-Boyle:
✅ Closes: #53

Added ability for consumer to pass onChange method.

Loom: https://www.loom.com/share/60d6eebb6f8a485395a9d0f66d858ac0

### DIFF
--- a/src/components/SQForm/SQFormMultiSelect.js
+++ b/src/components/SQForm/SQFormMultiSelect.js
@@ -60,10 +60,9 @@ function SQFormMultiSelect({
   const {
     formikField: {field},
     fieldState: {isFieldError},
-    fieldHelpers: {handleBlur, handleChange, HelperTextComponent}
+    fieldHelpers: {handleBlur, HelperTextComponent}
   } = useForm({
     name,
-    onChange,
     isRequired
   });
 
@@ -102,6 +101,7 @@ function SQFormMultiSelect({
     );
 
     setFieldValue(name, values);
+    onChange && onChange(value);
   };
 
   const toggleTooltip = () => {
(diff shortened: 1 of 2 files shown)


src/components/SQForm/SQFormMultiSelect.js
```diff
@@ -137,7 +139,7 @@ function SQFormMultiSelect({
           input={<Input disabled={isDisabled} name={name} />}
           value={field.value}
           onBlur={handleBlur}
-          onChange={handleMultiSelectChange}
+          onChange={handleMultiSelectChange && handleChange}
```
20BBrown14, line 142: With this change `handleMultiSelectChange` will never get called, right? Maybe I'm missing something.

SeanGroff: If we add a custom `onChange` prop to this component we want the original `handleMultiSelectChange` method to be called. Then if an `onChange` method is passed to the component, call it after the `handleMultiSelectChange` method has executed.

This component is imperatively calling setFieldValue already so I believe we wouldn't use the `handleChange` help method like we would in fields that are provided a single value.

I think what we need to do here is after `setFieldValue(name, values);` is called, add a new line with `onChange && onChange(value);`

Chris-Boyle: Good call.  Overlooked that by trying to use the useForm hook.  @SeanGroff and @20BBrown14 I have updated the code.
image

20BBrown14 (approved):

SeanGroff (approved):
LGTM nice work @Chris-Boyle 🤘

Issue #65, opened 2021-01-12
SQForm Autocomplete issue - add onInputChange prop


umerjamal100:
We need to use `onChange` for `SQFormAutocomplete` to get SmartyStreets addresses but it is not fired when we type or update, it is being fired when we select the option, while it is working fine for `SQFormTextField`.
The current functionality is that we have to select an option to trigger that. 

Suggestion:
We need to implement `onInputChange` in our component, currently, we have `onChange` implemented only.

Please refer to the attached video for further details.
[loom video related to issue](https://www.loom.com/share/ab1dd9b80628461098b6aba5dd70372b)

Chris-Boyle:
I can start looking at this.

PR #68, opened 2021-01-12
feat: 🎸 Added onInputChange prop to autocomplete component SSC-111

Chris-Boyle:
✅ Closes: #65

Added onInputProp to allow users to update local state when each key is pressed.  The use case for this is our address fields.  If we start typing an address we send an API request with the values to Smarty Streets and dynamically update the options in the dropdown.

LOOM: https://www.loom.com/share/734652bc9cd141349b2497c3df2a7a91


20BBrown14 (approved):

SeanGroff (approved):
Neat! I believe these are the changes needed in order to use the Google Places API as well in the future 🚀

Issue #70, opened 2021-01-13
SQFormDialogStepper, SQFormDialog add loading state the dialogs


Chris-Boyle:
In our dialogs we are not currently handling loading state.  There are times when we open the dialog and are not ready to render the components.  In this scenario we should show a loading spinner until the consumer says it is not loading anymore.  Currently a consumer can do this but it would make consumer code much cleaner to have the dialogs handle it.

I am proposing we show the Select Quote circles as a loading spinner.

20BBrown14:
> I am proposing we show the Select Quote circles as a loading spinner.

I suggest we allow the consumer to pass an optional loading component as well

Chris-Boyle:
> > I am proposing we show the Select Quote circles as a loading spinner.
> 
> I suggest we allow the consumer to pass an optional loading component as well

I think we should probably solve for the 80% and just use our SelectQuote loading spinner for now until there is a request to allow consumers to define one.

SeanGroff:
I can see this for the Stepper as it's more of an all in one solution.
I don't believe loading states are a concern of Dialogs though.

I think the first problem to solve is building a shared component that will take up 100% of the parent elements height and width. Then center the loading rings and below the rings an optional loading message. Then this styled loading component can be used anywhere without having to do something like

```jsx
<Dialog>
  {isLoading ? <div style={loadingStyle}><LoadingSpinner /></div> : content}
</Dialog>
```

instead it would be:

```jsx
{isLoading ? <LoadingSpinner /> : content}
```

Styling is already done for the consumer.

PR #71, opened 2021-01-13
feat: 🎸 Adding loading state to dialogs

Chris-Boyle:
✅ Closes: #70

Adds ability for consumers to set isLoading to have a loading spinner shown instead of children components.  This allows data fetching to finish before we try to render the components to the form.

Loom: https://www.loom.com/share/0cc6f344e23d4f0a93aaa1e42cd99527


SeanGroff (changes requested):
We don't want consumers to have to set up a CSS bundler just to use SQForm.
We'd have a new issue of SQForm not working with libraries like Next.js.

The changes I'm requesting are
1) Convert the CSS to CSS-in-JS using Material-UI useStyles hook. Examples can be found in some of the other SQForm components.
2) The `isLoading` prop for the Dialogs should default to `false`. We don't want to rely on `undefined` to not show a loading spinner.

Chris-Boyle:
> We don't want consumers to have to set up a CSS bundler just to use SQForm.
> We'd have a new issue of SQForm not working with libraries like Next.js.
> 
> The changes I'm requesting are
> 
> 1. Convert the CSS to CSS-in-JS using Material-UI useStyles hook. Examples can be found in some of the other SQForm components.
> 2. The `isLoading` prop for the Dialogs should default to `false`. We don't want to rely on `undefined` to not show a loading spinner.

Addressed the concerns you had @SeanGroff .

Chris-Boyle:
Currently our loading state is passed at the parent component level SQFormDialogStepper.  I wonder if it would make more sense to pass it in the child SQFormDialogStep.  If not then we would just have to keep pushing the isLoading state up to the parent.  I think that may be fine.  If we run into complexities doing that in the future we can look at revisiting where the loading state lives.

SeanGroff:
Please update this branch with the master branch or else a release won't occur

Chris-Boyle:
> Please update this branch with the master branch or else a release won't occur

Done!

Chris-Boyle:
@SeanGroff got sign off from Chris N. 
image

Chris-Boyle:
New loom with updated loading @SeanGroff 
https://www.loom.com/share/cbe571d55429434885c9d997d6ad2089

SeanGroff (approved):
LGTM nice work @Chris-Boyle 🚀

Issue #72, opened 2021-01-14
SQFormAutocomplete missing event in onInputChange


umerjamal100:
We needed to add a handler in SQFormAutocomplete which will trigger on every keyPress, which was added but it is missing the event to return the entered value.

![image](https://user-images.githubusercontent.com/20329437/104609364-9ad1b200-5648-11eb-96e6-4f3ab1bdddac.png)
Shows that the event details is undefined when we try to access them
![image](https://user-images.githubusercontent.com/20329437/104609422-aa50fb00-5648-11eb-8f70-84d8c1edd7fb.png)

SeanGroff:
Nice catch @umerjamal100 

https://github.com/SelectQuoteLabs/SQForm/blob/521a843070e75ec284d6bca57276e0c1ebe473d2/src/components/SQForm/SQFormAutocomplete.js#L147

`onInputChange` should be passed the event

PR #73, opened 2021-01-14
Issue-72-SQFormAutocomplete-missing-event-in-onInputChange

umerjamal100:
Closes: [#72 ](https://github.com/SelectQuoteLabs/SQForm/issues/72)

We needed to add a handler in `SQFormAutocomplete` which will trigger on every keyPress, which was added but it is missing the event to return the entered value.
So this PR fixes that issue.

[https://www.loom.com/share/97302a634cd845d39674958f4efd82ed](https://www.loom.com/share/97302a634cd845d39674958f4efd82ed)

### DIFF
--- a/src/components/SQForm/SQFormAutocomplete.js
+++ b/src/components/SQForm/SQFormAutocomplete.js
@@ -144,7 +144,7 @@ function SQFormAutocomplete({
   const handleInputChange = React.useCallback(
     (_event, value) => {
       setInputValue(value);
-      onInputChange && onInputChange(_event, value);
+      onInputChange && onInputChange(value);
     },
     [onInputChange]
   );
--- a/src/components/SQForm/SQFormAutocomplete.js
+++ b/src/components/SQForm/SQFormAutocomplete.js
@@ -144,7 +144,7 @@ function SQFormAutocomplete({
   const handleInputChange = React.useCallback(
     (_event, value) => {
       setInputValue(value);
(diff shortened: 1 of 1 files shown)


src/components/SQForm/SQFormAutocomplete.js
```diff
@@ -144,7 +144,7 @@ function SQFormAutocomplete({
   const handleInputChange = React.useCallback(
     (_event, value) => {
       setInputValue(value);
-      onInputChange && onInputChange();
+      onInputChange && onInputChange(_event, value);
```
Chris-Boyle, line 147: I don't think we need the `_event` to be passed.  It looks like the value is all a consumer would need.  
image

SeanGroff: > I don't think we need the `_event` to be passed. It looks like the value is all a consumer would need.
> image

Generally, the value is what we need.
However, for any generic event handlers, the expected API is for the first argument to be the `event`

The 2nd argument `value` is  a nicety from `formik` since the most common use from an event handler is to use `event.target.value` to Chris's point.

Our component `onChange` and `onClick` event handlers should include the `event` as the first argument.

In the case we should include the value as the second argument to maintain the same API

Chris-Boyle (approved):

SeanGroff (approved):
Thanks for the contribution @umerjamal100 💯

Issue #75, opened 2021-01-15
SQFormDialogStepper Next button logic


Chris-Boyle:
We need to update the next button logic so that it is disabled until all required fields in the current step are filled out.

Currently it is enabled when the step is initially entered because we don't have any errors in the step yet.  Once you leave a required field or you click Next the validation fires and disables the Next Button and shows you the errors.

LOOM> https://www.loom.com/share/d00d1b2bf98d4fc39ece54366fffa124

PR #76, opened 2021-01-15
fix: 🐛 Next button logic to look at current step values to enabled/disabled SSC-115

Chris-Boyle:
✅ Closes: #75

Updated the next button enabled/disabled logic.  It now only looks at the current steps values to see if any have been set.  Before it was looking at all values from all steps.  

The touched property is still being set on all fields when the onSubmit is fired in the form.
If we have initial values then the next button will be enabled.

Loom https://www.loom.com/share/0def27c6e270412db50018b750deab0a

### DIFF
--- a/src/components/SQFormDialogStepper/SQFormDialogStepper.js
+++ b/src/components/SQFormDialogStepper/SQFormDialogStepper.js
@@ -152,10 +152,12 @@ export function SQFormDialogStepper({
         return false;
       }
       const currentStepKeys = Object.keys(validationSchema.fields);
-      const formValues = Object.values(values).filter(val => val);
+      const stepValues = currentStepKeys.some(step => {
+        return !!values[step];
+      });
 
       if (
-        !formValues.length ||
+        !stepValues ||
         currentStepKeys.some(step => Object.keys(errors).includes(step)) ||
         !dirty
       ) {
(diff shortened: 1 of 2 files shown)


SeanGroff (approved):
Awesome. The last point you made in the Loom video about the fields being touched in Formik, is an existing issue. I've got the fixes locally, I just need to spend more time testing edge cases.

Coming Soon ™️

Chris-Boyle:
> spend

As soon as the next button is clicked the onSubmit fires and somewhere down in Formik land the state is set to touched for all fields.  It would also be nice to reset the dirty state when the Next button is clicked.  The only thing I have found so far is using the resetForm function from useFormikContext.

Chris-Boyle:
> Awesome. The last point you made in the Loom video about the fields being touched in Formik, is an existing issue. I've got the fixes locally, I just need to spend more time testing edge cases.
> 
> Coming Soon ™️

As soon as the next button is clicked the onSubmit fires and somewhere down in Formik land the state is set to touched for all fields.  It would also be nice to reset the dirty state when the Next button is clicked.  The only thing I have found so far is using the resetForm function from useFormikContext.

SeanGroff:
> > Awesome. The last point you made in the Loom video about the fields being touched in Formik, is an existing issue. I've got the fixes locally, I just need to spend more time testing edge cases.
> 
> > 
> 
> > Coming Soon ™️
> 
> 
> 
> As soon as the next button is clicked the onSubmit fires and somewhere down in Formik land the state is set to touched for all fields.  It would also be nice to reset the dirty state when the Next button is clicked.  The only thing I have found so far is using the resetForm function from useFormikContext.

Setting the nextState with resetForm might be what you're looking for. Take a peak. https://formik.org/docs/api/formik#resetform-nextstate-partialformikstatevalues--void

Issue #74, opened 2021-01-15
DatePicker Enhancements


SeanGroff:
We need to be able to disable the text input field on the `DatePicker` component in scenarios where the proper UX is to force the user to select a date from the calendar picker.

* Add a boolean `allowInputFromCalendarOnly` that disables text input and makes the entire input open the calendar picker on focus
* We need to be able to disable certain days within the Calendar Picker by exposing a new prop `shouldDisableDate` https://next.material-ui-pickers.dev/api/DatePicker

Per Chris Nemeth
> The days in the calendar should have consistent styles with the existing Calendars in the SCPlus CRM's

* Selectable days should have a rounded Teal border
* Disabled days should maintain current disabled styles
* Currently selected day should be filled in Teal
* Current day should have a rounded Orange border

SeanGroff:
Beginning work on this now to unblock a feature request on the Senior CRM

SeanGroff:
Rather than add complexity to the SQFormDatePicker component I'm composing the functionality needed for this use case and SQFormDatePicker together to make a new component.

Also need to add the ability to clear the date selection since the user loses the ability to do this by not having a text input to interact with.

SeanGroff:
> Date/Time pickers are quite simple controls from UX perspective, so most people just use the default appearance. That's why we are not providing any for-component classes API to override stylesheets for any particular component. **The only way to override existing stylesheets is with the use of global material-ui theme overrides**.

This is from the MUI Pickers docs... 👎

We manage our global MUI theme in the SSC library. I'll have to handle the Calendar Day styling changes there.

PR #80, opened 2021-01-19
feat: 🎸 Component to force user date selection via DatePicker

SeanGroff:
The SQFormDatePickerWithCalendarInputOnly component disables the text
input forcing the user to make a date selection via the picker. Also
allows the consumer of SQFormDatePicker the ability to disable any(s) in
the calendar view. See [Issue 74](https://github.com/SelectQuoteLabs/SQForm/issues/74) for more information.

[Loom](https://www.loom.com/share/a4aeb6ff51bb46b4ac8dc1e306ec3060)

The issue also unblocks AZT-1451 on the Senior CRM

✅ Closes: #74

### DIFF
--- a/src/components/SQForm/SQFormDatePicker.js
+++ b/src/components/SQForm/SQFormDatePicker.js
@@ -100,7 +100,10 @@ SQFormDatePicker.propTypes = {
   onBlur: PropTypes.func,
   /** Custom onChange event callback */
   onChange: PropTypes.func,
-  /** Disable specific date(s) (day: DateIOType) => boolean */
+  /** Disable specific date(s) (day: DateIOType) => boolean
+   * This is a predicate function called for every day of the month
+   * Return true to disable that day or false to enable that day
+   */
   setDisabledDate: PropTypes.func,
   /** Any valid prop for material ui datepicker child component - https://material-ui.com/components/pickers/  */
   muiFieldProps: PropTypes.object,
--- a/src/components/SQForm/SQFormDatePicker.js
+++ b/src/components/SQForm/SQFormDatePicker.js
@@ -25,7 +25,7 @@ function SQFormDatePicker({
   placeholder = '',
   onBlur,
   onChange,
-  setDisabledDates,
+  setDisabledDate,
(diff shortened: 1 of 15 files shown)


src/components/SQForm/SQFormDatePicker.js
```diff
@@ -61,9 +64,9 @@ function SQFormDatePicker({
               {...inputProps}
               name={name}
               color="primary"
-              disabled={isDisabled}
```
SeanGroff, line 64: This is redundant and doesn't do anything so removed it.

src/components/SQForm/SQFormDatePicker.js
```diff
@@ -25,7 +25,9 @@ function SQFormDatePicker({
   placeholder = '',
   onBlur,
   onChange,
-  muiFieldProps = {}
+  setDisabledDates,
```
saadurrehman-selectquote, line 28: should it be  `setDisabledDate` instead of  `setDisabledDates`

stories/SQFormDatePicker.stories.js
```diff
@@ -46,3 +52,50 @@ export const basicDatePicker = () => {
     </SQForm>
   );
 };
+
+export const onlyEnableFirstDayOfMonthDatePicker = () => {
+  const setDisabledDates = date => {
+    // disable all days EXCEPT first day of the month
+    if (date.date() !== 1) {
```
saadurrehman-selectquote, line 59: Awesome

SeanGroff: Yes, this is more accurate. I'll commit this change shortly thanks!
The Pickers iterates over every day in the calendar and calls this function on each day

caleboleary (approved):

caleboleary:
Nice work Sean, and thanks for providing examples. One non-blocker nit - might be good to mention in the proptypes that the setDisabledDate is just a predicate which every rendered day is checked against. Took me looking at your examples to understand how to use that.

20BBrown14 (approved):

Issue #69, opened 2021-01-13
SQFormDialogStepper expose a way to reset/clear a steps values


Chris-Boyle:
Currently in the SQForm we have the `SQFormResetButtonWithConfirmation` that will reset all of the forms values.  We need to have a way to do something similar to that in the stepper.  We can not use the SQFormResetButtonWithConfirmation component because it will reset all steps values and not just the current step.  If we want to give the ability to reset the current steps values should there be a button in the footer or just a function that we call that is triggered by a consumer adding something like a TextButton to the body of the step?

umerjamal100:
Also using `SQFormResetButtonWithConfirmation` it just reset values to `initialValues`, if there is already a value present, it will reset to it, like in current loom we have ` address1: 'test'` in `initialValues` so on clicking reset  ` address1` is always set to test but we want it that address is cleared it should not have previous values it should empty the fields.
[loom](https://www.loom.com/share/bdd7fcd60ec248c2a50a16ac141a9d08)

Chris-Boyle:
@umerjamal100 for this I think we need to expose a way to get just this steps values.  Then a way to reset just the values in the current step.  We can add a button on the consumer side that calls that function.

umerjamal100:
We can solve this issue, by using `resetForm` method and we can pass props containing `newInitialValues` as an object containing new values, whenever we want to clean the initial Values of form or place custom values in it. I Will use a textButton for it and it will be placed in the header like this. Here are the useful links, please let me know if any suggestions.


![https://user-images.githubusercontent.com/18453833/105382874-5c348c80-5c32-11eb-9567-ecbd260dd938.png](https://user-images.githubusercontent.com/18453833/105382874-5c348c80-5c32-11eb-9567-ecbd260dd938.png)

[https://formik.org/docs/api/formik#resetform-nextstate-partialformikstatevalues--void](https://formik.org/docs/api/formik#resetform-nextstate-partialformikstatevalues--void)
![https://user-images.githubusercontent.com/18453833/105382075-86d21580-5c31-11eb-9fa2-3d9f0f02071e.png](https://user-images.githubusercontent.com/18453833/105382075-86d21580-5c31-11eb-9fa2-3d9f0f02071e.png)

SeanGroff:
@umerjamal100 Absolutely.

I touch on this here https://github.com/SelectQuoteLabs/SQForm/issues/78#issuecomment-762993922

PR #82, opened 2021-01-22
Issue-69 SQFormDialogStepper expose a way to reset/clear steps values

umerjamal100:
Resolves [#69 ](https://github.com/SelectQuoteLabs/SQForm/issues/69)
Currently, in the SQForm we have the SQFormResetButtonWithConfirmation that will reset all of the values of the form. We need to have a way to do something similar to that in the stepper. We can not use the SQFormResetButtonWithConfirmation component because it will reset all steps values and not just the current step. If we want to give the ability to reset the current steps values should there be a button in the footer or just a function that we call that is triggered by a consumer adding something like a TextButton to the body of the step?
Also using SQFormResetButtonWithConfirmation it just reset values to initialValues, if there is already a value present, it will reset to it, like in current loom we have address1: 'test' in initialValues so on clicking reset address1 is always set to test but we want it that address is cleared it should not have previous values it should empty the fields.

[https://www.loom.com/share/5ca3e9797fd94c5a8e1f60e9ef641974](https://www.loom.com/share/5ca3e9797fd94c5a8e1f60e9ef641974)


stories/SQFormDialogStepper.stories.js
```diff
@@ -240,6 +242,17 @@ export const SQDialogStepperWithValidation = () => {
             lastName: Yup.string().required('Required')
           }}
         >
```
Chris-Boyle, line 244: For the stories we want the examples to be as realistic/good as the code a consumer would write.  We should push the textButton into a section header like we have done in our app.

stories/SQFormDialogStepper.stories.js
```diff
@@ -240,6 +242,17 @@ export const SQDialogStepperWithValidation = () => {
...
+          <SQFormResetInitialValuesButton
+            tooltip="Clear Addresses"
+            confirmationContent="You are about to reset this form. Any unsaved info for this customer will be removed"
+            initialValuesObject={{
```
Chris-Boyle, line 248: You should create a `personalDateInitialValues` object with the values for that step (firstName: '', lastName: '', middleI: '', alias, newAccount).  Then you can spread that into `initialValuesObject` here and above into the `initialValues` on line 224 so that we don't have mismatched initial values.

stories/SQFormDialogStepper.stories.js
```diff
@@ -273,6 +286,16 @@ export const SQDialogStepperWithValidation = () => {
             age: Yup.string().required('Required')
           }}
         >
+          <SQFormResetInitialValuesButton
+            tooltip="Reset Account Details"
+            confirmationContent="You are about to reset this form. Any unsaved info for this customer will be removed"
+            initialValuesObject={{
```
Chris-Boyle, line 292: same comment down here about creating one object for the initial values of this step and passing it here and to the initialValues above.

Chris-Boyle:
@umerjamal100 when you push up a new commit make sure you pick the right commit style.  In this case it should be "new feature".  This will make your commit message start with `feat: ` and that will tell our CI/CD pipeline to create a new tag of the library before it deploys.  Let me know if you have any questions on this.
image

stories/SQFormDialogStepper.stories.js
umerjamal100: @Chris-Boyle used them separately as I wanted to show that we can assign new values also If we create one object `personalDateInitialValues` for both` initialValuesObject` and `initialValues ` we won't be able to demonstrate that.

Chris-Boyle: What would be a scenario where we would want to do that?  Maybe we have some prefilled values for the initial values.  Then when we hit Reset we would want that step to have no values?  You can have one example doing that.  Any the other one doing what I suggested since I believe what I suggested before is what most consumers would do.

stories/SQFormDialogStepper.stories.js
```diff
@@ -221,15 +236,7 @@ export const SQDialogStepperWithValidation = () => {
...
-        initialValues={{
-          firstName: '',
-          lastName: '',
-          middleI: '',
-          newAccount: false,
-          accountID: '',
-          description: '',
-          age: ''
-        }}
+        initialValues={personalDateInitialValues}
```
Chris-Boyle, line 239: this should contain the `personalDataInitialValues` and the `accountDetailsInitialValues`
`initialValues={...personalDataInitialValues, ...accountDetailsInitialValues}`
The spelling on personalDataInitialValues is a little off too.  It says Date in the middle instead of Data

Chris-Boyle:
We need to add a notes MD file to describe this component and how to use it.  You can follow some of the other examples in the project.

Chris-Boyle (approved):

Issue #59, opened 2021-01-11
SQForm submit button disabled with valid initial values


20BBrown14:
When using SQForm, the submit button is disabled even when the initial values provided to the form are valid values that satisfy the supplied validation schema. I believe this is because the form isn't dirty which is one of the requirements to enable the submit button and allow the form to be submitted.

See loom video: https://www.loom.com/share/cd911ae685f74441a53fc472cf5d441e

This is causing odd and semi-blocked behavior on LHAQT.

Chris-Boyle:
Affected components: SQform, SQFormDialog, and SQFormDialogStepper

20BBrown14:
So the state of the SQForm button is controlled by [useFormButton](https://github.com/SelectQuoteLabs/SQForm/blob/master/src/components/SQForm/useFormButton.js#L4).

I think if we removed the [checks](https://github.com/SelectQuoteLabs/SQForm/blob/master/src/components/SQForm/useFormButton.js#L11) for [`dirty`](https://github.com/SelectQuoteLabs/SQForm/blob/master/src/components/SQForm/useFormButton.js#L15) the form will still work the way we want with the added benefit of having the "submit" button enabled in scenario described above.

Thoughts?

SeanGroff:
@20BBrown14 Precisely.

https://github.com/SelectQuoteLabs/SQForm/pull/76#pullrequestreview-569654374

The real leg work is smoke testing all of the forms in our storybook stories to ensure this doesn't cause a new edge case.

SeanGroff:
Working on this issue now that should close a few reported issues.
* Don't rely on `isTouched` for `isRequired` and `isFulfilled` states
* Same for `isDisabled` state for `useFormButton`
* Provide non-default option to disable Form button if no updates to the form have occurred

Issue #83, opened 2021-01-22
SQForm initialErrors is never set


SeanGroff:
Issue:

When using the `validationSchema` prop, the `errors` property from `Formik` is initialized to an object, where each key is the `name` of the key from the `validationSchema` and the value is `Required` for any field in the `validationSchema` set to `.required()`. If an entry in `validationSchema` is NOT required, that entry is excluded from the `errors` object in Formik.

For example:

```javascript
const validationSchema = {
  name: Yup.string().required(),
  state: Yup.string()
}

// errors object would be set to:
{
   name: 'Required'
}

// initialErrors is never set so it's set to:
{}
```

When calling the `resetForm` method from Formik, among other things this method resets the `errors` object to the value from `initialErrors`. Since `initialErrors` is never set, `errors` gets set to `{}`. 

Any functionality that depends on the presence of `errors` after the form has been reset will experience unexpected behavior

Issue #47, opened 2021-01-04
SQForm Required field isn't always fulfilled correctly


SeanGroff:
After digging into this issue with @chris-boyle-sq on the auto complete component I believe we need to change how we are calculating the isRequired and isFulfilled values.

Moving away from isTouched in favor of a field value for `isRequired` and `isFulfilled` makes a lot of sense.

PR #84, opened 2021-01-25
feat: 🎸 Fixes submit button, reset button, helper text behavior

SeanGroff:
Apologies for the large PR but I didn't want to split this up into multiple PR's because each would be its own version that would introduce new bugs. The re-work required all of this work to be together. Also of note, no breaking changes were needed.

This PR closes 3 issues, possibly more when I go through the issues to see if they are related.

It Fixes the submit button to be disabled when it shouldn't. Fixes the reset
button to reset to the proper initial errors state. Fixes the helper
text to always be in a correct state.

This PR doesn't affect the buttons on the Stepper component. Those buttons do not use the useSQButton hook because the Stepper component has to account for all the steps and forms as a whole.

In addition to the loom explanations, the Date picker and DateTimePicker were not properly rendering the helper text error icon when in the error state. This PR fixes that slight UI issue as well.

Looms:
https://www.loom.com/share/bb878c53cfb447c38ac144bcab0c8ef0
https://www.loom.com/share/7faa684ed86247248cd419ac7b65b026
https://www.loom.com/share/9a5d63dbaed045f3a42403b90a047298

✅ Closes: #59, #83, #47


stories/SQFormDialog.stories.js
```diff
@@ -42,6 +42,7 @@ export const sqFormDialog = () => (
         false,
         'Toggle disabled state of Save Button'
       )}
+      shouldRequireFieldUpdates={true}
```
20BBrown14, line 45: nit: Don't need to specify the `true` value

SeanGroff: https://reactjs.org/docs/jsx-in-depth.html#props-default-to-true

React recommends always passing a prop value

20BBrown14: News to me. Thanks

src/hooks/useInitialRequiredErrors.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+// Until Formik exposes the validationSchema (again) via Context, the solution has to be handled at the Form declaration level
+// There's a few open PR's on this issue, here's one for reference: https://github.com/formium/formik/pull/2933
+export function useInitialRequiredErrors(validationSchema = {}) {
+  const initialRequiredErrorsRef = React.useRef(
+    Object.entries(validationSchema).reduce((acc, [key, value]) => {
+      if (value._exclusive?.required) {
+        return {...acc, [key]: 'Required'};
+      }
```
natedean, line 10: It'd be totally cool to mutate the acc here, rather than creating the new object literal and returning it.  Ever so slightly more performant, arguably easier to read.

natedean (approved):

caleboleary (approved):
Props on the descriptive looms. Look forward to updating to this version.

Issue #60, opened 2021-01-11
SQForm Dialogs should let a user know that the form is dirty before they cancel/close


Chris-Boyle:
**Overview:** In an edit scenario we pull up a dialog with all of the initial values set.  If we change a value we can now submit the values with the one change.  We can also Cancel out of the dialog at any point.  If we have dirtied the form by changing a value and then click cancel/close our form should check that we didn't want to Submit/Save that value.  We should get some kind of "Are you sure you want to cancel without saving your changes?" message.

**Impacts**: SQFormDialog and SQFormDialogStepper

**Loom**: https://www.loom.com/share/06b59e404ebd492aac23a2fd782af555

Chris-Boyle:
Worked with UX on a comp.
Modal Title : “Cancel [Whatevs]”
Modal Body : You currently have unsaved changes which will be lost if you continue.
Primary Action : “Continue”
Secondary Action : “Go Back”

PR #85, opened 2021-01-28
Added Are you Sure? Dialog in SQFormDialog component

shahzaib-haider1:
✅ Closes: #60 
Added Are you Sure? Dialog Alert in SQFormDialog

AYS modal will be open if the form is dirty otherwise dialog will be close upon clicking the cancel button.

here is an updated  loom video
https://www.loom.com/share/c15d212716e9415eb2af9bc20e7a4c4a

### DIFF
--- a/src/components/SQFormDialog/SQFormDialog.js
+++ b/src/components/SQFormDialog/SQFormDialog.js
@@ -20,7 +20,9 @@ function SQFormDialog({
   initialValues,
   muiGridProps = {},
   shouldRequireFieldUpdates = false,
-  validationSchema
+  validationSchema,
+  cancelConfirmDialogProps = {},
+  showCancelConfirmDialog = true
 }) {
   const validationYupSchema = React.useMemo(() => {
     if (!validationSchema) return;
@@ -52,6 +54,8 @@ function SQFormDialog({
         shouldRequireFieldUpdates={shouldRequireFieldUpdates}
         title={title}
         muiGridProps={muiGridProps}
+        showCancelConfirmDialog={showCancelConfirmDialog}
+        cancelConfirmDialogProps={cancelConfirmDialogProps}
       />
     </Formik>
   );
@@ -82,13 +86,25 @@ SQFormDialog.propTypes = {
   initialValues: PropTypes.object.isRequired,
   /** Any prop from https://material-ui.com/api/grid */
   muiGridProps: PropTypes.object,
-  /** Whether or not the dialog form requires updates to the form to enable the submit button */
+  /** open/close props for cancel confirmation dialog*/
   shouldRequireFieldUpdates: PropTypes.bool,
   /**
    * Yup validation schema shape
    * https://jaredpalmer.com/formik/docs/guides/validation#validationschema
    * */
-  validationSchema: PropTypes.object
+  validationSchema: PropTypes.object,
+  /** Whether or not the Cancel Confirm Dialog nedeeded */
+  showCancelConfirmDialog: PropTypes.bool,
+  /**
+   * Any Valid props for Cancel Confirmation Dialog
+   * title: Modal Title
+   * secondaryButtonText: The secondary button text (Button located on left side of Dialog)
+   * primaryButtonText: The primary button text (Button located on right side of Dialog)
+   * content: The content to be rendered in the dialog body
+   * https://jaredpalmer.com/formik/docs/guides/validation#validationschema
+   *
+   * */
+  cancelConfirmDialogProps: PropTypes.object
 };
 
 export default SQFormDialog;
--- a/src/components/SQFormDialog/SQFormDialog.js
+++ b/src/components/SQFormDialog/SQFormDialog.js
@@ -22,7 +22,7 @@ function SQFormDialog({
   shouldRequireFieldUpdates = false,
   validationSchema,
   cancelConfirmDialogProps = {},
-  showCancelConfirmDialog = true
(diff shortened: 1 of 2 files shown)


src/components/SQFormDialog/SQFormDialog.js
```diff
@@ -88,7 +92,19 @@ SQFormDialog.propTypes = {
...
+  /**
+   * Any Valid props for Cancel Confirmation Dialog
+   * title: Modal Title
+   * secondaryButtonText: The secondary button text (Button located on left side of Dialog)
+   * primaryButtonText: The primary button text (Button located on right side of Dialog)
+   * content: The content to be rendered in the dialog body
+   * https://jaredpalmer.com/formik/docs/guides/validation#validationschema
+   *
+   * */
+  cancelConfirmDialogProps: PropTypes.object
```
SeanGroff, line 107: Instead of a comment, the prop type itself should document the shape. Use prop type shape instead of prop type object please.

src/components/SQFormDialog/SQFormDialogInner.js
```diff
@@ -40,54 +40,96 @@ function SQFormDialogInner({
...
+    isDialogAlertModalOpen,
+    {openDialog: openDialogAlertModal, closeDialog: closeDialogAlertModal}
+  ] = useDialog();
+
+  const handleCancel = () => {
+    if (!showCancelConfirmDialog) {
+      resetForm();
+      onClose();
+
+      return;
```
SeanGroff, line 60: For short and sweet if/else scenarios that don't return a value, if/else block is generally more declarative.

SeanGroff (changes requested):
Please use a conventional commit or else the npm version won't increment. 

npm run commit

Select "feature" for the commit type. 

Please see the Contributing section of the ReadMe for full details.

src/components/SQFormDialog/SQFormDialogInner.js
```diff
@@ -40,54 +40,96 @@ function SQFormDialogInner({
   saveButtonText,
   shouldRequireFieldUpdates = false,
   title,
-  muiGridProps
+  muiGridProps,
+  showCancelConfirmDialog,
```
umerjamal100, line 44: We need to add `showCancelConfirmDialog` and `cancelConfirmDialogProps` in `SQFormDialogInner.propTypes`.

shahzaib-haider1: Sure @umerjamal100 Thank you

src/components/SQFormDialog/SQFormDialogInner.js
saadurrehman-selectquote: @SeanGroff are we not following internal coding standards for SQForm?

umerjamal100 (approved):

src/components/SQFormDialog/SQFormDialogInner.js
```diff
@@ -40,54 +40,94 @@ function SQFormDialogInner({
...
+                {saveButtonText}
+              </SQFormButton>
+            )}
+          </DialogActions>
+        </Form>
+      </Dialog>
+
+      {openCancelConfirmationDialog && (
+        <DialogAlert
+          isOpen={isDialogAlertModalOpen}
```
Chris-Boyle, line 118: [blocker] It looks like the only props from `cancelConfirmDialogProps` that are being used is the content.  If we are not going to use the other props we should remove them from the propTypes and possibly simplify the `cancelConfirmDialogProps` to just one prop that is a string.  Possibly `cancelConfirmDialogBody`

Chris-Boyle: If UX did not request the dialog box to be customizable we should not allow consumers to customize it.  We should just use the values that UX gave us.

shahzaib-haider1: @Chris-Boyle I have updated the PR please have a look

Chris-Boyle (approved):

src/components/SQFormDialog/SQFormDialogInner.js
```diff
@@ -45,49 +45,76 @@ function SQFormDialogInner({
...
   const {resetForm} = useSQFormContext();
 
-  const handleClose = () => {
+  const [
+    isDialogAlertOpen,
+    {openDialog: openDialogAlert, closeDialog: closeDialogAlert}
+  ] = useDialog();
+
+  const handleCancel = () => {
+    openDialogAlert();
```
Chris-Boyle, line 54: Are you popping the modal open every time Cancel is clicked or only if the form is dirty?  I think we want to only do open the dialog if the form is dirty.

SeanGroff: This is still a blocker from my first review @shahzaib-haider1 
Your other changes look really good. From what I see, this is the last remaining issue.

The AYS modal should only open IF changes to the form have been made
Otherwise, the cancel button closes the modal as usual.

SeanGroff (changes requested):
Remaining changes I'm requesting
* AYS modal only opens IF changes to the form fields have been made
* Sync this feature branch with the latest master branch

SeanGroff (approved):

src/components/SQFormDialog/SQFormDialogInner.js
shahzaib-haider1: @SeanGroff I have updated the PR and now AYS modal will be open when the form is dirty

Issue #91, opened 2021-02-16
Change condition for rendering the Fulfilled Icon on fields


SeanGroff:
If a field contains a value AND is not in an error state -> Display fulfilled icon

Approved by UX

PR #93, opened 2021-02-16
fix: 🐛 Field is fulfilled if theres no error and theres a value

SeanGroff:
✅ Closes: #91

Loom: https://www.loom.com/share/c06513c248454d00b6690c75ad045056

Now, the fulfilled icon only appears when a field is not in an error state and a value is present.
Also fixed a bug where the value `0` (or any falsy number) in a numeric input would be treated as a non-value.

### DIFF
--- a/src/components/SQForm/useForm.js
+++ b/src/components/SQForm/useForm.js
@@ -14,11 +14,7 @@ function _handleError(name, isRequired) {
   }
 }
 
-function _getIsFulfilled(isRequired, hasValue, isError) {
-  if (!isRequired && !isError) {
-    return true;
-  }
-
+function _getIsFulfilled(hasValue, isError) {
   if (hasValue && !isError) {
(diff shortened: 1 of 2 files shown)


src/components/SQForm/useForm.js
```diff
@@ -33,11 +29,11 @@ function _getHasValue(meta) {
     return !!fieldValue.length;
   }
 
-  if (fieldValue) {
+  if (typeof fieldValue === 'number') {
     return true;
   }
 
-  return false;
+  return !!fieldValue ?? false;
```
lucas-homer, line 36: this got me thinking... what does `!!fieldValue` eval to when `fieldValue` is undefined or null? False! This is minor, and not a blocker per se, but... isn't the nullish coalescing unnecessary here?

lucas-homer (approved):

Issue #96, opened 2021-02-18
SQFormDialogStepper Add ability to disable submit/next button programmatically


Chris-Boyle:
**As a** user of the stepper
**I want to** be able to disable the next/submit button until an API call or some other function occurs outside of the Formik context
**So that I** can stop users from going forward to the next step or submitting.

PR #97, opened 2021-02-18
feat: 🎸 Added ability to disable next in the stepper

Chris-Boyle:
✅ Closes: #96

As a user of the stepper
I want to be able to disable the next/submit button until an API call or some other function occurs outside of the Formik context
So that I can stop users from going forward to the next step or submitting.

Loom uses local state that I added to the story.  Couldn't commit it but it was needed for the demo.
https://www.loom.com/share/8049ff6117704737b0080c52af89c0d4


20BBrown14 (approved):
🔥

SeanGroff (approved):

Issue #64, opened 2021-01-12
SQForm Radio Buttons


SeanGroff:
Build a Radio Buttons/Group component for the SQForm library.
Here's the blueprint which is the official example from Formik [Radio Group Example](https://github.com/formium/formik/tree/master/examples/radio-group)

PR #100, opened 2021-02-25
feat: 🎸 Added SQFormRadioButtonGroup component

20BBrown14:
Added SQFormRadioButtonGroup component to allow for a radio button input
within the context of SQForm

✅ Closes: #64

Checkout this mind blowing loom: https://www.loom.com/share/3bc2d0ed03e44fd19ef52bbe39b88381

Screen Shot 2021-03-01 at 11 38 38 AM
Screen Shot 2021-03-01 at 11 38 44 AM

### DIFF
--- a/src/components/SQForm/SQFormRadioButtonGroup.js
+++ b/src/components/SQForm/SQFormRadioButtonGroup.js
@@ -1,0 +1,87 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Grid from '@material-ui/core/Grid';
+import RadioGroup from '@material-ui/core/RadioGroup';
+import FormHelperText from '@material-ui/core/FormHelperText';
+import InputLabel from '@material-ui/core/InputLabel';
+import {SQFormRadioButtonGroupItem} from '../../../src';
+import {useForm} from './useForm';
+
+function SQFormRadioButtonGroup({
+  name,
+  onChange,
+  isRequired = false,
+  shouldDisplayInRow = false,
+  size = 'auto',
+  groupLabel,
+  children
+}) {
+  const {
+    formikField: {field},
+    fieldHelpers: {handleChange, HelperTextComponent}
+  } = useForm({
+    name,
+    isRequired,
+    onChange
+  });
+
+  const childrenToRadioGroupItems = () => {
+    return children.map(radioOption => {
+      const {label, value, isDisabled, inputProps} = radioOption;
+      return (
+        <SQFormRadioButtonGroupItem
+          label={label}
+          value={value}
+          isDisabled={isDisabled}
+          isRowDisplay={shouldDisplayInRow}
+          inputProps={inputProps}
+          key={`SQFormRadioButtonGroupItem_${value}`}
+        />
+      );
+    });
+  };
+
+  return (
+    <Grid item sm={size}>
+      <InputLabel id={groupLabel.toLowerCase()}>{groupLabel}</InputLabel>
+      <RadioGroup
+        value={field.value}
+        row={shouldDisplayInRow}
+        aria-label={`SQFormRadioButtonGroup_${name}`}
+        name={name}
+        onChange={handleChange}
+      >
+        {childrenToRadioGroupItems()}
+      </RadioGroup>
+      <FormHelperText required={isRequired}>
+        {HelperTextComponent}
+      </FormHelperText>
+    </Grid>
+  );
+}
+
+SQFormRadioButtonGroup.propTypes = {
+  /** Name of the radio group */
+  name: PropTypes.string.isRequired,
+  /** Function to call on value change */
+  onChange: PropTypes.func,
+  /** Whether this radio selection is required */
+  isRequired: PropTypes.bool,
+  /** Whether to display group in row */
+  shouldDisplayInRow: PropTypes.bool,
+  /** Size of the input given full-width is 12. */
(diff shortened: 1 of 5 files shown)


src/components/SQForm/SQFormRadioButtonGroup.js
```diff
@@ -0,0 +1,79 @@
...
+import Grid from '@material-ui/core/Grid';
+import RadioGroup from '@material-ui/core/RadioGroup';
+import FormHelperText from '@material-ui/core/FormHelperText';
+import InputLabel from '@material-ui/core/InputLabel';
+import {SQFormRadioButtonGroupItem} from '../../../src';
+import {useForm} from './useForm';
+
+function SQFormRadioButtonGroup({
+  name,
+  onChange = () => {},
```
SeanGroff, line 12: Let's leave this `undefined` (no default value)
See this line in `useForm`

https://github.com/SelectQuoteLabs/SQForm/blob/755d6e9b2df812daf500f3ce8d7bfc4dab78a947/src/components/SQForm/useForm.js#L54

src/components/SQForm/SQFormRadioButtonGroup.js
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Grid from '@material-ui/core/Grid';
+import RadioGroup from '@material-ui/core/RadioGroup';
+import FormHelperText from '@material-ui/core/FormHelperText';
+import InputLabel from '@material-ui/core/InputLabel';
+import {SQFormRadioButtonGroupItem} from '../../../src';
+import {useForm} from './useForm';
+
+function SQFormRadioButtonGroup({
```
SeanGroff, line 10: Can we also expose `onBlur` to maintain a consistent API with our other form elements?

src/components/SQForm/SQFormRadioButtonGroup.js
```diff
@@ -0,0 +1,79 @@
...
+    onChange
+  });
+
+  const childrenToRadioGroupItems = () => {
+    return children.map(radioOption => {
+      return (
+        <SQFormRadioButtonGroupItem
+          label={radioOption.label}
+          value={radioOption.value}
+          inputProps={radioOption.inputProps}
```
SeanGroff, line 33: Can we add an `isDisabled` prop in addition to the `inputProps`?
I know we can accomplish a disabled state with `inputProps` but it's such a common need it's good to make it clear in the props API without having to go to the MUI docs.
This would also maintain a consistent API with the other SQForm elements

SeanGroff (changes requested):
Seriously outstanding work @20BBrown14 💪
In addition to the minor tweaks, can you demo this with Chris Nemeth to perform a design review?

It would be great to see the Radio Group added to an existing story so we can see how it lines up with the other elements and ensure the typography matches up.

src/components/SQForm/SQFormRadioButtonGroup.js
20BBrown14: Do Radio buttons blur, though? MUI `[RadioGroup](https://material-ui.com/api/radio-group/)` doesn't expose an `onBlur` prop. Neither does the `[Radio](https://material-ui.com/api/radio/) component.

SeanGroff: IF MUI doesn't expose it, then disregard it. If the need develops in the future, we can circle back.

SeanGroff (approved):

20BBrown14:
> In addition to the minor tweaks, can you demo this with Chris Nemeth to perform a design review?

I talked with Chris and he only one suggestion. When the radio buttons are in a row he'd like to see more space inbetween the buttons. I'm working on that now.

20BBrown14:
> > In addition to the minor tweaks, can you demo this with Chris Nemeth to perform a design review?
> 
> I talked with Chris and he only one suggestion. When the radio buttons are in a row he'd like to see more space inbetween the buttons. I'm working on that now.

When the radio group is displayed in a row there is a right margin of 30px now, defaultw as 16. Per Chris N. I wanted to make sure that there was no extra margin if the group was displayed vertically to avoid any unexpected layout issues so that's why i added that check in the group item specifically.

src/components/SQForm/SQFormRadioButtonGroup.js
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Grid from '@material-ui/core/Grid';
+import RadioGroup from '@material-ui/core/RadioGroup';
+import FormHelperText from '@material-ui/core/FormHelperText';
+import InputLabel from '@material-ui/core/InputLabel';
+import {SQFormRadioButtonGroupItem} from '../../../src';
```
hvilander, line 7: absolute imports?

hvilander (approved):
Nice work

src/components/SQForm/SQFormRadioButtonGroup.js
20BBrown14: Absolute import won't work here. I *think* it's because webpack starts resolving things, for the purpose of absolute imports, starting in the `src` directory. So importing from `src` directory directly, as an absolute import, webpack can't resolve it.

SeanGroff: I've never configured webpack to handle absolute imports for this project. Since it's not a create-react-app project I have to customize webpack. Since I don't believe we ever go deeper than 3 or 4 levels we are probably ok without it.

SeanGroff (approved):

Issue #92, opened 2021-02-16
SQFormDropdown not Orange in Error State


SeanGroff:
When the SQFormDropdown component falls into an error state, its label and input bottom border line doesn't change to Orange.

See Example Screenshot

![Screen Shot 2021-02-16 at 3 45 32 PM](https://user-images.githubusercontent.com/4367312/108125602-4b7b0900-706e-11eb-9079-39ef18507414.png)

PR #102, opened 2021-03-02
fix: 🐛 SQFormDropdown - add error props to InputLabel and Select

codyarose:
✅ Closes: #92 

`SQFormDropdown` now properly shows the orange label and underline when in an error state.
Loom: https://www.loom.com/share/24321e12b6a347caaa1e6b9c73104b99


SeanGroff (approved):
Thank you for the 🐛 fix @codyarose!!

Issue #101, opened 2021-03-01
SQForm Checkbox group with label


20BBrown14:
Per Chris Nemeth: 

"Utilize the work Branden has done in Issue #64 for Radio Button Fields to do so for Checkbox Fields."

PR #109, opened 2021-03-03
feat: 🎸 Rename SQFormCheckboxGroup and replace it

20BBrown14:
Renamed SQFormCheckboxGroup(Item) to SQFormInclusionList(Item) and readd
SQFormCheckboxGroup(Item) to creage a checkbox group to be used within
SQForm and other form elements

BREAKING CHANGE: 🧨 Rename SQFormCheckboxGroup(Item) to SQFormInclusionList(Item)

Diff is pretty bad since we reused a file.

✅ Closes: #101

Look at this loom! https://www.loom.com/share/0373e3e4221b4a04a60c5c1568477410

Screen Shot 2021-03-03 at 10 46 30 AM

### DIFF
--- a/notes/SQForm.md
+++ b/notes/SQForm.md
@@ -30,7 +30,11 @@ Use like any other checkbox. The value is provided from SQForm like the rest of 
 
 **SQFormCheckboxGroup & SQFormCheckboxGroupItem**
 
-`SQFormCheckboxGroup` acts as `SQFieldArray` does, using render props to pass `arrayHelpers` to your array of `SQFormCheckboxGroupItem`s. The required `name` prop must match the `initialValues` array property to store the checked items to be saved. `useSelectAll` triggers the 'Select All' checkbox at the top of the list, and if set to `true`, it must also receive an array `selectAllData` that represent the list of items to save if 'Select All' is checked. See 'Form With Checkbox Group' story for full example.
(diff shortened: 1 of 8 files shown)


SeanGroff (approved):
This looks great! Nice work @20BBrown14 🚀

20BBrown14:
> If you have not yet done so, can you warn the Squad PM and tech lead that will be effected by this breaking change so they account for it when the time comes to update their SQForm version?

Taken care of. I let Danish S know who is the tech lead on the one project i found that used the old SQFormCheckboxGroup

codyarose (changes requested):
Looks great!
issue: The only thing I see that may be missing is error handling on the form components using `isFieldError` from `useForm`

lucas-homer (approved):

20BBrown14:
> Looks great!
> issue: The only thing I see that may be missing is error handling on the form components using `isFieldError` from `useForm`

I'm not sure that applies to this component. MUI doesn't provide an error field. Not from what I can see at least. And it makes sense intuitively that a checkbox group wouldn't have an error other minimum selected, which can be provided through yup.

If a user isn't allowed to select specific combinations then the consumer of this component should be listening for the changes and dynamically disable options as appropriate. 

Did you have a specific case in mind about why we would need to utilize isFieldError in another capacity besides the HelperTextComponent?

codyarose:
> > Looks great!
> > issue: The only thing I see that may be missing is error handling on the form components using `isFieldError` from `useForm`
> 
> I'm not sure that applies to this component. MUI doesn't provide an error field. Not from what I can see at least. And it makes sense intuitively that a checkbox group wouldn't have an error other minimum selected, which can be provided through yup.
> 
> If a user isn't allowed to select specific combinations then the consumer of this component should be listening for the changes and dynamically disable options as appropriate.
> 
> Did you have a specific case in mind about why we would need to utilize isFieldError in another capacity besides the HelperTextComponent?

Okay, I looked through the MUI docs more and I see what you mean 👍

Issue #17, opened 2020-12-03
SQDialogForm only body should be scrollable


20BBrown14:
I would like to see the option to have only the body be scrollable in the SQDialogForm so the header and footer buttons are always visible. You can see current functionality here: https://www.loom.com/share/dd97c034de844526afa290fbfb9adf0a
Let me know if any more details are needed.

SeanGroff:
@20BBrown14 Looking over the Loom video can you try adjusting the `maxWidth` prop of the Modal?
Ideally, modals should never scroll vertically. It's a UI smell.

Can you try making the maxWidth larger to accommodate for the large body of text and report back? I would try `xl`

Here are the `maxWidth` available options

```
  /** Determine the max-width of the dialog. The dialog width grows with the size of the screen. Set to false to disable maxWidth. */
  maxWidth: PropTypes.oneOf(['xs', 'sm', 'md', 'lg', 'xl', false]),
```

20BBrown14:
I'm not so sure that'll solve the issue. If we have a long block of text that's wrapping and causing the modal to need to scroll this could be an issue. But if I have a lot of questions (specific to this instance) with larger or smaller label text this won't solve the issue. See: https://www.loom.com/share/006d03958fbe41d08b2de2755be1e144

Chris-Boyle:
@20BBrown14 have you tried wrapping your content body in a div with a defined height?

20BBrown14:
@Chris-Boyle good call. What ultimately worked as setting a max-height to allow for dynamic sizes of the dialog depending on the size of the dialog body.

codyarose:
Sounds like some `position: sticky` on `<DialogTitle>` and `<DialogActions>` could solve this. Here's a quick demo just using inline styles: https://www.loom.com/share/5c5888520b8149ae99e9c763cc950513

PR #111, opened 2021-03-05
fix: 🐛  make only dialog body scroll

codyarose:
Used `position: sticky` to make the title and actions sections stay in
place while the body scrolls. Demo, deets, and edge case considerations in the loom

https://www.loom.com/share/ec3caa896f654b218bd6a40826bf5427

✅ Closes: #17

### DIFF
--- a/src/components/SQFormDialog/SQFormDialogInner.js
+++ b/src/components/SQFormDialog/SQFormDialogInner.js
@@ -10,6 +10,7 @@ import {
   Slide,
   makeStyles
 } from '@material-ui/core';
+import {useTheme} from '@material-ui/core/styles';
 import {Form} from 'formik';
 import {RoundedButton, DialogAlert, useDialog} from 'scplus-shared-components';
 import {useSQFormContext} from '../../index';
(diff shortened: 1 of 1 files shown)


Chris-Boyle (approved):
We probably need this in the stepper too.

src/components/SQFormDialog/SQFormDialogInner.js
```diff
@@ -19,12 +20,28 @@ const Transition = React.forwardRef((props, ref) => {
...
+    borderBottom: ({palette}) => `1px solid ${palette.divider}`
+  }
+});
 const useActionsStyles = makeStyles({
   root: {
     display: 'flex',
     justifyContent: 'space-between',
     flex: '1 1 100%',
-    padding: '16px 24px'
+    padding: '16px 24px',
```
SeanGroff, line 41: We can pass a callback to makeStyles that gives us access to the theme where we could use the built in theme spacing. 

theme.spacing(2) would result in 16. Material Design uses multiples of 8.

src/components/SQFormDialog/SQFormDialogInner.js
```diff
@@ -19,12 +20,28 @@ const Transition = React.forwardRef((props, ref) => {
   return <Slide direction="down" ref={ref} {...props} />;
 });
 
+const stickyStyles = {
+  position: 'sticky',
+  background: ({palette}) => palette.background.paper,
+  zIndex: 1
```
SeanGroff, line 26: We generally want to avoid magic numbers for z-index's by using the z-index values available in the theme. 
Since this is inside a dialog to make an element sticky this should be harmless.

SeanGroff (approved):
Nice solution @codyarose and great job presenting the drawbacks to max-height on the content 👌

PR #112, opened 2021-03-06
feat: 🎸 add input masking component

codyarose:
Added a masking text field component for phone, zip, and currency

https://www.loom.com/share/637f18e7a37c45fe92a71c5fcbc49a08


lucas-homer (approved):
this all looks good to me. my only question is if we have any other common mask types we should support from the the outset. Deferring to @SeanGroff on that, though.

20BBrown14:
Seems like we could make this mask component general enough for the consumer to pass a custom mask? Or one of the default options provided.

codyarose:
> Seems like we could make this mask component general enough for the consumer to pass a custom mask? Or one of the default options provided. 

Yeah, that's probably a good thing to add, otherwise the component user will have to request that a mask be added if it's not already an option

SeanGroff:
Converting this PR to a Draft.

This is discussion-worthy but here's what I'm thinking initially. 

We leverage inversion of control. The MaskInput component accepts a custom mask only.

We then export on Object/Map containing pre-defined masks we can supply to the consumer. Then the consumer inserts the mask they import from SQForm into the MaskInput component.

Whenever we discover we need a new mask, we PR the new mask to or Masks Map. No changes to the MaskInput component would ever be required to allow for new masks.

20BBrown14:
I think like idea @SeanGroff personally. But one question. Are you suggesting that this component *should* only be used with the masks that we're exporting? And we won't necessarily support a custom mask being passed in. 

I was thinking that we can provide common masks (and add more if we decide to) but the consumer can pass custom masks with the help of the `text-masks` module instead of user one of ours out of the box.

Just trying to make sure I understand what you're envisioning.

SeanGroff:
> I think like idea @SeanGroff personally. But one question. Are you suggesting that this component _should_ only be used with the masks that we're exporting? And we won't necessarily support a custom mask being passed in.
> 
> I was thinking that we can provide common masks (and add more if we decide to) but the consumer can pass custom masks with the help of the `text-masks` module instead of user one of ours out of the box.
> 
> Just trying to make sure I understand what you're envisioning.

So what I'm saying is the InputMaskComponent `mask` prop only accepts a "custom mask"  meaning it's up to the consumer to define the mask and pass it into the prop.

This custom prop can either be the (yet to be created) common masks we expose via the SQForm library, or a custom mask the developer comes up with for a special use case.

20BBrown14:
> This custom prop can either be the (yet to be created) common masks we expose via the SQForm library, or a custom mask the developer comes up with for a special use case.

Great. I love. This solution has my vote 👍

codyarose:
Sounds good to me 👍
I suppose the existence of that masks object would be mentioned in the docs/story?

SeanGroff:
> Sounds good to me 👍
> I suppose the existence of that masks object would be mentioned in the docs/story?

Precisely. In the future, we can add this to our Docs site for SQForm so it's more apparent.

src/components/SQForm/SQFormMaskedTextField.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import MaskedInput from 'react-text-mask';
+import SQFormTextField from './SQFormTextField';
+
+function TextFieldMask({inputRef, mask, ...rest}) {
+  return (
+    <MaskedInput
+      {...rest}
```
gdoohen1, line 9: This looks awesome, I actually need something just like this =)
Could we maybe have `placeholderChar` and `showMask` be props with defaults that way they are able to be specified with inputProps and overridden if needed?

SeanGroff: @gdoohen1 Could you clarify what `placeholderChar` and `showMask` would be used for.
Is this something that occurs in FlexForm you're needing? If so, can you screenshot or record a Loom to show the intent of these 2 props?

gdoohen1: @SeanGroff I don't have a use-case at the moment to utilize these, I was more just asking from the perspective of erroring on the side of exposing something rather preventing the user from accessing it - but if we have strong UX standards of never showing masking and using the same placeholder than probably the extra level of safety in locking these up is the better option.

src/components/SQForm/SQFormMaskedTextField.js
```diff
@@ -0,0 +1,68 @@
...
+      inputProps={{
+        ...inputProps,
+        mask
+      }}
+      muiFieldProps={muiFieldProps}
+    />
+  );
+}
+
+SQFormMaskedTextField.propTypes = {
```
20BBrown14, line 63: Our proptypes don't match the props we defined in the component definition.

src/utils/masks.js
```diff
@@ -0,0 +1,24 @@
...
+    /\d/,
+    /\d/,
+    /\d/,
+    '-',
+    /\d/,
+    /\d/,
+    /\d/,
+    /\d/
+  ],
+  zip: [/\d/, /\d/, /\d/, /\d/, /\d/],
```
20BBrown14, line 20: Would we also want a default for zip + 4?

src/components/SQForm/SQFormMaskedTextField.js
SeanGroff: Right on.
We try to keep the API small for ease of use then expand the props API as needs arise.
If the need arises for either of these features we'll enhance this component at that time.

src/utils/masks.js
```diff
@@ -0,0 +1,24 @@
+import createNumberMask from 'text-mask-addons/dist/createNumberMask';
+
+export const MASKS = {
+  phone: [
```
20BBrown14, line 4: Do we want a default phone mask that includes country code?

notes/SQForm.md
```diff
@@ -56,6 +56,10 @@ As shown in the component prop-types, the `children` of this component must be a
 
 If you want an empty option in the dropdown list options, set the `displayEmpty` prop to `true`.
 
+**SQFormMaskedTextField**
```
20BBrown14, line 59: Could we perhaps mention the module we're using to help with the masks? A link would be great so someone using this component could do their own DD if needed.

src/components/SQForm/SQFormMaskedTextField.js
gdoohen1: Alright that sounds good to me!

src/components/SQForm/SQFormMaskedTextField.js
codyarose: Are you referring to the props that get passed to SQFormTextField?

20BBrown14: Right

notes/SQForm.md
codyarose: Good call, I'll add that

src/components/SQForm/SQFormMaskedTextField.js
codyarose: Gotcha, I'll add that. Still getting used to how propTypes is set up, I'm more familiar with typing props in TypeScript

codyarose:
I'll defer to you all that are more familiar with how this would be used in production in regards to additional mask options, like the zip + 4 and phone country code @20BBrown14 mentioned. 

If you let me know what shape the mask should be in I'll work on adding it

SeanGroff:
> I'll defer to you all that are more familiar with how this would be used in production in regards to additional mask options, like the zip + 4 and phone country code @20BBrown14 mentioned.
> 
> If you let me know what shape the mask should be in I'll work on adding it

We should steal the masks that already exist for flexform. I believe they live in a utils folder on the CRM's. Check sc2-senior.

20BBrown14 (approved):

CHANGELOG.md
```diff
@@ -7,342 +7,298 @@
 
 ### [4.1.1](https://github.com/SelectQuoteLabs/SQForm/compare/v4.1.0...v4.1.1) (2021-03-09)
 
-
 ### Bug Fixes
 
-* 🐛 make only dialog body scroll ([60beb7b](https://github.com/SelectQuoteLabs/SQForm/commit/60beb7b1e1bcf4da4fd01a2debb6fd50310504ae)), closes [#17](https://github.com/SelectQuoteLabs/SQForm/issues/17)
+- 🐛 make only dialog body scroll ([60beb7b](https://github.com/SelectQuoteLabs/SQForm/commit/60beb7b1e1bcf4da4fd01a2debb6fd50310504ae)), closes [#17](https://github.com/SelectQuoteLabs/SQForm/issues/17)
```
SeanGroff, line 12: This file is auto-generated. @codyarose can you dig into what might be causing this issue? My guess is Prettier for VS Code is formatting this markdown file..

SeanGroff: Spoke with Cody and he's going to dig into this

src/utils/masks.js
```diff
@@ -0,0 +1,31 @@
+import createNumberMask from 'text-mask-addons/dist/createNumberMask';
+import emailMask from 'text-mask-addons/dist/emailMask';
+
+export const MASKS = {
```
SeanGroff, line 4: Can we expose the existing REGEX masks we seem to copy/paste from project to project?
This way we ensure we expose masks that already meet the business's needs.
https://bitbucket.org/search?q=validationhelpers.js&account=%7Bf0e46ff7-117d-4015-9027-923794846585%7D

SeanGroff (changes requested):
This is looking amazing. I'm excited to finally have masked fields in SQForm.
I requested to small changes. I'm good with merging this PR once the ChangeLog is removed from the git diff and the existing Regex masks we copy/pasta from project to project are exposed here.

src/utils/masks.js
codyarose: @SeanGroff  Seems like those regex are formatted different from what `react-text-mask` expects. Should I gather the ones that are common across all those files and create an equivalent of them for `react-text-mask`, or are you suggesting we pull in all of that validation logic as well?

SeanGroff: > @SeanGroff Seems like those regex are formatted different from what `react-text-mask` expects. Should I gather the ones that are common across all those files and create an equivalent of them for `react-text-mask`, or are you suggesting we pull in all of that validation logic as well?

Gather the ones that are common across all those files and create an equivalent of them for `react-text-mask` would be ideal! I haven't looked at react-text-mask in a long time so I don't have a better suggestion.

I just want to ensure how we validate a phone or a zipcode remains the same as whats in production for several apps currently.

codyarose: Gotcha. I'll work on that

codyarose:
@SeanGroff  Added an SSN mask, updated the zip mask to handle 5 or 5 + 4 digits, and added a little bit of regex validation to the phone number mask. Beyond that, the rest of the stuff in `validationHelpers` seems to be more for pure validation rather than masking.

I fear adding too much validation logic to the masks by preventing the user to input certain characters may end up making the inputs appear broken. Whereas being looser with what we allow the user to input and relying on yup validation for feedback we can let the user know that there's something wrong with what they've typed into the input rather than the input itself.

SeanGroff (approved):
Amazing @codyarose 🚀

PR #115, opened 2021-03-08
fix: 🐛 ?? operator instead of || for grid spacing prop check

20BBrown14:
Use nullish coalescing operator (??) instead of or (||) operator to
check for existence of `spacing` value which can be 0 which is falsy
causing unexpected values.

✅ Closes: 114


codyarose (approved):
Nice! Perfect use case for the nullish coalescing operator

SeanGroff (approved):
Thanks @20BBrown14 🔥

Issue #113, opened 2021-03-08
SQFormMultiSelect tooltip values & field values sorted differently


caleboleary:
<img width="340" alt="Screen Shot 2021-03-08 at 1 18 40 PM" src="https://user-images.githubusercontent.com/12816579/110370658-95259680-8011-11eb-92db-ff212ab545ac.png">

This was found by one of the QA testers on one of our projects.

When selecting multiple values in a SQFormMultiSelect drop down, the tooltip that displays when you hover will list your selections in the order they were clicked in, while the field itself will list selected options in order as they appear in the list.

In the above screenshot (taken in the storybook), I selected "Joe, Jessica, Jane, Jack" in that order, and that reflects in the tooltip but not the field itself.

PR #116, opened 2021-03-09
fix: 🐛 make multiselect renderValue and tooltip order match

codyarose:
Standardize the way the selected options are rendered in the tooltip and Select renderValue prop.

In `getRenderValue` the return value is based on filtering the `children` array using the selected values, then getting the labels of whatever is left - this resulted in the order of the displayed selected options being based on the order in which they occurred in the `children` array.

In `getTooltipValue` we start with the selected values then get their labels by finding the matching values within `children`, so the result is a list of the selected values based on the order in which they were selected.

I added a new function, `selectedDisplayValue`, based on how `getTooltipValue` handles it that will standardize both of them.

image

✅ Closes: #113

### DIFF
--- a/src/components/SQForm/SQFormMultiSelect.js
+++ b/src/components/SQForm/SQFormMultiSelect.js
@@ -31,16 +31,20 @@ const MenuProps = {
   getContentAnchorEl: null
 };
 
+const selectedDisplayValue = (values, options) => {
+  return values
+    .map(value => {
+      return options.find(option => option.value === value).label;
+    })
+    .join(', ');
+};
+
 const getToolTipTitle = (formikFieldValue, options) => {
   if (!formikFieldValue.length) {
     return 'No value(s) selected';
   }
 
-  return formikFieldValue
-    .map(value => {
-      return options.find(option => option.value === value).label;
-    })
-    .join(', ');
+  return selectedDisplayValue(formikFieldValue, options);
 };
 
 function SQFormMultiSelect({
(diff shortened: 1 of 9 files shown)


caleboleary (approved):

src/components/SQForm/SQFormMultiSelect.js
```diff
@@ -31,16 +31,20 @@ const MenuProps = {
   getContentAnchorEl: null
 };
 
+const selectedDisplayValue = (values, options) => {
+  return values
+    .map(value => {
+      return options.find(option => option.value === value).label;
```
SeanGroff, line 37: Looking at the prop-type it appears `options` could possibly be `undefined`

```javascript
  /** Multiselect options to select from */
  children: PropTypes.arrayOf(
    PropTypes.shape({
      label: PropTypes.string,
      value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
    })
  ),
```

Can we add in optional chaining to guard against `options`, `option.value`, and `option.label` potentially being `undefined`?

SeanGroff (changes requested):

src/components/SQForm/SQFormMultiSelect.js
20BBrown14: This brings up another question: should we allow children to be `undefined`? I think `children`, `option`, `option.label` and `option.value` should be required. Otherwise it's possible to have a select with no options. is that a usecase we want/care about?

codyarose: Okay, I'll test some undefined values and see if I can figure out where it may fail

codyarose:
@SeanGroff 
Sprinkled in some optional chaining that will handle children or the initial value being undefined. 
- if the initial value is undefined we'll get the React/MUI controlled/uncontrolled warning in the console when options are selected
- if `children` is undefined it will render an empty dropdown list, or just a "select all" option if that's enabled, when clicked. If the lone "select all" is checked it still won't pass validation.

20BBrown14 (approved):

SeanGroff (approved):

PR #119, opened 2021-03-11
feat: 🎸added disable option in menuItem for disabling dropdown option

saadurrehman-selectquote:
feat: 🎸 In this feature we can disable dropdown option

In this feature we can disable SQFormdropdown option specifically by
providing isDisable value

✅ Closes: 118

Screenshot:
image

### DIFF
--- a/src/components/SQForm/SQFormDropdown.js
+++ b/src/components/SQForm/SQFormDropdown.js
@@ -79,7 +79,11 @@ function SQFormDropdown({
       >
         {options.map(option => {
           return (
-            <MenuItem key={option.value} value={option.value}>
+            <MenuItem
+              key={option.value}
+              disabled={option.isDisabled}
+              value={option.value}
+            >
               {option.label}
             </MenuItem>
           );


SeanGroff (changes requested):
This is not a breaking change. In the commit it says there is a breaking change. 

When there's no breaking change, during the commit wizard choose N or press enter.

saadurrehman-selectquote:
> This is not a breaking change. In the commit it says there is a breaking change.
> 
> When there's no breaking change, during the commit wizard choose N or press enter.

Done

SeanGroff:
> > This is not a breaking change. In the commit it says there is a breaking change.
> 
> > 
> 
> > When there's no breaking change, during the commit wizard choose N or press enter.
> 
> 
> 
> Done

image

The commit still shows a breaking change?

src/components/SQForm/SQFormDropdown.js
```diff
@@ -79,7 +79,11 @@ function SQFormDropdown({
       >
         {options.map(option => {
           return (
-            <MenuItem key={option.value} value={option.value}>
+            <MenuItem
+              key={option.value}
+              disabled={option.isDisabled}
```
20BBrown14, line 84: We should add this `isDisabled` key to the `children` propTypes shape as optional

saadurrehman-selectquote: This property will be used if we provide isDisbaled in the options
This is not passing as a prop in SQFormDropdown 
example: const MOCK_STATE_OPTIONS = [
  {label: 'Arizona', value: 'AZ', isDisabled: true},
  {label: 'Kansas', value: 'KS',  isDisabled: true},
];

20BBrown14:
> The commit still shows a breaking change?

FWIW I think even if the PR description doesn't show the breaking change the commit still has the breaking change and the bot(s) will see the breaking change when it matters. For instance, if you `git log` this commit the breaking change is still there. I think the whole commit will need to be remade.

src/components/SQForm/SQFormDropdown.js
20BBrown14: It's being passed as an optional part of the children to SQFormDropdown. SQFormDropdown accepts it's dropdown options as children.

saadurrehman-selectquote:
> > The commit still shows a breaking change?
> 
> FWIW I think even if the PR description doesn't show the breaking change the commit still has the breaking change and the bot(s) will see the breaking change when it matters. For instance, if you `git log` this commit the breaking change is still there. I think the whole commit will need to be remade.

Should i decline and create PR again ?

PR #120, opened 2021-03-11
feat: 🎸added disable option in menuItem for disabling dropdown option

saadurrehman-selectquote:
feat: 🎸 In this feature we can disable dropdown option

In this feature we can disable SQFormdropdown option specifically by
providing isDisable value

✅ Closes: 118

screenshot:
image


SeanGroff (approved):
LGTM, thanks for the contribution @saadurrehman-selectquote 💜

codyarose (approved):

Issue #123, opened 2021-03-11
Debounce


aleerobinson:
To prevent dev from having to think about things like debounce every time we use an SQForm can we implement a default of say 500ms across the board on all submits? This would help with QA complaining that the app lets them do things like the image below. Of course we can allow the user to pass in a prop to remove/override the debounce if needed.

![image](https://user-images.githubusercontent.com/43287561/110853614-70cbf300-8279-11eb-83c5-e1e893757cae.png)

SeanGroff:
Thanks @aleerobinson 
While speaking with Adam I believe the SQForm buttons should be debounced by default with a 500ms wait.
I honestly can't think of a situation where we'd want to allow a user to click a form button rapidly so I don't even believe we need to expose a prop to disable the debounce.

With that said, we can always add this prop later as an enhancement if the need arises.

codyarose:
@SeanGroff 
I think I have a solution for this, but wanted to clarify something before making a PR: are you saying to debounce the `onSubmit` on `SQForm` or the `onClick` on `SQFormButton`?

SeanGroff:
> @SeanGroff 
> 
> I think I have a solution for this, but wanted to clarify something before making a PR: are you saying to debounce the `onSubmit` on `SQForm` or the `onClick` on `SQFormButton`?

Debounce any Form button to prevent a user from spam clicking a button.

codyarose:
I'm able to debounce the form onSubmit, but `SQFormIconButton` and `SQFormButton` don't expose an onClick handler to intercept. I may be missing some knowledge concerning how `SQForm`/`Formik` handles these though..

SeanGroff:
> I'm able to debounce the form onSubmit, but `SQFormIconButton` and `SQFormButton` don't expose an onClick handler to intercept. I may be missing some knowledge concerning how `SQForm`/`Formik` handles these though..

Yea I haven't looked but I think the `debounce` may have to occur in the scplus-shared-components library

codyarose:
Okay it looks like `SQFormButton` uses scplus-shared-components `RoundedButton`, which does expose the onClick handler. However, we'd have to debounce the `SQForm` onSubmit and the buttons onClick - if I debounce the onClick in `SQFormButton` the form onSubmit won't be debounced because onSubmit isn't triggered by a submit button's onClick.

codyarose:
[...]

PR #137, opened 2021-03-17
feat: 🎸 add debounce to SQForm onSubmit and buttons onClick

codyarose:
Add [debounced callback](https://www.npmjs.com/package/use-debounce) to the `SQForm` onSubmit and `SQFormButton`/`SQFormIconButton` onClick. For the buttons the debounce is handled within `useFormButton`.

The timer is set to 500ms, as discussed in the issue, and the config is set to allow leading calls, but not trailing. Meaning the event will be fired immediately upon initiation, then can't be fired again until the timer has completed.

Loom: https://www.loom.com/share/e2522c52a8b9422b9c969359a06359c2

✅ Closes: #123

### DIFF
--- a/src/components/SQForm/SQFormButton.js
+++ b/src/components/SQForm/SQFormButton.js
@@ -25,14 +25,20 @@ function SQFormButton({
     return isButtonDisabled;
   }, [dirty, isButtonDisabled, type]);
 
+  const getClickHandler = (...args) => {
+    if (type === 'reset') {
+      return handleReset;
+    } else if (typeof onClick !== 'undefined') {
+      return handleClick(...args);
+    }
+  };
+
   return (
     <RoundedButton
       title={title}
       type={type}
       isDisabled={isSQFormButtonDisabled}
-      onClick={
-        type === 'reset' ? handleReset : onClick ? handleClick : undefined
-      }
+      onClick={getClickHandler}
     >
       {children}
     </RoundedButton>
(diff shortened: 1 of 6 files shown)


src/components/SQForm/SQFormButton.js
```diff
@@ -28,7 +30,9 @@ function SQFormButton({
       title={title}
       type={type}
       isDisabled={isSQFormButtonDisabled}
-      onClick={type === 'reset' ? handleReset : undefined}
+      onClick={
+        type === 'reset' ? handleReset : onClick ? handleClick : undefined
```
20BBrown14, line 34: I'm not the biggest fan of nested ternaries. This one isn't terrible. Would be possible do a `getClickHandler` to make determining the click handler for this button more readable? If others don't think this is necessary that's fine.

codyarose: Yeah, that's fair. I'll add a fix

20BBrown14 (approved):

SeanGroff:
@codyarose Can you provide a Loom video to show the expected functionality within a few stories still functions?
Bonus points if you can show the debounce working.

codyarose:
> @codyarose Can you provide a Loom video to show the expected functionality within a few stories still functions?
> Bonus points if you can show the debounce working.

👍 https://www.loom.com/share/e2522c52a8b9422b9c969359a06359c2

SeanGroff (approved):
Great work @codyarose 
Great idea @aleerobinson

Issue #140, opened 2021-03-17
Common masks object not exported correctly


codyarose:
I goofed on the MASKS object export in `utils` - it currently can't be imported from `@selectquotelabs/sqform`

PR #141, opened 2021-03-18
fix: 🐛 correctly export MASKS from utils

codyarose:
I'm not very familiar with exporting things for use in a component library, so please let me know if there's a better to do it

✅ Closes: #140


lucas-homer:
lgtm! 👺

lucas-homer (approved):

Issue #127, opened 2021-03-15
SQFormDatePicker not showing error message until we click on the datepicker again


saadurrehman-selectquote:
https://www.loom.com/share/1a96b6a0ee35492b8d273dcc45f35266
SQFormDatePicker not showing error message until we click on the datepicker again, as i explained in the loom video

caleboleary:
+1 https://www.loom.com/share/ed06b59c924447e782aa8daf072ff960

PR #143, opened 2021-03-18
fix: 🐛 run date validation on date picker window close

caleboleary:
Date validation was only running on blur of the text element, meaning if
you only used the window, it didn't validate until you focused and
blured the text input or tried to submit. Used setTouched to remedy
this.

✅ Closes: #127

before: https://www.loom.com/share/b8946d66630342f396d11444bac681d9
after: https://www.loom.com/share/e4c1ecd7f8994d25ba53ba5e1518ae8b

There are still some issues here with both the datetimepicker and calendaronlydatepicker components. This did not fix them, I think their issues might be more related to some yup and moment clash but didn't have time to ferret those out in this ticket. I'll try to make an issue demoing both of those issues tomorrow.

### DIFF
--- a/src/components/SQForm/SQFormDatePicker.js
+++ b/src/components/SQForm/SQFormDatePicker.js
@@ -58,6 +58,9 @@ function SQFormDatePicker({
         shouldDisableDate={setDisabledDate}
         value={value}
         onChange={handleChange}
+        onClose={() => {
+          helpers.setTouched();
+        }}
         renderInput={inputProps => {
           return (
             <TextField
--- a/stories/SQFormDatePicker.stories.js
+++ b/stories/SQFormDatePicker.stories.js
@@ -46,3 +46,26 @@ export const basicDatePicker = () => {
     </SQForm>
   );
 };
+
+const schemaBefore2024 = {
+  date: Yup.date()
+    .required()
+    .max(new Date('2024-01-01'), 'Date must be before 2024')
+    .typeError('Invalid date')
+};
+
+export const datePickerBefore2024 = () => {
+  return (
+    <SQForm
+      initialValues={MOCK_INITIAL_STATE}
+      onSubmit={handleSubmit}
+      validationSchema={schemaBefore2024}
+    >
+      <SQFormDatePicker
+        name="date"
+        label="Date"
+        isDisabled={boolean('isDisabled')}
+      />
+    </SQForm>
+  );
+};


codyarose (approved):

saadurrehman-selectquote (approved):
Great fix

src/components/SQForm/SQFormDatePicker.js
```diff
@@ -58,6 +58,9 @@ function SQFormDatePicker({
         shouldDisableDate={setDisabledDate}
         value={value}
         onChange={handleChange}
+        onClose={() => {
+          helpers.setTouched();
```
20BBrown14, line 62: Hm. I'm conflicted on this. If we open the window and don't change anything is the form touched?
Would it make more sense to just run the validation via `validateForm` from FormikContext?

caleboleary: I did try that, but I don't think it'll validate an untouched field (which this will be) - but I also just might not know what I'm doing.

Perhaps there's a smart way to see if value === originalValue, and if different, mark touched?

codyarose: Isn't this sort of how the textfield works? If you tab out of a required textfield without changing anything it's also considered touched, right?

20BBrown14: Is it? I'm not entirely sure. If literally touching a field without changing the value makes a field touched by Formik's standards then I'm totally okay with solution to this issue.

codyarose: I'm not super familiar with Formik, but that seems to be how it works in the SQForm validation story

laurelbean: I actually just completed the Formik tutorial and yes, if you tab out of a field without changing the value it's still considered touched.

20BBrown14: > I actually just completed the Formik tutorial and yes, if you tab out of a field without changing the value it's still considered touched.

Great! :shipit:

20BBrown14 (approved):

lucas-homer (approved):

src/components/SQForm/SQFormDatePicker.js
SeanGroff: A Formik tutorial? Cool!
Touched or dirty is HTML jargon. A field is not considered dirty until a user interacts with the field. Tabbing or touching a field counts as an interaction.

SeanGroff (approved):
Thanks, @caleboleary for figuring this out! 🤘

Issue #144, opened 2021-03-19
SQFormAutocomplete textField does not update when using setFieldValue


laurelbean:
Affected components: SQFormAutocomplete

When trying to set the value for a SQFormAutocomplete based on the selection of a different SQFormAutocomplete, the value for the field gets set, but it doesn't reflect in the visible text field.

Background:  After selecting a Status in the 1st SQFormAutocomplete, there is a corresponding list of Actions that could be taken, which also contain a priority. We wanted to default the Action to the highest priority after selecting a Status. But even though the value is getting set, the SQFormAutocomplete isn't showing it on the screen.

Loom video: https://www.loom.com/share/6bdb1460eb9347e2aa984df2d99ae196

laurelbean:
After talking with @20BBrown14 , it looks like there's 2 different value props: `initialValue` and `inputValue`. `inputValue` is what the underlying TextField component is using to display the selected value, so the solution would be to make sure that `inputValue` gets updated when `ititialValue` is updated as well.

PR #145, opened 2021-03-19
fix: 🐛 setFieldValue now updates inputValue on SQFormAutocomple

laurelbean:
setFieldValue was only updating the initialValue of SQFormAutocomplete
so the visible TextField would not reflect the change. Now that we also set the inputValue, the set option will display.

Loom showing fixed example: https://www.loom.com/share/59e0d88fe6aa470cb01ddf8f9261a5e1

✅ Closes: #144


20BBrown14 (approved):

SeanGroff (approved):
Thanks @laurelbean for your contribution 🚀

Issue #132, opened 2021-03-16
SQFormAutocomplete is always uncontrolled with empty initial value


20BBrown14:
Affected Components: SQFormAutocomplete

It seems like SQFormAutocomplete is *always* uncontrolled if the initial value is empty. It seems if we use `''` as the initial value it sets the autocomplete value to null, thus uncontrolled. 

Loom: https://www.loom.com/share/83b4de4d48a6438ca45b47cac6fa98c2

PR #146, opened 2021-03-22
fix: 🐛 Autocomplete no longer uncontrolled with '' initialValue

20BBrown14:
SQFormAutocomplete used to be uncontrolled if you supplied an empty
string iitialValue. This is not something we would expect. These changes
fix that.

✅ Closes: #132

Loom: https://www.loom.com/share/2ac13968b25241cda1542f7cd29ca79b

### DIFF
--- a/src/components/SQForm/SQFormAutocomplete.js
+++ b/src/components/SQForm/SQFormAutocomplete.js
@@ -14,6 +14,8 @@ import {useForm} from './useForm';
 // MUI uses px, a numeric value is needed for calculations
 const LISTBOX_PADDING = 8; // px
 
+const EMPTY_OPTION = {label: '', value: ''};
+
 const useStyles = makeStyles({
   listbox: {
     '& ul': {
@@ -162,12 +164,12 @@ function SQFormAutocomplete({
         disableListWrap
         classes={classes}
         ListboxComponent={ListboxVirtualizedComponent}
-        options={children}
+        options={[...children, EMPTY_OPTION]}
         onBlur={handleAutocompleteBlur}
         onChange={handleAutocompleteChange}
         onInputChange={handleInputChange}
         inputValue={inputValue}
-        value={initialValue}
+        value={initialValue || EMPTY_OPTION}
         disableClearable={isDisabled}
         freeSolo={isDisabled}
         getOptionLabel={option => option.label}


SeanGroff (changes requested):
@20BBrown14 It looks like a regression was introduced somewhere in our recent versions of SQForm.
We should consider the Synthetic Event warning a bug we need to fix.
Generally, this can be solved by saving the `event` to a variable within an event handler.
I'm guessing somewhere within the `onChange` but I haven't looked through the code.
Can we resolve that warning as part of this PR as well?

20BBrown14:
> @20BBrown14 It looks like a regression was introduced somewhere in our recent versions of SQForm.
> We should consider the Synthetic Event warning a bug we need to fix.
> Generally, this can be solved by saving the `event` to a variable within an event handler.
> I'm guessing somewhere within the `onChange` but I haven't looked through the code.
> Can we resolve that warning as part of this PR as well?

I'm 90% sure this is related to the `action` function from storybook we're providing to the `onInputChange` handler in this specific story. So this is specific to the storybook and not something we'll see in production. I tested this by removing that `action` call in the storybook and the error was gone. Links below

https://github.com/SelectQuoteLabs/SQForm/blob/master/stories/SQForm.stories.js#L169

Loom explaining: https://www.loom.com/share/7b6d72b9660f40acb89fc5ff77cbd8fb

src/components/SQForm/SQFormAutocomplete.js
```diff
@@ -162,12 +165,12 @@ function SQFormAutocomplete({
...
         classes={classes}
         ListboxComponent={ListboxVirtualizedComponent}
-        options={children}
+        options={[...children, EMPTY_OPTION]}
         onBlur={handleAutocompleteBlur}
         onChange={handleAutocompleteChange}
         onInputChange={handleInputChange}
         inputValue={inputValue}
-        value={initialValue}
+        value={initialValue || EMPTY_OPTION}
```
SeanGroff, line 173: [question] - If `displayEmpty` is `false` how come we are adding the empty option to the dropdown options on line 168? Same question on line 173 about falling back to an empty option.

I only expect to see the empty dropdown option when `displayEmpty` is `true`

(I haven't had a chance to watch the loom yet 😬)

20BBrown14: Whoops, I meant to remove `displayEmpty` entirely. This empty option is required to avoid to the uncontrolled/controlled warning beacuse of the way autocomplete works. It *needs* an option for whatever value is provided. If we provide an empty string there *has* to be an option there. So I decided to always include that option.

Validation wise, if this field is required the user will be unable to select this empty option and proceed as an empty string, the value of this empty option, will still fail the validation.

SeanGroff (approved):
LGTM!

Issue #149, opened 2021-03-23
Add missing attributes to components to improve testing


codyarose:
As discussed in #148, some components are missing attributes that would make them more easily queryable by React Testing Library. 

For example, `SQFormTextField` is missing the `id` attribute, which MUI uses to connect the input and its label. If the `id` is added we can use the [`*byLabelText` RTL queries](https://testing-library.com/docs/queries/bylabeltext).

Adding these should also improve user experience

Issue #117, opened 2021-03-09
Add Testing


codyarose:
Gathering information on the why's and how's for adding testing to SQForm. 

Right now our testing is basically us eyeballing components/changes to try to find any possible issues, then if we don't catch them they'll show up in the projects that use this library. Unit and integration tests can help improve confidence that everything is working as intended while we're adding new features and fixing bugs.

I think [`@testing-library/react`](https://testing-library.com/docs/react-testing-library/intro/) would be the best option and [Storybook](https://storybook.js.org/docs/react/workflows/unit-testing) seems to already be set up to work well with it. When writing tests we'll want to avoid testing implementation details whenever possible, so later on if a component needs to be refactored we likely won't need to rewrite the its tests if the end functionality stays the same.

Running the tests can be added to the CI workflow so failing tests will be caught before a PR is merged, and it can likely be added to the `commit` script to prevent failing components/tests from being committed to begin with.

codyarose:
## Scripts
(Need to test these more to make sure I have the args setup correctly)
```json
"scripts": {
  "test": "react-scripts test",
  "test:ci": "CI=true react-scripts test"
  "test:coverage": "npm run test:ci -- --coverage"
},
```
`test` run tests in watch mode
`test:ci` run tests then exit
`test:coverage` run and create a coverage report, then exit

## CI

Example `CI.yml` jobs section with tests; I think it's as simple as this
```yaml
jobs:
  ci:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-node@v2-beta
        with:
          node-version: '14'
      - run: npm ci
      - run: npm run lint
      - run: npm run build
      - run: npm run test # <--
```

## Commit hook
Run tests in commit hook - pre-commit or pre-push
```json
"husky": {
  "hooks": {
    "pre-commit": "CI=true npm run test",
  }
}
```
`CI=true` will force Jest to run in CI-mode, and tests will only run once instead of launching the watcher.

## Filename conventions
According to the [CRA docs](https://create-react-app.dev/docs/running-tests/#filename-conventions):
>Jest will look for test files with any of the following popular naming conventions:
>
>Files with `.js` suffix in `__tests__` folders.
>Files with `.test.js` suffix.
>Files with `.spec.js` suffix.

My personal preference is to collocate the test files as close as possible to the component that it is testing; something like this:
```
├── src
│   ├── components
│   │   ├── ComponentA
[...]

PR #150, opened 2021-03-23
refactor: 💡 add attributes to improve querying in tests

codyarose:
Adding missing attributes to components as I find them while writing the `BasicForm` story test

✅ Closes: #149

### DIFF
--- a/src/components/SQForm/SQFormCheckboxGroup.js
+++ b/src/components/SQForm/SQFormCheckboxGroup.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import InputLabel from '@material-ui/core/InputLabel';
+import FormControl from '@material-ui/core/FormControl';
+import FormLabel from '@material-ui/core/FormLabel';
 import Grid from '@material-ui/core/Grid';
 import FormHelperText from '@material-ui/core/FormHelperText';
 import FormGroup from '@material-ui/core/FormGroup';
@@ -80,13 +81,17 @@ function SQFormCheckboxGroup({
 
   return (
     <Grid item sm={size}>
-      <InputLabel id={groupLabel.toLowerCase()}>{groupLabel}</InputLabel>
-      <FormGroup row={shouldDisplayInRow}>
-        {childrenToCheckboxGroupItems()}
-      </FormGroup>
-      <FormHelperText required={isRequired}>
-        {HelperTextComponent}
-      </FormHelperText>
+      <FormControl component="fieldset">
+        <FormLabel component="legend" classes={{root: 'MuiInputLabel-root'}}>
+          {groupLabel}
+        </FormLabel>
+        <FormGroup row={shouldDisplayInRow}>
+          {childrenToCheckboxGroupItems()}
+        </FormGroup>
+        <FormHelperText required={isRequired}>
+          {HelperTextComponent}
+        </FormHelperText>
+      </FormControl>
     </Grid>
   );
 }
--- a/src/components/SQForm/SQFormReadOnlyField.js
+++ b/src/components/SQForm/SQFormReadOnlyField.js
@@ -12,6 +12,7 @@ function SQFormReadOnlyField({label, name, size = 'auto', muiFieldProps = {}}) {
   return (
     <Grid item sm={size}>
       <TextField
+        id={label.toLowerCase()}
         label={label}
         name={name}
         value={field.value || '- -'}
--- a/src/components/SQForm/SQFormReadOnlyField.js
+++ b/src/components/SQForm/SQFormReadOnlyField.js
@@ -12,7 +12,7 @@ function SQFormReadOnlyField({label, name, size = 'auto', muiFieldProps = {}}) {
   return (
     <Grid item sm={size}>
       <TextField
-        id={label.toLowerCase()}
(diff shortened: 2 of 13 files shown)


src/components/SQForm/SQFormReadOnlyField.js
```diff
@@ -12,6 +12,7 @@ function SQFormReadOnlyField({label, name, size = 'auto', muiFieldProps = {}}) {
   return (
     <Grid item sm={size}>
       <TextField
+        id={label.toLowerCase()}
```
SeanGroff, line 15: If a `label` is "First Name" is `first name` ok for the `id`?
Or do we need to hyphenate it? `first-name`

codyarose: Hmm, good question. I used `label.toLowerCase()` because that's what I saw was being used in other components, such as [`SQFormMultiSelect`](https://github.com/SelectQuoteLabs/SQForm/blob/master/src/components/SQForm/SQFormMultiSelect.js#L91).
I think a spaced string is technically valid, but something like kebab-case or camelCase would probably be better in terms of "best practices"

src/components/SQForm/SQFormCheckboxGroup.js
```diff
@@ -80,13 +81,17 @@ function SQFormCheckboxGroup({
...
+        <FormLabel component="legend" classes={{root: 'MuiInputLabel-root'}}>
+          {groupLabel}
+        </FormLabel>
+        <FormGroup row={shouldDisplayInRow}>
+          {childrenToCheckboxGroupItems()}
+        </FormGroup>
+        <FormHelperText required={isRequired}>
+          {HelperTextComponent}
+        </FormHelperText>
+      </FormControl>
```
codyarose, line 94: Changing `SQFormCheckboxGroup` and `SQFormRadioButtonGroup` to use `FormControl` and `FormLabel` instead of `InputLabel` will group the checkbox/radios together and allow us to query them by group, e.g., `screen.getByRole('group', { name: /field group name/i })`.

When writing tests this will help narrow down our queries when there are multiple inputs with the same name or value. 

A contrived example: a form may have three different radio groups where the options are 'yes' or 'no', with this structure we can confidently query an option within a specific group.
```js
// Narrowing it down to the group then picking the radio within that group
userEvent.click(
    within(screen.getByRole('group', { name: /group three/i })).getByRole('radio', { name: /yes/i })
)

// vs

// Querying all the radio buttons on the page by name then trying to find the right one within that array. 
userEvent.click(screen.getAllByRole('radio', { name: /yes/i })[2])
// The big issue here is that if another yes/no radio group is inserted in the dom ahead of the one we're trying to target,
// our intended target's place in the array will change and we'll now be clicking the wrong radio
```

This may not be all that useful when writing tests for these components, but if we expand testing to the other projects that use this library it should make writing tests for those much easier

src/components/SQForm/SQFormReadOnlyField.js
```diff
@@ -12,6 +12,7 @@ function SQFormReadOnlyField({label, name, size = 'auto', muiFieldProps = {}}) {
   return (
     <Grid item sm={size}>
       <TextField
+        id={label.toLowerCase().replaceAll(' ', '-')}
```
SeanGroff, line 15: I believe this is going to be a common string transformation.
Probably worth creating a helper util function

Not sure of the best name but something like 

`id={toLowerCaseAndHyphenate(label)}`

codyarose: `toKebabCase(label)`? I think kebab-case is traditionally all lower case

SeanGroff:
I think this should be a `fix` commit so that these component changes get distributed to npm.
For example, if our QA automated testing team wanted to leverage these query selectors tomorrow, they wouldn't be able to because there's no new version of the code.

codyarose:
> I think this should be a `fix` commit so that these component changes get distributed to npm.
> For example, if our QA automated testing team wanted to leverage these query selectors tomorrow, they wouldn't be able to because there's no new version of the code.

Ah okay. Adding a new `fix` commit would trigger that, right?

SeanGroff:
> I think this should be a `fix` commit so that these component changes get distributed to npm.
> For example, if our QA automated testing team wanted to leverage these query selectors tomorrow, they wouldn't be able to because there's no new version of the code.



> > I think this should be a `fix` commit so that these component changes get distributed to npm.
> > For example, if our QA automated testing team wanted to leverage these query selectors tomorrow, they wouldn't be able to because there's no new version of the code.
> 
> Ah okay. Adding a new `fix` commit would trigger that, right?

Yep!

Chris-Boyle (approved):

Chris-Boyle:
This is interesting stuff.  I would really like to learn more about the plan/direction we are going in for testing this project.  Maybe a team meeting would be a good place to outline the vision.

codyarose:
> This is interesting stuff. I would really like to learn more about the plan/direction we are going in for testing this project. Maybe a team meeting would be a good place to outline the vision.

I've outlined some things in #148 and #117, but there are certainly more things to discuss

lucas-homer (approved):
looks great to me! nice work @codyarose

SeanGroff (approved):
LGTM nice work!!

Issue #153, opened 2021-03-25
Disabled components should not have helper text


20BBrown14:
Affected components: All of them

Disabled components should not have helper text/icons that indicate validation state below them. Consumers should take care that disabled fields do not violate any validation logic. Having validation helper text/icons could be confusing a user as they're unable to interact with that field.

There's also cases where we'd like to use a disabled multiselect, for instance, to get the tooltip and string display logic without having to rewrite it for a readonly field. As it stands a disabled multiselect will include the helper text which in a field of other readonly fields, for instance, is distracting and bad UI.

This should be an easy change in most components. if the `isDisabled` prop is true we should not render the `HelperText` component.

PR #156, opened 2021-03-25
fix: 🐛 HelperText no longer shows on disabled components

20BBrown14:
The HelperText / icon was showing on disabled components. It's not
longer doing that.

✅ Closes: #153

Loom: https://www.loom.com/share/62dcfcf886be4547ad7500b4d5ba8597


codyarose (approved):

stories/SQFormDisableableComponents.stories.js
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
```
SeanGroff, line 1: I like that you even created a story to show the components that can be disabled

SeanGroff (approved):
Nice catch, thanks for fixing this @20BBrown14!

stories/SQFormDisableableComponents.stories.js
20BBrown14: Say "Disableable" 10 times fast

SeanGroff: > Say "Disableable" 10 times fast

I can't even say it once

Issue #160, opened 2021-03-26
Refactor components to use `FormControl`


codyarose:
[`FormControl`](https://material-ui.com/api/form-control/#formcontrol-api) can automatically pass error, disabled, and required state down to the label, input, and helper text, rather than specifying those props on each individual component.

https://www.loom.com/share/dc737dc31af34ba3b1b6726090caf908

PR #163, opened 2021-03-27
fix: 🐛 change components to use `FormControl` for state props

codyarose:
Inputs, labels, and helper text will get error, disabled, and required
state from `FormControl`. Reduces duplication, functionality remains the
same

https://www.loom.com/share/bf0424fdf5e5423fa65d9234f73c3b1d

✅ Closes: #160

### DIFF
--- a/src/components/SQForm/SQFormCheckboxGroup.js
+++ b/src/components/SQForm/SQFormCheckboxGroup.js
@@ -81,16 +81,20 @@ function SQFormCheckboxGroup({
 
   return (
     <Grid item sm={size}>
-      <FormControl component="fieldset">
-        <FormLabel component="legend" classes={{root: 'MuiInputLabel-root'}}>
+      <FormControl component="fieldset" required={isRequired}>
+        <FormLabel
+          component="legend"
+          classes={{
+            root: 'MuiInputLabel-root',
+            asterisk: 'MuiInputLabel-asterisk'
+          }}
+        >
           {groupLabel}
         </FormLabel>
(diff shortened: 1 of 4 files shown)


SeanGroff:
@codyarose A wild merge conflict appears 👻

codyarose:
@SeanGroff Should be good now

20BBrown14 (approved):

SeanGroff:
@codyarose Can you take a look at the questions I had around the Asterix classes?
I wasn't sure what those are doing. I know Nemeth doesn't want to render an asterisk for required fields.

codyarose:
> @codyarose Can you take a look at the questions I had around the Asterix classes?
> I wasn't sure what those are doing. I know Nemeth doesn't want to render an asterisk for required fields.

Oh, I can't find where those questions are, but the classes are there to make the asterisk not render. For some reason `FormLabel` makes them appear while `InputLabel` does not, so the class makes the asterisk behave the way it does in `InputLabel`.

SeanGroff (commented):
> > @codyarose Can you take a look at the questions I had around the Asterix classes?
> > I wasn't sure what those are doing. I know Nemeth doesn't want to render an asterisk for required fields.
> 
> Oh, I can't find where those questions are, but the classes are there to make the asterisk not render. For some reason `FormLabel` makes them appear while `InputLabel` does not, so the class makes the asterisk behave the way it does in `InputLabel`.

TIL, good to know!

SeanGroff (approved):
Awesome sauce!

Issue #165, opened 2021-03-29
SQFormDialog not showing errors on mount


abbas-ali123:
I am currently working on SQFormDialog and need to display the error message when the form is open. Currently, the error is displayed when it is blurred. Also, when the field is corrected but the changes are not saved and the form is reopened it is showing the green valid icon below the text field. And again on blurring it shows the error message.
Below is the loom describing the issue:
https://www.loom.com/share/d795583fd21447f3be0921bb199698c5

saadurrehman-selectquote:
+1 i double checked that, its happening

saadurrehman-selectquote:
incase of loading invalid data from DB, it doesn't show error until we do onBlur, 
its happening in case of DatePicker and TextField.

Issue #147, opened 2021-03-22
SQFormDateTimePicker && SQFormDatePickerWithCalendarInputOnly not validating on date selection (with window picker)


caleboleary:
Same issue as https://github.com/SelectQuoteLabs/SQForm/issues/127

When selecting an invalid date (defined by the yup schema) using only the datepicker (the only option in the `SQFormDatePickerWithCalendarInputOnly` variant) - error messages are not showing on change, only when you interact with the field a second time.

The same fix proposed in this https://github.com/SelectQuoteLabs/SQForm/pull/143/files PR didn't seem to fix the problem (though its probably a good start in the case of `SQFormDateTimePicker`).

I did a little more digging and wondered if it might be something to do with comparing a string of a date vs a moment date object or something, but I could be totally off base with that.

Here's a loom where I show both of them failing to validate and say a lot of this again: https://www.loom.com/share/427a01cf8c0546f4ac96182fb5eac068

codyarose:
I think the fix from that PR on `SQFormDatePicker` isn't working because the `onClose` is being overwritten in the other components. For example, in `SQFormDatePickerWithCalendarInputOnly` a different `onClose` is being passed through `muiFieldProps`, so the `helpers.setTouched()` in `SQFormDatePicker` will never run.

PR #171, opened 2021-03-31
fix: 🐛 correct error state when initial values provided

### LINKED ISSUE #89: Different validation errors onChange and onBlur
I've noticed a weird thing with some validation. If a field is required AND has a specific answer required to be valid the UI doesn't update with onChange properly. Instead when it blurs it shows the proper validation message. I've created a branch that recreates the issue in Storybook. Also there's a loom.

Branch: https://github.com/SelectQuoteLabs/SQForm/compare/form_validation_showcase

Loom: https://www.loom.com/share/e40a450e6da3459a8828b83086bdf0bf

codyarose:
Validation errors aren't showing on mount because `isFieldError` uses
`isTouched`. The field won't be considered touched until the user clicks
it, which is why we're seeing the current behavior. If we use `hasValue`
it will show those errors when the field has an initial value.

https://www.loom.com/share/2049ee6738ee477dabb9df1bce9d5596

Also fixes #89: https://www.loom.com/share/1b49b2a0125d4fd8b082097b31c950c0

✅ Closes: #165, #89, #147

### DIFF
--- a/src/components/SQForm/SQFormDatePicker.js
+++ b/src/components/SQForm/SQFormDatePicker.js
@@ -58,9 +58,6 @@ function SQFormDatePicker({
         shouldDisableDate={setDisabledDate}
         value={value}
         onChange={handleChange}
-        onClose={() => {
-          helpers.setTouched();
-        }}
         renderInput={inputProps => {
           return (
             <TextField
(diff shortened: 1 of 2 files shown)


codyarose:
> Nice fix! 
> Would this also be a good candidate for a test in the future? 
> 
> Test to ensure proper error State on initial load?

Yes, we should create a method for testing the error state that can be reused across all the component tests to ensure that their behavior is consistent.

codyarose:
> It's been a hot minute since I looked at it but I think this might fix #89 as well.

Yep, confirmed that this will solve that issue as well

SeanGroff (approved):
Those closed issues!!
Great fix @codyarose 🚀 🚀 🚀

PR #178, opened 2021-04-02
✏️  Docs/restructure storybook

### LINKED ISSUE #164: Unable to scroll vertically in Storybook
While creating a new story I noticed that when in the "Docs" tab in storybook you can't scroll vertically down to the table. This would be super helpful to be able to do because, with the storybook docs addon, in that table we can have all of those sweet, sweet component prop details.

https://www.loom.com/share/7f6378cb474845918fce8c098bd7d789

I've narrowed down the problem to an `overflow: hidden;` on the `body` tag that I _think_ is coming from `scplus-shared-components` somehow. On the left in vscode is a css file in `scplus-shared-components` and the right is the elements pane for `SQForm`s storybook docs tab.

image
_looks familiar 🤔_

I couldn't find any mention of those styles in `SQForm`, so I'm guessing it's coming from the `muiTheme` imported from `scplus-shared-components`?

codyarose:
Restructured our Storybook to match the current standards and make it overall easier to use and more useful as documentation. 

- fixed body overflow issue #164 
- `addons-notes` and `addon-info` have been [deprecated](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#60-deprecations) in favor of `addon-docs`
    - removed those and migrated any use of them to the `addon-docs` format
- added `addon-controls`, the successor to `addon-knobs`
    - `addon-knobs` will be [deprecated](https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#how-will-this-replace-addon-knobs) in 7.0; I left it installed for now until we migrate the few areas we use it to `controls`
- refactored [deprecated](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-addparameters-and-adddecorator) `addParameters` and `addDecorators` to new syntax
- moved theme handling from parameters object to [`manager.js` using `addons.setConfig`](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-immutable-options-parameters)
- added `createDocsPage` as a way to create auto-generated docs for the components in a consistent format

https://www.loom.com/share/1e60a4e7d8224cdd8fd733f2e56c59ad

Fixes #164


Chris-Boyle:
@codyarose this is all great stuff!  We should probably move some of the components to the Components section and not have them live under Forms.  The only true Forms we have are SQDialogForm, SQDialogStepper, and SQForm.  I think the rest of the SQForm components should live under the Forms header since they are not true forms.

.storybook/preview.js
```diff
@@ -14,30 +13,39 @@ const centerStyle = {
   justifyContent: 'center',
   alignItems: 'center',
   height: '100%',
-  padding: '15rem'
+  padding: '3rem'
```
hvilander, line 16: Question, I know internally we try to stick to pixels. What is the reason we are using rem here?

.storybook/storybook.css
```diff
@@ -0,0 +1,4 @@
+.sb-show-main {
+  /* overrides `overflow: hidden;` on body */
+  overflow: visible;
+}
```
hvilander, line 4: OMG THANK YOU

hvilander: do we need to keep the commented out line?

hvilander (approved):

.storybook/preview.js
SeanGroff: Back in the day everything was in REM and I'm not really sure why. Assuming Cody used REM because that's what the original code was using. 

We can use px here.

codyarose: Yeah, I just stuck with the format that was already there - I'll switch it to px.

Also, I meant to make a note that I decreased the padding because `15rem` was making the stories in the canvas blocks gigantic in the docs tab.

.storybook/storybook.css
codyarose: I wouldn't say that we _need_ to keep it, but I added it to give some context since it's the only thing in the whole file and the class name doesn't really give us much info about what it's for either

hvilander: I did not have enough coffee in my system when reading. I agree this is good context to have.

codyarose:
> @codyarose this is all great stuff! We should probably move some of the components to the Components section and not have them live under Forms. The only true **Forms** we have are SQDialogForm, SQDialogStepper, and SQForm. I think the rest of the SQForm components should live under the **Components** header since they are not true forms.

Yeah, I'm planning on going through those as more component stories are added because they'll need to be refactored to match the new storybook format, so it's more of a temporary holding area for now

Chris-Boyle:
> > @codyarose this is all great stuff! We should probably move some of the components to the Components section and not have them live under Forms. The only true **Forms** we have are SQDialogForm, SQDialogStepper, and SQForm. I think the rest of the SQForm components should live under the **Components** header since they are not true forms.
> 
> Yeah, I'm planning on going through those as more component stories are added because they'll need to be refactored to match the new storybook format, so it's more of a temporary holding area for now

Oh I thought these changes were dropping the stories into the Forms section instead of the Components section.  
image

codyarose:
> Oh I thought these changes were dropping the stories into the Forms section instead of the Components section.

It is, but I see what you're saying. Some of those in the Forms section really are just components; I'll move those over to the components section now then refactor them to the new format in a separate PR

Chris-Boyle (approved):

SeanGroff (approved):
Much awesome @codyarose 
Thank you for taking the time to do all of this!!

Issue #168, opened 2021-03-29
Add more tests - SQFormDropdown


laurelbean:
Now that we've gotten a base testing setup, we should start covering each component.
This issue will focus on covering SQFormDropdown.

PR #180, opened 2021-04-02
Add tests for SQFormDropdown

laurelbean:
Added a small suite of tests all relating to SQFormDropdown. I tried to stick with the concept of only testing what the user would do instead of implementation details.

This will close out #168


src/utils/tests.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import {render} from '@testing-library/react';
+import SQForm from '../components/SQForm/SQForm';
+
+//export const renderSQForm = (ui, initialValues, options = {}) => {
```
hvilander, line 5: nit: looks like this comment can be removed

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+  userEvent.tab();
+  expect(expandButton).toHaveFocus();
+
+  userEvent.tab();
+  expect(expandButton).not.toHaveFocus();
+
+  const required = screen.getByText(/required/i);
+  expect(required).toBeVisible();
+  //expect required to be orange / highlighted
+  //expect(required).toHaveAttribute('class', 'MuiFormHelperText-root Mui-error Mui-required');
```
hvilander, line 201: nit: I suspect these comments can be removed

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+
+  renderSQForm(renderDropdown, DEFAULT_INTITIAL_VALUES);
+
+  const expandButton = screen.getByRole('button', /dropdown example/i);
+  userEvent.click(expandButton);
+
+  const option = screen.getByText('Test 1');
+  userEvent.click(option);
+
+  expect(handleChange).toHaveBeenCalled();
```
Chris-Boyle, line 76: Do we need to have two expects here.  I think the one that specifies how many times it was called is sufficient. We can remove the one on line 76.

src/components/SQForm/__tests__/SQFormDropdown.test.js
hvilander: or perhaps they are the part of what you would like to be testing?

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+    ...DEFAULT_PROPS,
+    isDisabled: false
+  };
+
+  const renderDropdown = createDropdown(DEFAULT_OPTIONS, fieldProps);
+
+  renderSQForm(renderDropdown, DEFAULT_INTITIAL_VALUES);
+
+  const expandButton = screen.getByRole('button', /dropdown example/i);
+  expect(expandButton).not.toHaveAttribute('aria-disabled', 'true');
```
hvilander, line 146: If the expand button was not found, would this test be a false positive?

src/components/SQForm/__tests__/SQFormDropdown.test.js
laurelbean: Yes, this is the last piece I think I need to get a whole test. I haven't been able to find a working solution to check that the field is in an error state / highlighted. The only thing I could see that changed was adding `.Mui-error`.

src/components/SQForm/__tests__/SQFormDropdown.test.js
laurelbean: I believe `getByRole()` will fail / error out if it doesn't find it. I'll double check that.

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+
+  renderSQForm(renderDropdown, DEFAULT_INTITIAL_VALUES);
+
+  const expandButton = screen.getByRole('button', /dropdown example/i);
+  userEvent.click(expandButton);
+
+  const optionsList = screen.getByRole('listbox', /dropdown example/i);
+
+  expect(optionsList).toBeInTheDocument();
+  expect(optionsList.children.length).toBe(3);
```
codyarose, line 43: **suggestion**(non-blocking): This could utilize the `toHaveLength()` matcher:
`expect(optionsList.children).toHaveLength(3);`
or even more explicitly:
`expect(within(optionsList).getAllByRole('option')).toHaveLength(DEFAULT_OPTIONS.length);`

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+  }
+};
+
+const DEFAULT_OPTIONS = [
+  {label: 'Test 1', value: 1},
+  {label: 'Test 2', value: 2},
+  {label: 'Test 3', value: 3}
+];
+
+const createDropdown = (
```
SeanGroff, line 25: Nitpicky so I apologize.

Can we call this function `renderDropdown`?

This is a common pattern in React and it more accurately reflects what this function is doing.

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
+import React from 'react';
+import * as Yup from 'yup';
+import {screen} from '@testing-library/react';
+import SQFormDropdown from '../SQFormDropdown';
+import {renderSQForm} from '../../../utils/tests';
+import userEvent from '@testing-library/user-event';
```
20BBrown14, line 6: minor: reoder imports so modules are at the tope

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+
+const createDropdown = (
+  options = DEFAULT_OPTIONS,
+  fieldProps = DEFAULT_PROPS
+) => {
+  return <SQFormDropdown {...fieldProps}>{options}</SQFormDropdown>;
+};
+
+it('should render list of options', () => {
+  const renderDropdown = createDropdown();
```
20BBrown14, line 33: nit: can we make this `renderedDropdown` instead?

SeanGroff, line 33: Related to my above comment, the naming suggestion would also help here:

`const sqFormDropdown = renderDropdown()`

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+
+  renderSQForm(renderDropdown, DEFAULT_INTITIAL_VALUES);
```
SeanGroff, line 35: Love the choice to create the abstraction for renderSQForm 💜

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+
+  const expandButton = screen.getByRole('button', /dropdown example/i);
+  userEvent.click(expandButton);
+
+  const option = screen.getByText('Test 1');
+  userEvent.click(option);
+
+  expect(handleChange).toHaveBeenCalled();
+  expect(handleChange).toHaveBeenCalledTimes(1);
```
20BBrown14, line 77: Should we be checking what this is called with? Sometimes it's appropriate not sure if it is here, though.

20BBrown14 (approved):

20BBrown14:
Didn't mean to approve. Oops. Doesn't look like there's a way I can undo it. Either way, great work so far here! I love it.

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+
+  const labelValue = screen.getByRole('button', /dropdown example/i);
+  expect(labelValue).toHaveTextContent('Test 1');
```
codyarose, line 80: **suggestion**: A value check before making the change would give more confidence that this test is working as intended.
e.g. `expect(labelValue).not.toHaveTextContent('Test 1')`

Also **nitpick**: I think we should use something like `DEFAULT_OPTIONS[0].label` for these values because if later on someone reorders/changes that object it'll cause this test to fail when it really shouldn't

src/components/SQForm/__tests__/SQFormDropdown.test.js
SeanGroff: > Do we need to have two expects here. I think the one that specifies how many times it was called is sufficient. We can remove the one on line 76.

+1

Question - `expect(handleChange).toHaveBeenCalled();` Reads really well. Are we concerned with how many times it's been called? If there was a bug I guess it guards against that possibility? 

If we have to say `expect(handleChange).toHaveBeenCalledTimes(1);` it's hard for me to picture when `expect(handleChange).toHaveBeenCalled();` would ever be used?

Question 2 - Where do we test what args `handleChange` is called with? Meaning, how do we ensure it's being called with the expected arguments? Or do we care?

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+  const renderDropdown = createDropdown(options, DEFAULT_PROPS);
+
+  renderSQForm(renderDropdown, DEFAULT_INTITIAL_VALUES);
+
+  const expandButton = screen.getByRole('button', /dropdown example/i);
+  userEvent.click(expandButton);
+
+  const optionsList = screen.getByRole('listbox', /dropdown example/i);
+
+  expect(optionsList.childNodes[1]).toHaveAttribute('aria-disabled', 'true');
```
codyarose, line 99: **suggestion**: I think using `childNodes` relies a bit too much on implementation details - this would fail if the markup changed in the future even though the functionality would stay the same.
A more future-proof alternative: 
```js
expect(within(optionsList).getAllByRole('option')[1]).toHaveAttribute(
  'aria-disabled',
  'true'
);
```

SeanGroff, line 99: Question - How do we know what to check for to ensure the child node is disabled and that it gets properly disabled?

An `aria` property ensures accessibility.
A CSS class should appear to ensure it's styled properly
An HTML property `disabled` should appear to ensure it cannot be clicked on.

Which of these is preferred to test against to ensure the element is actually being disabled?
I'm learning on the fly so I apologize if this question is a stupid one.

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+    ...DEFAULT_PROPS,
+    isDisabled: true
+  };
+
+  const renderDropdown = createDropdown(DEFAULT_OPTIONS, fieldProps);
+
+  renderSQForm(renderDropdown, DEFAULT_INTITIAL_VALUES);
+
+  const expandButton = screen.getByRole('button', /dropdown example/i);
+  expect(expandButton).toHaveAttribute('aria-disabled', 'true');
```
SeanGroff, line 132: Does the presence of `aria-disabled` ensure the element is not clickable?

My nooby brain expected something like `expect(handleChange).toHaveBeenCalledTimes(0);`

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,202 @@
...
+
+  renderSQForm(renderDropdown, DEFAULT_INTITIAL_VALUES);
+
+  const expandButton = screen.getByRole('button', /dropdown example/i);
+  userEvent.click(expandButton);
+
+  const optionsList = screen.getByRole('listbox', /dropdown example/i);
+
+  expect(optionsList.children.length).toBe(4);
+  expect(optionsList.childNodes[0]).toHaveTextContent('- -');
```
codyarose, line 118: **suggestion**: 
```js
const options = within(optionsList).getAllByRole('option');
expect(options).toHaveLength(4);
expect(options[0]).toHaveTextContent('- -');
```

stories/__tests__/SQForm.stories.test.js
```diff
@@ -77,3 +83,41 @@ it('renders BasicForm and calls alert on submit', async () => {
...
+          hobby: '',
+          cool: false,
+          lame: false
+        },
+        null,
+        2
+      )
+    )
+  );
+});
```
SeanGroff, line 123: This test was super easy to follow along with. 💯

src/components/SQForm/__tests__/SQFormDropdown.test.js
codyarose: Yeah, that's how I tested it: `.toHaveClass(Mui-error)`

codyarose (changes requested):

src/components/SQForm/__tests__/SQFormDropdown.test.js
codyarose: Personally, I would just test whatever `handleChange` does rather than testing that it was called. If all `handleChange` does is change the selected value then we could just test that and leave out the `toHaveBeenCalled`

src/components/SQForm/__tests__/SQFormDropdown.test.js
codyarose: Yeah, I think this will be really helpful 👍

src/components/SQForm/__tests__/SQFormDropdown.test.js
Chris-Boyle: I do like the `toHaveBeenCalled` with the precision of how many times to cover scenarios where it get's triggered more than once by accident, a bug in the code, or a re-render.

codyarose:
I think in the future we may switch to testing the component stories rather than the component file, but if we do that there shouldn't much to change here to make that work

SeanGroff:
> I think in the future we may switch to testing the component stories rather than the component file, but if we do that there shouldn't much to change here to make that work

Agreed. 

@laurelbean As part of this draft, would you mind moving the test, to test against the story itself? https://github.com/SelectQuoteLabs/SQForm/blob/master/stories/SQFormDropdown.stories.js

src/components/SQForm/__tests__/SQFormDropdown.test.js
SeanGroff: > Personally, I would just test whatever `handleChange` does rather than testing that it was called. If all `handleChange` does is change the selected value then we could just test that and leave out the `toHaveBeenCalled`

I think this is what i'm looking for. `toHavebeenCalled` or the number of times it's been called doesn't give me much confidence the function is working properly. 

Testing what the function should do, does give me confidence the function is working as expected.

src/components/SQForm/__tests__/SQFormDropdown.test.js
codyarose: I think this is an issue with Mui. Seems like they just add `aria-disabled` and a `Mui-disabled` class, no `disabled` attribute.
Maybe testing that when clicked it doesn't reveal the options would suffice?

codyarose:
> @laurelbean As part of this draft, would you mind moving the test, to test against the story itself? https://github.com/SelectQuoteLabs/SQForm/blob/master/stories/SQFormDropdown.stories.js

You could give [this package](https://github.com/storybookjs/testing-react) a shot as well. I haven't tried it out yet, so it may or may not be the right choice for us

src/components/SQForm/__tests__/SQFormDropdown.test.js
SeanGroff: > Maybe testing that when clicked it doesn't reveal the options would suffice?

What do you have in mind for how to write this test? Essentially, when clicking the disabled option, we expect nothing to happen.

codyarose: > What do you have in mind for how to write this test? Essentially, when clicking the disabled option, we expect nothing to happen.

My though process is:
- expect options list `.not.toBeInTheDocument` or `.not.toBeVisible`
- click disabled input
- expect options list still `.not.toBeInTheDocument` or `.not.toBeVisible`

That sounds clunky, but it does match how the user should experience it. Though there may be a better way to write that

src/components/SQForm/__tests__/SQFormDropdown.test.js
laurelbean: So what's going wrong is that the `Mui-error` class isn't being applied. So the test fails because it's not there. So I'm somehow not triggering the error state. But I wrote it how I made it happen testing manually. So that's where I'm stuck.

src/components/SQForm/__tests__/SQFormDropdown.test.js
laurelbean: So right under this I have
`const labelValue = screen.getByRole('button', /dropdown example/i);
  expect(labelValue).toHaveTextContent('Test 1');`
  to test that it actually changes the value.

I can remove these expects all together if we think they're not useful.

src/components/SQForm/__tests__/SQFormDropdown.test.js
codyarose: Oh, I think it may be because `validationSchema` should be passed to SQForm instead of the SQFormDropdown

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -0,0 +1,240 @@
...
+
+  const disabledOption = within(optionsList).getAllByRole('option')[1];
+  expect(disabledOption).toHaveAttribute('aria-disabled', 'true');
+
+  /*
+    This actually still fires the event! So everything after this fails
+    I found this when trying to research why:
+    https://github.com/testing-library/dom-testing-library/issues/92#issuecomment-530524324
+  
+    userEvent.click(disabledOption);
```
laurelbean, line 115: For some reason it's still letting me click the disabled option.

laurelbean:
I've updated most of the tests based on the comments/discussions.

src/components/SQForm/__tests__/SQFormDropdown.test.js
codyarose: Here's a possible solution
```js
const expandButton = screen.getByRole('button', {name: EMPTY_OPTION});

// Disabled option is skipped
userEvent.type(expandButton, '{tab}{enter}{arrowdown}{arrowdown}{enter}');
const updatedExpandButton = await screen.findByRole('button', {
  name: /test 3/i
});

expect(updatedExpandButton).toBeInTheDocument();
```

codyarose: Some context on why that's async/await:
When we update the value of that component a state change is triggered in Formik. Normally you'd get a warning saying you need to wrap the action that changes the state in `act`, but we can avoid doing that if we `await` the change in state that we're expecting

laurelbean:
Added a few more test cases and tried to make both the tests against the component and the tests against the story as close as possible.

I prefer the tests against the component because I have full control over the test data so a wider variety of tests can happen.

src/components/SQForm/__tests__/SQFormDropdown.test.js
```diff
@@ -238,3 +252,26 @@ it('should highlight field if required but no value selected', () => {
...
+
+  renderSQForm(nullSQFormDropdown, DEFAULT_INTITIAL_VALUES);
+
+  const expandButton = screen.getByRole('button', {name: EMPTY_OPTION});
+  userEvent.click(expandButton);
+
+  expect(screen.queryByRole('option')).not.toBeInTheDocument();
+});
+
+it('should show empty value if initial value not in options', () => {
```
codyarose, line 267: These tests are looking great!

For this one you can suppress that noisy console warning with a spy:
```js
// mock console.warn
const consoleWarnSpy = jest
  .spyOn(console, 'warn')
  .mockImplementation(() => {});

const initialValues = {
  dropdownExample: 5
};

renderSQForm(renderDropdown(), {initialValues});

const expandButton = screen.queryByRole('button', {name: EMPTY_OPTION});

expect(expandButton).not.toBeInTheDocument();
// check that the expected warning was called
expect(consoleWarnSpy).toHaveBeenCalledWith(
  expect.stringMatching(/requested display value.*could not be found/i)
);
// restore the mock to prevent it leaking into subsequent tests
consoleWarnSpy.mockRestore();
```

This way we can check that the test passes and the appropriate console warning was given while avoiding the clutter in the terminal output

stories/__tests__/SQFormDropdown.stories.test.js
```diff
@@ -23,6 +23,12 @@ it('should render with empty initial value', () => {
   expect(labelValue).toHaveTextContent('- -');
 });
 
+/* There's not a way for me to test non-empty intiial values
```
codyarose, line 26: This is a limitation caused by how we have to construct the stories. The component has to be wrapped in SQForm for it to work, and the storybooks args have to go on the target component for the storybook stuff to work. That means we don't really have a straight forward way to pass anything to the component's parent (SQForm).

cc @SeanGroff 

We could probably pass in a custom arg like `args.SQFormInitialValues` to SQForm's `initialValues` prop, but I feel like that's going to start complicating the stories because we're trying to make them do something they're not really meant to do. If we did this there may also be the overhead of keeping track of which storybook args are meant for the component and which are meant for SQForm for the purpose of testing.

I think some more discovery is needed to figure out if testing stories is really the right way to go for this library

stories/__tests__/SQFormDropdown.stories.test.js
```diff
@@ -138,3 +151,27 @@ it('should highlight field if required but no value selected', () => {
   expect(required).toBeVisible();
   expect(required).toHaveClass('Mui-error');
 });
+
+/* Can't test these as there's no way to pass in a different options list
```
codyarose, line 155: I haven't been able to figure out a way to make this work yet. This would require passing the options array to the `children` prop in the story, but for some reason that always gives an error saying `children` is undefined

stories/__tests__/SQFormDropdown.stories.test.js
SeanGroff: I don't have my head fully wrapped around this BUT my fire from the hip reaction is we need to supplement our integration tests with Cypress e2e tests in Storybook. 

Or maybe it's the inverse, we create Cypress e2e tests and supplement the e2e tests with these integration tests.

src/components/SQForm/__tests__/SQFormDropdown.test.js
laurelbean: Nice! Thanks for this. Added to this test and the other one that also has a console.warn message.

stories/SQFormDropdown.stories.js
```diff
@@ -37,7 +37,10 @@ const defaultArgs = {
 
 export const SQFormDropdown = args => (
   <div style={{minWidth: 250}}>
-    <SQFormStoryWrapper initialValues={{[defaultArgs.name]: ''}}>
+    <SQFormStoryWrapper
+      initialValues={{[defaultArgs.name]: ''}}
+      validationSchema={args.validationSchema}
```
codyarose, line 42: This should be destructured from `args`, otherwise it will be passed into the dropdown component on line 47. So
```js
const {validationSchema, ...dropdownProps} = args;
```
then you can spread `dropdownProps` on the dropdown component instead of `args`

laurelbean:
Added @storybook/testing-react library and converted to use composeStories instead of the story directly. Merged in the work from @codyarose that allowed me to send props to the SQForm wrapper. So now both the component and story files have the same tests running against them.

laurelbean:
The dropdown should behave the same in SQForm and SQFormDialog. Should we make an SQFormDialog story under the SQFormDropdown.stories and create similar tests or are we going to handle the Dialog tests separately?

codyarose:
> The dropdown should behave the same in SQForm and SQFormDialog. Should we make an SQFormDialog story under the SQFormDropdown.stories and create similar tests or are we going to handle the Dialog tests separately?

The dialog tests should be separate. My thought process is in `SQFormDropdown.stories.test` we confirm that the dropdown component works properly in a form. Then if for some reason there's an issue with it when used in `SQFormDialog`, that's _probably_ an issue with `SQFormDialog` and should be confirmed in the dialog tests.

stories/__tests__/SQFormDropdown.stories.test.js
```diff
@@ -0,0 +1,191 @@
...
+it('should render with empty initial value', () => {
+  render(<SQFormDropdown size="auto" />);
+
+  const labelValue = screen.getByRole('button', {name: /state/i});
+  expect(labelValue).toHaveTextContent('- -');
+});
+
+it('should render with non-empty initial value', () => {
+  render(
+    <SQFormDropdown size="auto" SQFormProps={{initialValues: {state: 'MO'}}} />
```
codyarose, line 31: nitpick: it _may_ be useful to export the `MOCK_STATE_OPTIONS` from the dropdown story so it can be used in the test, that might give a bit more context than 'MO' and 'Missouri'. Not a super necessary change though

codyarose (approved):
Great job on these 😃 After a lot of trial and error I think we've figured out a pretty good system for testing stories

(also I'm going to try to figure out why the `size` prop needs to be specified every time, that's kind of annoying)

PR #181, opened 2021-04-02
fix: 🐛 Adds ClickAwayListener to SQFormDatePicker

hvilander:
Uses MUI's ClickAwayListener and some state to close the picker if the
user clicks away or clicks the icon to open the calendar.

✅ Closes: 170

**Note:** we may want to do this to SQFormDateTimePicker, but I doubt we will want to put it on SQFormDatePickerWithCalendarInputOnly. I wanted to post this PR first to make sure we want to add this to DatePicker. I believe that we would want this behavior for all datepicker instances. If not I will add an example to the docs or something. 

https://www.loom.com/share/85246215cddc4ca88e3a691effb33f11


hvilander:
If we do like this and want to apply it SQFormDateTimePicker, I can update it here or follow up with a chore.

SeanGroff:
@hvilander I encountered this bug yesterday. Legit issue. I'm confused why this doesn't work out of the box. I suspect we broke this within our implementation.

Look at the examples in the docs, the functionality is there without the need to add a click-away listener ourselves.
https://material-ui-pickers.dev/demo/datepicker

Would you mind doing some digging around in our code to maybe figure out why this isn't working out of the box. I'm not against this solution. I just have a hunch this should work out of the box.

codyarose:
> @hvilander I encountered this bug yesterday. Legit issue. I'm confused why this doesn't work out of the box. I suspect we broke this within our implementation.
> 
> Look at the examples in the docs, the functionality is there without the need to add a click-away listener ourselves.
> https://material-ui-pickers.dev/demo/datepicker
> 
> Would you mind doing some digging around in our code to maybe figure out why this isn't working out of the box. I'm not against this solution. I just have a hunch this should work out of the box.

I think this is an issue with Mui, and probably portals, that others have had as well 
https://github.com/mui-org/material-ui/issues/25135

Chris-Boyle:
@hvilander If you have the datepicker open and you click on the text field for that date component does the picker piece close and the input part has the focus?

codyarose:
I'm curious if this change has any effect on how the dialog backdrop click functions.


> @hvilander If you have the datepicker open and you click on the text field for that date component does the picker piece close and the input part has the focus?  That was one scenario I didn't see in the Loom.

If it doesn't you could probably just add the `handleClickAway` to the text field `onClick`.

hvilander:
> I'm curious if this change has any effect on how the dialog backdrop click functions.
> 
> > @hvilander If you have the datepicker open and you click on the text field for that date component does the picker piece close and the input part has the focus?  That was one scenario I didn't see in the Loom.
> 
> If it doesn't you could probably just add the `handleClickAway` to the text field `onClick`.

Yep, good call Chris. https://www.loom.com/share/bf02acd51c0e41319e368b05d5838f93

Chris-Boyle (approved):

codyarose (approved):

hvilander:
Loom for changes to SQFormDateTimePicker https://www.loom.com/share/c7e7dd9c3fda45b0a80518d269c481f6

SeanGroff (approved):
Thank you @hvilander for fixing this bug 🚀

Issue #186, opened 2021-04-05
Update DatePicker stories


codyarose:
Update DatePicker stories to use new Storybook format to auto-generate docs

PR #188, opened 2021-04-06
docs: ✏️ update DatePicker stories

codyarose:
Update DatePicker stories to use new Storybook format to auto-generate
docs

Also added `SQFormStoryWrapper` component to be used in place of `SQForm` in stories. It takes the basic props a component story would need and handles the onSubmit so that doesn't have to be set up with storybook actions for each story

https://www.loom.com/share/b7ea539287e5431aa7e10a315b1eafcd

Update: Refactored `SQFormStoryWrapper` to include a checkmark icon button for submitting because some components don't have a way to submit via keyboard. Also updated `SQFormDropdown/MultiSelect` to use wrapper

✅ Closes: #186

### DIFF
--- a/stories/SQFormDatePicker.stories.js
+++ b/stories/SQFormDatePicker.stories.js
@@ -1,71 +1,68 @@
 import React from 'react';
-import {withKnobs, boolean} from '@storybook/addon-knobs';
 import * as Yup from 'yup';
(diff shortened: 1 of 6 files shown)


SeanGroff (approved):
Clean! 🧼

Issue #189, opened 2021-04-06
Upgrade Storybook to 6.2


codyarose:
https://storybook.js.org/blog/storybook-6-2/

PR #190, opened 2021-04-06
docs: ✏️ upgrade Storybook to 6.2

codyarose:
### [Storybook 6.2](https://storybook.js.org/blog/storybook-6-2/)

Relevant benefits:

- [Single story hoisting](https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#single-story-hoisting)
- Moving to ESM
- Early stages of [ESBuild/Vite support](https://storybook.js.org/blog/storybook-for-webpack-5/)
- JSON editor in Controls
- Controls sync to story URL, so you can share links to stories with dynamic settings

Note: I had to add `react` and `react-dom` to devDependencies to quell [this error](https://github.com/storybookjs/storybook/issues/13209) when running `npm run storybook`: `Error: Cannot find module 'react/package.json'` (the same error showed about `react-dom` if only `react` was added to devDependencies)

✅ Closes: #189


package.json
```diff
@@ -126,14 +126,16 @@
     "husky": "^3.1.0",
     "lint-staged": "^9.4.2",
     "prettier": "^1.18.2",
+    "react": "^17.0.2",
```
SeanGroff, line 129: @codyarose `react` and `react-dom` are listed above as peer dependencies.
It should be one or the other to my knowledge.

Also, most of our projects are on React version 16.x. They will also receive warnings that SQForm requires v17 or higher. Will these storybook updates work with React 16.x?

SeanGroff (changes requested):

package.json
codyarose: Yeah that's what I thought, but for some reason it won't run unless they're both manually installed.

Just tried reinstalling those using the versions in peerDeps and it does work with 16.x as well

codyarose: If this would be an issue I can make this a draft then try to poke around the storybook repo for a better solution/explanation

codyarose: @SeanGroff I found this from one of the maintainers https://github.com/storybookjs/storybook/issues/13269#issuecomment-736980525

If I understand correctly, react and react-dom are peerDeps in Storybook starting with 6.1, so the project needs to have those packages installed for it to work.

I think this problem isn't more widespread because in npm 7 peerDeps are automatically installed. We're on npm 6, so this requires manual installation of those dependencies

SeanGroff: > @SeanGroff I found this from one of the maintainers [storybookjs/storybook#13269 (comment)](https://github.com/storybookjs/storybook/issues/13269#issuecomment-736980525)
> 
> If I understand correctly, react and react-dom are peerDeps in Storybook starting with 6.1, so the project needs to have those packages installed for it to work.
> 
> I think this problem isn't more widespread because in npm 7 peerDeps are automatically installed. We're on npm 6, so this requires manual installation of those dependencies

Sorry for the late response. I believe all you need to do is to remove react and react-dom from peer deps. The change the version of react and react-dom in the dev dependencies to ^16.12.0 or whatever version they were previously in the peer deps

codyarose: Shouldn't we keep the peerDeps to make sure the consumers of the library have the right react version?

SeanGroff: > Shouldn't we keep the peerDeps to make sure the consumers of the library have the right react version?

Right, you added React as a dev dependency and not a dependency on this project. Now that my heads on straight, this implementation is fine. The peer dep for React is v16.x, and the dev dep that doesn't get shipped to the consumer can be whatever we need, v17.x in this case.

SeanGroff (approved):
@codyarose  It only took us a week to update to the latest Storybook. This is a new record lol.

PR #191, opened 2021-04-06
Issue 185 close datepicker with calendar input only modal

saadurrehman-selectquote:
fix: 🐛 close the datepicker modal  when we click outside

close the datepicker modal when we click outside the datepicker modal

✅ Closes: 185

https://www.loom.com/share/073f042148ed43f3a3fc6dc27b7b3848

screenshot of story:
image

### DIFF
--- a/src/components/SQForm/SQFormDatePicker.js
+++ b/src/components/SQForm/SQFormDatePicker.js
@@ -27,7 +27,9 @@ function SQFormDatePicker({
   onChange,
   setDisabledDate,
   muiFieldProps = {},
-  muiTextInputProps = {}
+  muiTextInputProps = {},
+  isOpen,
+  closeCalendar
 }) {
   const {
     formikField: {field, helpers},
@@ -51,6 +53,9 @@ function SQFormDatePicker({
   const handleClickAway = () => {
     if (isCalendarOpen) {
       setIsCalendarOpen(false);
+    }
+    if (isOpen) {
+      closeCalendar();
     }
   };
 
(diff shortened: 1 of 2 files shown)


abbas-ali123 (approved):

codyarose (commented):
With the recent changes to `SQFormDatePicker`, continuing to handle the open state of the calendar in `SQFormDatePickerWIthCalendarInputOnly` as well seems redundant. 

An alternative to this would be to expose the `toggleCalendar` function from `SQFormDatePicker` using `useImperativeHandle`

hvilander:
Managing the state in both components seems excessive, I have an alternate solution. Posting it shortly.

hvilander:
I figured it was easiest to show in a PR https://github.com/SelectQuoteLabs/SQForm/pull/192

hvilander:
My comments now seem a bit stale, I missed Cody's comment above. That sound maybe even better than my proposal.

codyarose:
This is overkill in comparison to #192, but figured I'd demo this type of potential solution anyway: https://www.loom.com/share/646b1af1de69423a8337a078316942a0

hvilander:
Thanks for the demo, I have not used an imperative handler before so that is quite informative. I do agree with it being overkill in this case.

SeanGroff:
Should this PR be cancelled? I believe we fixed this 🐛 earlier today?

codyarose:
@SeanGroff Yes, this was fixed by #192

PR #192, opened 2021-04-06
fix: 🐛 Handles dismissal better on Calendar Only Date picker

hvilander:
Now that SQFormDatePicker has state for open and closing the calendar;
this resolves an issue with the calendar only variant not working well
with the new changes.

✅ Closes: 185

Loom: https://www.loom.com/share/aa1e2daebc664e7c90b30115a50da516


src/components/SQForm/SQFormDatePicker.js
```diff
@@ -59,6 +60,14 @@ function SQFormDatePicker({
   // An empty string will not reset the DatePicker so we have to pass null
   const value = field.value || null;
 
+  const handleClickTextField = () => {
+    if (isCalendarOnly) {
+      setIsCalendarOpen(!isCalendarOpen);
```
saadurrehman-selectquote, line 65: we can call toggleCalendar function here,

codyarose:
I think this is a good solution as well, much more simple than using `useImperativeHandle`.

saadurrehman-selectquote (approved):

SeanGroff (approved):
Great solution @hvilander 🚀

Issue #194, opened 2021-04-07
Stories for TextField, Textarea, and ReadOnlyField


codyarose:
Ref #161

PR #195, opened 2021-04-07
docs: ✏️ stories for TextField, Textarea, and ReadOnlyField

codyarose:
Stories with auto-generated controls and docs. 

You may have noticed that I've been disabling the `name` prop in the controls tab - this is to prevent errors that would be caused by deleting the content of that field (because it's a required prop and we don't have a fallback for when it's empty). Also, if you change the name then submit, the submitted value shows the initial name _and_ the new name, which doesn't represent how the component actually works.
This only affects the prop rendering in Controls, it still shows in the props table in the Docs tab.

https://www.loom.com/share/bc1b17ed9f3e4dfd8175623a96a3e101

✅ Closes: #194

### DIFF
--- a/stories/SQFormReadOnlyField.stories.js
+++ b/stories/SQFormReadOnlyField.stories.js
@@ -1,0 +1,47 @@
+import React from 'react';
+
+import {SQFormReadOnlyField as SQFormReadOnlyFieldComponent} from '../src';
+import {SQFormStoryWrapper} from './components/SQFormStoryWrapper';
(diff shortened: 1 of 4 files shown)


SeanGroff (approved):

SeanGroff:
Approved pending the new merge conflicts @codyarose

codyarose:
> Approved pending the new merge conflicts @codyarose

Resolved 👍

Issue #204, opened 2021-04-08
Stories tests are failing because of `addon-docs/blocks`


codyarose:
For some reason the tests that use stories are failing. Based on the error message, we probably just need to ignore the offending package with some kind of transform

```zsh
● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/en/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /Users/cody.rose/Dev/github/SQForm/node_modules/@storybook/addon-docs/blocks.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){export * from './dist/esm/blocks';
                                                                                             ^^^^^^

    SyntaxError: Unexpected token 'export'

      1 | import React from 'react';
    > 2 | import {
        | ^
      3 |   Title,
      4 |   Description,
      5 |   ArgsTable,

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1350:14)
      at Object.<anonymous> (stories/utils/createDocsPage.js:2:1)
```

PR #205, opened 2021-04-09
chore: 🤖 ignore `@storybook/addon-docs` in jest config

codyarose:
The tests using stories were failing because
`@storybook/addon-docs/blocks` is using esm, which jest currently does
not understand. I've added that package to the `transformIgnorePatterns`
list in the jest config to solve that, but also had to change the
`.babelrc` to `babel.config.json` due to updates in Babel v7

Apparently, the deal with Babel 7 is that it no longer automatically loads the `.babelrc`. It now has the concept of a `root` directory, which is the current working directory by default, where it looks for a `babel.config.json` file. With this change Jest will be able to properly load the babel config.

References: 
- [similar situation with a different storybook package using esm](https://github.com/storybookjs/storybook/issues/9470#issuecomment-574790318)
- [Babel 7 config files docs](https://babeljs.io/docs/en/config-files#project-wide-configuration)
- [Jest issue demonstrating `.babelrc` vs `babel.config.js` when using `transformIgnorePatterns`](https://github.com/facebook/jest/issues/10256)

✅ Closes: #204

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,5 +1,5 @@
 {
-  "presets": [["@babel/preset-env"],["@babel/preset-react"]],
+  "presets": [["@babel/preset-env"], ["@babel/preset-react"]],
   "plugins": [
     "react-docgen",
     "@babel/plugin-proposal-nullish-coalescing-operator",
--- a/jest.config.js
+++ b/jest.config.js
@@ -7,5 +7,6 @@ module.exports = {
   },
   moduleNameMapper: {
     '\\.(css|less)$': '<rootDir>/__mocks__/styleMock.js'
-  }
+  },
+  transformIgnorePatterns: ['/node_modules/(?!@storybook/addon-docs).+\\.js$']
 };


SeanGroff:
@codyarose Do we want to bump our package version for this fix? Jest shouldn't affect our consumers.
Should this be a `chore` commit?

codyarose:
> @codyarose Do we want to bump our package version for this fix? Jest shouldn't affect our consumers.
> Should this be a `chore` commit?

Ah, right.. I suppose that's true - as long as the change from `.babelrc` &#8594; `babel.config.json` doesn't affect anything?

SeanGroff:
> > @codyarose Do we want to bump our package version for this fix? Jest shouldn't affect our consumers.
> > Should this be a `chore` commit?
> 
> Ah, right.. I suppose that's true - as long as the change from `.babelrc` → `babel.config.json` doesn't affect anything?

Since they are both JSON syntax nothing should be affected. The consumer transpiles their own code.

codyarose:
@SeanGroff Okay I amended the commit locally and pushed it up. That should do it, right?

SeanGroff (approved):

PR #211, opened 2021-04-09
chore: 🤖 fix `SnackBar` storybook deploy issue

codyarose:
The storybook deploy action has been failing because of the use of
`@material-ui/core/SnackBar`. Hoping that switching to the `Snackbar`
from `scplus-shared-components` will fix this.

Functionality of the snackbar from the `SQFormStoryWrapper` is the same, but it now appears in the top right corner by default


SeanGroff (approved):
Great!

Issue #217, opened 2021-04-20
Add more tests - SQFormButton


laurelbean:
Now that we have a better idea of how the tests should look and run, let's start working on adding tests for all the components.

This issue will focus on SQFormButton

PR #218, opened 2021-04-20
test: 💍 Add SQFormButton tests

laurelbean:
Adding tests for the SQFormButton stories.

I moved the order of the args in the stories so that I could override the onClick with a mock/spy function.

✅ Closes: #217

### DIFF
--- a/stories/__tests__/SQFormButton.stories.test.js
+++ b/stories/__tests__/SQFormButton.stories.test.js
@@ -1,0 +1,161 @@
+import React from 'react';
+import {render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import {composeStories} from '@storybook/testing-react';
+import * as stories from '../SQFormButton.stories';
+
+const {
+  Default: SQFormButton,
+  WithTestField: SQFormButtonWithField
+} = composeStories(stories);
+
+describe('Default Tests', () => {
+  it('should render a button with text', () => {
+    render(<SQFormButton />);
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    expect(submitButton).toBeInTheDocument();
+    expect(submitButton).toHaveTextContent('Submit');
+  });
+
+  it('should render a reset button given the type', () => {
+    render(<SQFormButton type="reset" />);
+
+    const resetButton = screen.getByRole('button', {name: /form submission/i});
+
+    expect(resetButton).toBeInTheDocument();
+    expect(resetButton).toHaveAttribute('type', 'reset');
+  });
+
+  it('should call a function when clicked', () => {
+    const onClickSpy = jest.fn();
+    render(<SQFormButton onClick={onClickSpy} />);
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    userEvent.click(submitButton);
+
+    expect(onClickSpy).toHaveBeenCalled();
+  });
+
+  it('should show as disabled', () => {
+    render(<SQFormButton isDisabled={true} />);
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    expect(submitButton).toBeDisabled();
+  });
+
+  it('should not fire onClick when disabled', () => {
+    const onClickSpy = jest.fn();
+    render(<SQFormButton isDisabled={true} onClick={onClickSpy} />);
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    userEvent.click(submitButton);
+
+    expect(onClickSpy).not.toHaveBeenCalled();
+  });
+});
+
+describe('Button with Test Field Tests', () => {
+  it('should render a button and a text field', () => {
+    render(<SQFormButtonWithField />);
+
+    const testField = screen.getByLabelText(/test field/i);
+
+    expect(testField).toBeInTheDocument();
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    expect(submitButton).toBeInTheDocument();
+    expect(submitButton).toHaveTextContent('Submit');
+  });
+
+  it('should disable when shouldRequireFieldUpdates is true', () => {
+    render(<SQFormButtonWithField shouldRequireFieldUpdates={true} />);
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    expect(submitButton).toBeDisabled();
+  });
+
+  it('should enable when text field is filled out', () => {
+    render(<SQFormButtonWithField shouldRequireFieldUpdates={true} />);
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+    expect(submitButton).toBeDisabled();
+
+    const testField = screen.getByLabelText(/test field/i);
+
+    userEvent.type(testField, 'Hello');
+
+    expect(submitButton).toBeEnabled();
+  });
+
+  it('should submit form when submit button clicked', () => {
+    const onClickSpy = jest.fn();
+    //const onClickOverride = jest.
+    render(
+      <SQFormButtonWithField
+        shouldRequireFieldUpdates={true}
+        onClick={onClickSpy}
+      />
+    );
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+    expect(submitButton).toBeDisabled();
+
+    userEvent.click(submitButton);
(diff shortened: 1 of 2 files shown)


stories/__tests__/SQFormButton.stories.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+  it('should call a function when clicked', () => {
+    const onClickSpy = jest.fn();
+    render(<SQFormButton onClick={onClickSpy} />);
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    userEvent.click(submitButton);
+
+    expect(onClickSpy).toHaveBeenCalled();
```
hvilander, line 39: Should we specify, one click should be called once?

stories/__tests__/SQFormButton.stories.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    userEvent.click(submitButton);
+
+    expect(onClickSpy).not.toHaveBeenCalled();
+  });
+});
+
+describe('Button with Test Field Tests', () => {
```
hvilander, line 62: [question] maybe this is just my past, is it convention to have one test file per component? Should these be split out into their own file?

stories/__tests__/SQFormButton.stories.test.js
laurelbean: Yeah, ideally I could test that data that was submitted, but I'd have to make some changes to how the current stories are implemented to be able to override onSubmit or to use console.warn or something like that I can mock.
I will update it to be specific to 1 call.

stories/__tests__/SQFormButton.stories.test.js
laurelbean: So we changed to testing against the stories rather than the component itself. So the SQFormButton has 2 stories associated with it. So I thought it would be nice to group the tests against each story together. I thought, they're both testing buttons, but just using different base stories. I'm open to other ideas for organizing tests. We didn't go through it with SQFormDropdown since there was only 1 story for it.

stories/__tests__/SQFormButton.stories.test.js
SeanGroff: I think we can specify what the `spy` is called with right? (The args it should be invoked with)

stories/__tests__/SQFormButton.stories.test.js
```diff
@@ -0,0 +1,161 @@
...
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    expect(submitButton).toBeInTheDocument();
+    expect(submitButton).toHaveTextContent('Submit');
+  });
+
+  it('should render a reset button given the type', () => {
+    render(<SQFormButton type="reset" />);
+
+    const resetButton = screen.getByRole('button', {name: /form submission/i});
```
SeanGroff, line 25: If the button is a reset button, I wouldn't expect the name to be `form submission` 🤔

stories/__tests__/SQFormButton.stories.test.js
SeanGroff: > I think we can specify what the `spy` is called with right? (The args it should be invoked with)

Ah nm, I don't think we can do this because we are in the component story

laurelbean: I did attempt it, but the actual args that go to onClick was a lot to try and match. The SQForm.stories ones that we check the args for are with console.warn

codyarose: I think that would be implementation details territory. Here we're testing that the onClick is set up properly - what it's called with is dependent on how/where the button is implemented, right? If the button is always used to submit a form, that would be a test better suited for the form component

stories/__tests__/SQFormButton.stories.test.js
codyarose: I think that's something we'll figure out as more tests are written. I prefer single files, but if the component has many stories and the test suites for each are verbose then breaking them out into a folder of test files would probably be easier to manage

stories/__tests__/SQFormButton.stories.test.js
laurelbean: That's the default title in SQFormButton. I can pass in an override to make it say "Reset Form" or something similar.

codyarose: Sounds like an improvement that could be made to the component itself

SeanGroff: > That's the default title in SQFormButton. I can pass in an override to make it say "Reset Form" or something similar.

Is this a bug though? (I did this LOL)

If the `type` prop is set to `RESET`, the title should change to Reset Form

SeanGroff (approved):
Amazing work @laurelbean 💯

stories/__tests__/SQFormButton.stories.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+  it('should disable when shouldRequireFieldUpdates is true', () => {
+    render(<SQFormButtonWithField shouldRequireFieldUpdates={true} />);
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    expect(submitButton).toBeDisabled();
+  });
+
+  it('should enable when text field is filled out', () => {
```
codyarose, line 84: nitpick: I think the test name could be a bit more explicit in describing what it's testing. Something like `should enable after field value is changed when shouldRequireFieldUpdates is true`

stories/__tests__/SQFormButton.stories.test.js
```diff
@@ -0,0 +1,161 @@
...
+    expect(submitButton).toBeDisabled();
+
+    const testField = screen.getByLabelText(/test field/i);
+
+    userEvent.type(testField, 'Hello');
+
+    expect(submitButton).toBeEnabled();
+  });
+
+  it('should submit form when submit button clicked', () => {
```
codyarose, line 97: I'm wondering if this test is really necessary here since the onClick has already been tested in the previous suite and testing the form submission has more to do with testing SQForm rather than the button

stories/__tests__/SQFormButton.stories.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+    expect(testField).toBeInTheDocument();
+
+    const submitButton = screen.getByRole('button', {name: /form submission/i});
+
+    expect(submitButton).toBeInTheDocument();
+    expect(submitButton).toHaveTextContent('Submit');
+  });
+
+  it('should disable when shouldRequireFieldUpdates is true', () => {
```
codyarose, line 76: nitpick: I think this test would fit better categorically in the default tests suite because it's testing a behavior that would happen regardless of if it was associated with an input field

stories/__tests__/SQFormButton.stories.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+    userEvent.type(testField, 'Hello');
+    expect(submitButton).toBeEnabled();
+
+    userEvent.click(submitButton);
+
+    expect(onClickSpy).toHaveBeenCalled();
+  });
+
+  it('should be disabled when button is type reset', () => {
```
codyarose, line 123: nitpick: same idea as the comment I left on line 76.
If this test is moved into the default suite, the title of the following test would need to be updated to be more explicit about what it's testing

codyarose (changes requested):

stories/__tests__/SQFormButton.stories.test.js
laurelbean: I think you're right. This test isn't even really testing the form submit like the name suggests. I'll remove it.

laurelbean:
Updated based on current feedback! The diffs on this last one look pretty hard to read so I'd suggest just looking at the file directly

20BBrown14 (approved):

SeanGroff:
> Updated based on current feedback! The diffs on this last one look pretty hard to read so I'd suggest just looking at the file directly

@codyarose Can you give this PR a gander when you have a moment? I see some outstanding conversations I want to ensure are resolved. Or maybe still need attention.

stories/__tests__/SQFormButton.stories.test.js
codyarose: created issue #220 for this

PR #219, opened 2021-04-27
issue-212: Fix misaligned select field

20BBrown14:
Fixed the components that misaligned bottom borders due to the MUI Select. SQFormDropdown and SQFormMultiselect.

Before: 
image

After:
Screen Shot 2021-04-27 at 10 17 23 AM

### DIFF
--- a/src/components/SQForm/SQFormDropdown.js
+++ b/src/components/SQForm/SQFormDropdown.js
@@ -7,6 +7,7 @@ import MenuItem from '@material-ui/core/MenuItem';
 import InputLabel from '@material-ui/core/InputLabel';
 import FormHelperText from '@material-ui/core/FormHelperText';
 import Grid from '@material-ui/core/Grid';
+import {makeStyles} from '@material-ui/core/styles';
 
 import {useForm} from './useForm';
 import {
@@ -18,6 +19,14 @@ import {EMPTY_LABEL} from '../../utils/constants';
 
 const EMPTY_VALUE = '';
 const EMPTY_OPTION = {label: EMPTY_LABEL, value: EMPTY_VALUE};
+
+const useStyles = makeStyles({
+  selectHeight: {
+    '& .MuiSelect-selectMenu': {
+      height: '1.1876em'
+    }
+  }
+});
 
 function SQFormDropdown({
   children,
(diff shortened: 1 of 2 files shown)


codyarose (approved):

src/components/SQForm/SQFormDropdown.js
```diff
@@ -19,6 +20,14 @@ import {EMPTY_LABEL} from '../../utils/constants';
 const EMPTY_VALUE = '';
 const EMPTY_OPTION = {label: EMPTY_LABEL, value: EMPTY_VALUE};
 
+const useStyles = makeStyles({
+  selectHeight: {
+    '& .MuiSelect-selectMenu': {
+      height: '1.1876em'
```
SeanGroff, line 26: Can you convert this to a `px` value, please?

src/components/SQForm/SQFormMultiSelect.js
```diff
@@ -60,6 +61,14 @@ const getToolTipTitle = (formikFieldValue, options) => {
   return selectedDisplayValue(formikFieldValue, options);
 };
 
+const useStyles = makeStyles({
+  selectHeight: {
+    '& .MuiSelect-selectMenu': {
+      height: '1.1876em'
```
SeanGroff, line 67: px 🙏

src/components/SQForm/SQFormDropdown.js
20BBrown14: I used the exact value that other MUI fields are using (from MUI directly). I figure it'd be best to match it exactly to ensure there not be any issues in the future.

SeanGroff: > I used the exact value that other MUI fields are using (from MUI directly). I figure it'd be best to match it exactly to ensure there not be any issues in the future.

Everything from MUI results in `px`. If we leave the unit off a value, MUI will assume `px`.

I don't know `em` as well as `rem` but that should equate to `14px`. 
Can we test if using `px` will result in a pixel perfect alignment?

codyarose: The internal style Mui uses is in `em` to make the height of the element scale with the font-size. I believe `1.1876em` would equate to `14px` with the current font-size, but if the font-size on the component is changed from the default value somewhere the height won't scale properly if it's set to `14px` explicitly.

SeanGroff (approved):
Nice work, thanks for fixing this @20BBrown14!

Issue #221, opened 2021-04-28
Add more tests - SQFormIconButton


laurelbean:
Continuing to add more tests to components.

This issue will focus on SQFormIconButton.

PR #222, opened 2021-04-28
💍 Add tests for SQFormIconButton

laurelbean:
Added some tests for SQFormIconButton. These are generally the same as the tests for SQFormButton since they have similar functions.

✅ Closes: #221

### DIFF
--- a/stories/__tests__/SQFormIconButton.stories.test.js
+++ b/stories/__tests__/SQFormIconButton.stories.test.js
@@ -1,0 +1,153 @@
+import React from 'react';
+import CheckCircle from '@material-ui/icons/CheckCircle';
+import {composeStories} from '@storybook/testing-react';
+import {render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import * as stories from '../SQFormIconButton.stories';
+
+const {
+  Default: SQFormIconButton,
+  WithTestField: SQFormIconButtonWithField
+} = composeStories(stories);
+
+describe('SQFormIconButton Tests', () => {
+  describe('Button Only', () => {
+    it('should render a submit button', () => {
+      render(<SQFormIconButton exampleIcons={CheckCircle} />);
+
+      const iconButton = screen.getByRole('button', {name: /form submission/i});
+
+      expect(iconButton).toBeInTheDocument();
+    });
+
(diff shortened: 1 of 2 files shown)


codyarose (changes requested):
Looks good - I think the only thing that's missing is a test to ensure that an icon is rendered within the button. Probably just need to check that the SVG element is there - and maybe that the SVG's `title` attribute matches the name of whichever Mui icon is specified, that may not be possible or necessary though

stories/__tests__/SQFormIconButton.stories.test.js
```diff
@@ -0,0 +1,153 @@
...
+
+      const iconButton = screen.getByRole('button', {
+        name: /form submission/i
+      });
+
+      expect(iconButton).toBeDisabled();
+    });
+  });
+
+  describe('Button with Test Field', () => {
```
20BBrown14, line 105: typo: `text`

20BBrown14 (approved):

stories/__tests__/SQFormIconButton.stories.test.js
laurelbean: I meant it to be Test. The name of the story is `WithTestField`

stories/__tests__/SQFormIconButton.stories.test.js
```diff
@@ -12,12 +12,23 @@ const {
...
 describe('SQFormIconButton Tests', () => {
   describe('Button Only', () => {
+    it('should render a button with an icon', () => {
+      render(<SQFormIconButton exampleIcons={CheckCircle} />);
+
+      const iconButton = screen.getByRole('button', /form submission/i);
+      const svg = document.getElementsByTagName('svg');
+
+      expect(iconButton).toBeInTheDocument();
+      expect(svg.length).toEqual(1);
```
laurelbean, line 22: @codyarose This is the closest I could get it to test for the icon. The <svg> is `area-hidden`, doesn't have a role, and doesn't have any text.

SeanGroff: @laurelbean @codyarose I'm going to merge this PR, then i'll add a comment to #228 that we should fix this test as part of the PR that resolves issue #228

SeanGroff (approved):

Issue #236, opened 2021-05-10
SQFormAsyncAutocomplete does not show given initialValues


laurelbean:
Loom video: https://www.loom.com/share/e75a9cb653344748979eb119b444130b

Looks like this component is missing the code to display the initial value.

Found while working #229

PR #237, opened 2021-05-10
fix: 🐛 Add initialValues display to SQFormAsyncAutocomplete

laurelbean:
Even when you passed in a value to the initialValues object, an
SQFormAsyncAutocomplete field would not displaythe label that
coresponded with the given value.

I based this on how SQFormAutocomplete handles the initialValues.

✅ Closes: #236


SeanGroff (approved):
LGTM, nice fix @laurelbean 🚀 
Hooray for testing exposing issues!

Issue #244, opened 2021-05-11
Migrate SQForm to TypeScript


SeanGroff:
Perform initial TypeScript setup on this library so devs can begin to convert form components to TS over time. 

---

@codyarose edit: 
## Typescript Migration

We can start the migration process! We'll be working against this branch: `issue/244-EnableTypeScript`, and this issue will be used to track the files that need to be converted from JS to TS in a task list.

### The process will be :
- pick a component/file/folder to migrate
- create an issue to claim it
- create a branch off the remote branch `issue/244-EnableTypeScript`
- convert the file to TS
- create a PR that will merge into `issue/244-EnableTypeScript`
    - add a reference to this issue in the PR description; `Ref #244`

When everything is converted to TS, #248 will be merged to `master`

## Types can be tricky -- take your best stab at it. 
If you get stuck on something there's a good chance someone else has, or would, too, so ask for help in Slack or leave some comments on your PR about where you got stuck or what you're not sure about.

---

'folderName/`*`' means all files in that folder must be converted to Typescript as one task

- [x] src/
	- [x] `index.js`
	- [x] hooks/`*`
	- [x] #402
	- [x] components/
		- [x] #301 
		- [x] #359
		- [x] SQForm/
			- [x] #524
			- [x] #555
			- [x] #556
			- [x] #346
			- [x] #406
			- [x] #384
			- [x] `SQFormButton.js`
			- [x] #336 
			- [x] #314 
			- [x] #315 
			- [x] #335 
			- [x] #369
			- [x] #432
			- [x] #428
			- [x] #378
			- [x] #332
			- [x] #351
			- [x] #344
			- [x] #379
			- [x] #394
			- [x] #409
			- [x] #400
			- [x] #342 
			- [x] #311 
			- [x] #364
			- [x] #368 
			- [x] #366
			- [x]  #341 
			- [x] `useForm.js`
			- [x] #349
			- [x] #474
		- [x] #395
		- [x] #506
		- [x] #478
		- [x] #526
		- [x] #566
		- [x] #567
- [x] stories/
	- [x] #415
	- [x] #417
	- [x] #348
		- [x] #591
		- [x] #588
		- [x] #606
		- [x] #593
		- [x] #549
		- [x] #552
		- [x] #545
		- [x] `SQFormButton.stories.test.js`
		- [x] #408
		- [x] #539
		- [x] #427
		- [x] #422 
		- [x] #534
		- [x] #424 
		- [x] #542
[...]

PR #248, opened 2021-05-11
feat: 🎸 Support for TypeScript

SeanGroff:
✅ Closes: #244

Going to mark this PR as a Draft since I ran out of time to do more thorough testing. This seems to work as expected. I might be missing some library types packages.

Would appreciate it if anyone else would be so kind to pull these changes down and test them out.


caleboleary:
@SeanGroff I pulled this & converted SQFormDialog (quick & messy) to TS - storybook was broken, but changing the webPackFinal in it's main.js to the following seems to fix it:

`webpackFinal: config => {
    config.module.rules = [
      ...config.module.rules,
      {
        test: /.jsx?$|.tsx?$/,
        exclude: /node_modules/,
        use: {
          loader: 'babel-loader',
          options: {
            presets: [
              '@babel/preset-env',
              '@babel/preset-typescript',
              '@babel/preset-react'
            ]
          }
        }
      }
    ];

    return config;
  }`

caleboleary:
Also, FYI, I had to force TS to ignore some of the JS imports that it can't get typings from like this: 
```
// @ts-ignore
import SQFormDialogInner from './SQFormDialogInner';
// @ts-ignore
import {useInitialRequiredErrors} from '../../hooks/useInitialRequiredErrors';
```

codyarose:
Found a few things:

- This could be added to `.eslintrc.json`
```js
"extends": [
    ...
    "plugin:@typescript-eslint/recommended"
  ],
```

- For tests to work, 
    - `ts-jest` and some jest config changes need to be added, https://basarat.gitbook.io/typescript/intro-1/jest
    - `@types/testing-library__jest-dom` for the jest matcher types

caleboleary:
Screen Shot 2021-05-25 at 11 02 22 AM

npm packed it into SRTS2 and I'm getting type warnings! 

Also, storybook looks to be working well and grabbing types for the controls panel. Nice!

codyarose:
@SeanGroff Would this be considered a breaking change for any projects that use TS because of the added types? If they aren't using the proper types it could cause errors that would prevent a build

caleboleary:
In our project, I had a file at root `declare module '@selectquotelabs/sqform'; ` that prevented any type warnings at all for the lib, had to remove it to receive any. I *think* I could theoretically upgrade to a TS based sqform, leave that file there, and still use the wrong types somewhere if I was, but not positive.

codyarose:
> In our project, I had a file at root `declare module '@selectquotelabs/sqform'; ` that prevented any type warnings at all for the lib, had to remove it to receive any. I _think_ I could theoretically upgrade to a TS based sqform, leave that file there, and still use the wrong types somewhere if I was, but not positive.

Ah, yeah the same was true when I tried it in my project. So in that case we it wouldn't be a breaking change, but they'd need to know about remove that `d.ts` file in order to unblock the types

SeanGroff:
> > In our project, I had a file at root `declare module '@selectquotelabs/sqform'; ` that prevented any type warnings at all for the lib, had to remove it to receive any. I _think_ I could theoretically upgrade to a TS based sqform, leave that file there, and still use the wrong types somewhere if I was, but not positive.
> 
> Ah, yeah the same was true when I tried it in my project. So in that case it wouldn't be a breaking change, but they'd need to know about removing that `d.ts` file (if it exists) in order to unblock the types

Ah yes....I think the right thing to do is NOT merge this to `master` until we've fully converted over to TypeScript.
So when we create an issue with a task list we need to make sure to specify the instructions of working against this branch, and PRing to this branch.

Then once we're fully migrated to TypeScript, merge this PR

codyarose:
> I don't think we want to tell consumers to manually add a .d.ts file. Then remove it at a later point.

True. I assumed that if anyone is using this library with TS they already have a `d.ts` for it, and if they're using it with JS they won't have typechecking so it wouldn't matter.

codyarose:
Updated from master to fix the merge conflicts

hvilander:
Resolved some merge conflicts with that merge, it was only consuming the onTertiaryClick in our dialog so I figured it was safe to change the relevant propTypes into types. Also I managed to commit with formatting issues, pushed that change as well.

20BBrown14 (approved):

hvilander (approved):

caleboleary (approved):

Issue #253, opened 2021-05-17
Add Tests for SQFormTextField and SQFormTextarea


laurelbean:
Convert existing SQFormTextField tests to use the story instead and add any other test cases needed. SQFormTextarea should have a similar set of tests.

PR #255, opened 2021-05-17
test: 💍 Tests for SQFormTextField and SQFormTextarea

laurelbean:
Converted the current tests for SQFormTextField to use the story instead
of the component. Also added more test cases for SQFormTextField. Since
the test cases needed for SQFormTextarea are very similar, I also added
those in.

As always, please let me know if I missed any test cases.

✅ Closes: #253


stories/__tests__/SQFormTextarea.stories.test.js
```diff
@@ -0,0 +1,130 @@
...
+import {render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import * as stories from '../SQFormTextarea.stories';
+
+const {
+  Default: SQFormTextarea,
+  WithValidation: SQFormTextareaWithValidation
+} = composeStories(stories);
+
+describe('SQFormTextField Tests', () => {
```
codyarose, line 12: The `describe` block titles need to be updated to SQFormTextarea

stories/__tests__/SQFormTextarea.stories.test.js
```diff
@@ -0,0 +1,130 @@
...
+  describe('Text Field Only', () => {
+    it('should render a label and text box', () => {
+      render(<SQFormTextarea size="auto" />);
+
+      //There's 2 labels but one has a parent with aria-hidden
+      const label = screen.getAllByText(/textarea/i)[0];
```
codyarose, line 18: It's annoying that Mui does this and doesn't give them proper roles.. 🙄

What we could do, instead of relying on the order of the returned array of elements, is find the element in the array that has the `<label>` parent element.
```js
const label = screen
        .getAllByText(/textarea/i)
        .find(item => item.parentElement.tagName === 'LABEL');
```
Kind of hacky and tests implementation details, but it at least gives us the confidence that we're always targeting the right "label"

stories/__tests__/SQFormTextarea.stories.test.js
```diff
@@ -0,0 +1,130 @@
...
+      expect(charCount).toBeInTheDocument();
+
+      const textbox = screen.getByRole('textbox', {name: /textarea/i});
+
+      userEvent.type(textbox, 'Hello World');
+
+      expect(textbox).toHaveValue('Hello');
+      const charCountAfter = screen.getAllByText(
+        `Textarea : ${maxCharacters}/${maxCharacters}`
+      )[0];
```
codyarose, line 99: The same thing could be used in this test
```js
const label = screen
          .getAllByText(`Textarea : 0/${maxCharacters}`)
          .find(item => item.parentElement.tagName === 'LABEL')
```

codyarose (changes requested):

codyarose:
Oh, and the old `SQFormTextField.test.js` file could be removed

laurelbean:
Made updates as suggested by @codyarose. Thanks Cody!

codyarose (approved):
👍

lucas-homer (approved):

PR #257, opened 2021-05-20
test: 💍 SQFormReadOnlyField

lucas-homer:
✅ Closes: 254

My first crack at shipping actual tests, so any feedback is appreciated!

### DIFF
--- a/stories/__tests__/SQFormReadOnlyField.stories.test.js
+++ b/stories/__tests__/SQFormReadOnlyField.stories.test.js
@@ -10,7 +10,7 @@ const {
 } = composeStories(stories);
 
 describe('SQFormReadOnlyField Tests', () => {
-  describe('Field Only', () => {
+  describe('Default ReadOnly Field', () => {
     it('should render a label and read-only text field', () => {
       render(<SQFormReadOnlyField size="auto" />);
 
@@ -22,7 +22,7 @@ describe('SQFormReadOnlyField Tests', () => {
       expect(textField).toBeInTheDocument();
     });
 
-    it('should render with "- -" initial value', () => {
+    it('should render with default "- -" initialValue', () => {
       render(<SQFormReadOnlyField size="auto" />);
 
       const textField = screen.getByRole('textbox', {name: /readonly field/i});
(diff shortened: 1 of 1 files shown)


stories/__tests__/SQFormReadOnlyField.stories.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+      const label = screen.getByText(/readonly field/i);
+      expect(label).toBeInTheDocument();
+      expect(label).toHaveTextContent('ReadOnly Field');
+
+      const textField = screen.getByRole('textbox', {name: /readonly field/i});
+      expect(textField).toBeInTheDocument();
+    });
+
+    it('should render with "- -" initial value', () => {
```
SeanGroff, line 25: `'should render "- -" when no initial value provided'`

Something to describe _why_ `- -` should render

laurelbean (approved):
Great job Lucas! Thanks for contributing tests

SeanGroff (approved):
LGTM!

stories/__tests__/SQFormReadOnlyField.stories.test.js
lucas-homer: isn't that what "when no initial value provided" answers?

SeanGroff: > isn't that what "when no initial value provided" answers?

The test currently reads `'should render with "- -" initial value'`

I'm just trying to provide a suggestion that adds immediate clarity to the test. 

Below, we have a test description of `'ReadOnly Field With Initial Value'`

For this test, we could move it so it's more grouped together with the related test.

Then our test could be `'ReadOnly Field Without Initial Value'`

Just trying to call out that users without tribal knowledge of our `- -` pattern may not get much context from the current test description.

lucas-homer: I see what you're saying. So arguably I could name that one "should render default "- -" when no initial value provided. I guess what drove me to what I initially did is to group them by the stories. We have one story called "Field Only" and one "ReadOnly Field With Initial Value" which might be better called "ReadOnly Field with Custom Initial Value." In any case, I can make some adjustments to clarify your point. Thanks!

Issue #247, opened 2021-05-11
Storybook Action Failing


SeanGroff:
The action started failing on this commit by yours truly 😅  
https://github.com/SelectQuoteLabs/SQForm/commit/ea2ee59307c7d116df6bb050a319db6dc0033fc5

lucas-homer:
PR #259 fixes this issue, closing...

PR #259, opened 2021-05-20
chore: 🤖 fix import error failing storybook deploy

lucas-homer:
import SQForm component from this project instead of
scplus-shared-components

(made a demo for this fix)[https://www.loom.com/share/8381f082f68542a0b3c8b3bfbaac0137]

### DIFF
--- a/stories/SQFormDisableableComponents.stories.js
+++ b/stories/SQFormDisableableComponents.stories.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import {SQForm} from 'scplus-shared-components';
 import {withKnobs, boolean} from '@storybook/addon-knobs';
 import * as Yup from 'yup';
 import Card from '@material-ui/core/Card';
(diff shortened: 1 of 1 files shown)


stories/SQFormDisableableComponents.stories.js
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
-import {SQForm} from 'scplus-shared-components';
 import {withKnobs, boolean} from '@storybook/addon-knobs';
```
SeanGroff, line 2: Dangit I missed this!

SeanGroff (approved):

stories/SQFormDisableableComponents.stories.js
lucas-homer: heh happens to us all! I guess we won't know until we try, but this MIGHT resolve issue #247

SeanGroff: I think so as well. The failure started happening after I did that huge breaking change refactor to the SSC library.

Issue #250, opened 2021-05-13
SQFormAutocomplete displayEmpty displays blank instead of default empty value


laurelbean:
Loom with issue: https://www.loom.com/share/19d2c1638bde427fa85dd0ed8d2bf383

When resetting an SQFormAutocomplete that has `displayEmpty` as true, resetting the field or using `setFieldValue(field, '')` does not show the empty value. Instead it shows blank.

PR #272, opened 2021-05-21
fix: 🐛 Correctly reset displayempty autocomplete to '- -'

caleboleary:
When resetting an entire form, all autocompletes would reset to empty
string, even if their default state should be '- -'. This fix handles
both cases correctly.

✅ Closes: #250

LoOoOoOOoOO0oom https://www.loom.com/share/6d4a93ccba264c59a52e4fe39bd93565


Chris-Boyle (approved):
Great Loom!  Very thorough!

lucas-homer (approved):

codyarose (approved):
✅

laurelbean (approved):
Nice work

PR #281, opened 2021-05-24
Issue 268  -- SQFormDialog.stories.test.js

lucas-homer:
I wrote some basic tests for SQFormDialog. I didn't want to get too complicated with these because already have tests for SQForm more generally, as well as a lot of our inputs, so I tried to focus on the "dialog" part of things, specifically the `title` prop, the `onSave` and `onClose` callbacks (including the save button enabled/disabled integration with validationSchema), and the auto-focus feature.

As always, any feedback is great, as I'm somewhat new to writing tests.

### DIFF
--- a/stories/__tests__/SQFormDialog.stories.test.js
+++ b/stories/__tests__/SQFormDialog.stories.test.js
@@ -1,25 +1,20 @@
 import React from 'react';
-import {
-  fireEvent,
-  render,
-  screen,
-  waitFor,
-  waitForElementToBeRemoved,
-  within
-} from '@testing-library/react';
+import {render, screen, waitFor} from '@testing-library/react';
 import {composeStories} from '@storybook/testing-react';
 
 import userEvent from '@testing-library/user-event';
 import * as stories from '../SQFormDialog.stories';
 
-const {Default} = composeStories(stories);
+const {Default, WithValidation, WithAutoFocus} = composeStories(stories);
 
 window.alert = jest.fn();
 const handleClose = jest.fn();
+const handleSave = jest.fn();
 
 afterEach(() => {
   window.alert.mockClear();
   handleClose.mockClear();
+  handleSave.mockClear();
 });
 
 const mockData = {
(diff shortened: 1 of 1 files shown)


stories/__tests__/SQFormDialog.stories.test.js
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import {render, screen, waitFor} from '@testing-library/react';
+import {composeStories} from '@storybook/testing-react';
+
+import userEvent from '@testing-library/user-event';
+import * as stories from '../SQFormDialog.stories';
+
+const {Default, WithValidation, WithAutoFocus} = composeStories(stories);
```
laurelbean, line 8: I like to alias the default to the component name, but that's a personal choice

stories/__tests__/SQFormDialog.stories.test.js
```diff
@@ -0,0 +1,94 @@
...
+  handleClose.mockClear();
+  handleSave.mockClear();
+});
+
+const mockData = {
+  hello: 'howdy'
+};
+
+describe('Tests for Default', () => {
+  it('renders Default with correct Title, form updates, and calls onSave on submit', async () => {
```
laurelbean, line 25: I like to have 1 main focus for each test, so I would consider moving the onSave/submit parts into their own test.
I think of it like 1 test for default rendering/no actions, 1 test for the submit action.

lucas-homer: this is great feedback. thanks!

stories/__tests__/SQFormDialog.stories.test.js
```diff
@@ -0,0 +1,126 @@
...
+    render(
+      <Default
+        isOpen={true}
+        onSave={handleSave}
+        onClose={handleClose}
+        title={dialogTitleValue}
+      />
+    );
+
+    const dialogTitle = screen.getByText(dialogTitleValue);
```
codyarose, line 37: You're getting those `wrapped in act(...) blah blah` errors because there's a state change happening inside of the Formik component. I'm _pretty_ sure that state change is the dialog opening because of the prop `isOpen={true}`. Obviously this state change didn't really affect the test because it still passes, but ideally you should ensure that any state changes have completed before making assertions.

Here the error is caused by using `getByText` before the dialog open state change is complete. So what you can do is "wait" for that text to appear, which should mean that the open state change has completed
```js
await waitFor(() => screen.getByText(dialogTitleValue));
const dialogTitle = screen.getByText(dialogTitleValue);

// or even better

const dialogTitle = await screen.findByText(dialogTitleValue);
```
Then the `expect( ... )` assertion that follows won't run until that `findByText` has resolved.

https://testing-library.com/docs/dom-testing-library/api-async#findby-queries

stories/__tests__/SQFormDialog.stories.test.js
```diff
@@ -0,0 +1,126 @@
...
+    expect(dialogTitle.textContent).toEqual(dialogTitleValue);
```
codyarose, line 38: There's a more simplified matcher for this:
```js
expect(dialogTitle).toHaveTextContent(dialogTitleValue);
```

stories/__tests__/SQFormDialog.stories.test.js
```diff
@@ -0,0 +1,126 @@
...
+  });
+
+  it('should handle form updates and submit values on `enter` keydown', async () => {
+    render(<Default isOpen={true} onSave={handleSave} onClose={handleClose} />);
+
+    const textField = screen.getByLabelText(/hello/i);
+    userEvent.type(textField, mockData.hello);
+
+    // https://testing-library.com/docs/ecosystem-user-event/#specialchars
+    userEvent.type(textField, `${specialChars.enter}`);
```
codyarose, line 62: You can save an import and simplify this to:
```js
userEvent.type(textField, '{enter}');
```
It will automatically recognize that as the `enter` key. They should probably update that in the `testing-library.com` docs

https://github.com/testing-library/user-event#special-characters

stories/__tests__/SQFormDialog.stories.test.js
```diff
@@ -0,0 +1,126 @@
...
+
+describe('Tests for WithAutoFocus', () => {
+  it('should render a dialog and automatically focus the form input', async () => {
+    render(
+      <WithAutoFocus isOpen={true} onSave={handleSave} onClose={handleClose} />
+    );
+
+    const dialogTitleValue = 'With Auto Focus';
+    const dialogTitle = screen.getByText(dialogTitleValue);
+    expect(dialogTitle.textContent).toEqual(dialogTitleValue);
```
codyarose, line 121: Same thing here about `toHaveTextContent`, and `findByText` to get rid of the `act(...)` error

codyarose (changes requested):

stories/__tests__/SQFormDialog.stories.test.js
codyarose: One benefit of doing this is it's easier to narrow down what exactly is failing if later on this component is refactored and the tests fail. If you're testing multiple things in one test you have to then figure out which of those things is actually failing before you can fix the issue.

stories/__tests__/SQFormDialog.stories.test.js
lucas-homer: amazing. thank you!

Issue #264, opened 2021-05-21
Add tests for SQFormCheckboxGroup


laurelbean:
SQFormCheckboxGroup needs a suite of tests.

PR #285, opened 2021-05-25
test: 💍 Add SQFormCheckboxGroup Tests

laurelbean:
Adding in tests for SQFormCheckboxGroup

✅ Closes: #264

### DIFF
--- a/stories/__tests__/SQFormCheckboxGroup.stories.test.js
+++ b/stories/__tests__/SQFormCheckboxGroup.stories.test.js
@@ -42,8 +42,8 @@ describe('SQFormCheckboxGroup Tests', () => {
         SHOPPING_LIST_OPTIONS[0].label
       );
 
-      const filteredCheckboxes = checkboxes.filter(checkbox =>
-        checkbox.hasAttribute('checked')
+      const filteredCheckboxes = checkboxes.filter(
+        checkbox => checkbox.checked
       );
       expect(filteredCheckboxes).toEqual(
         expect.arrayContaining([targetCheckbox])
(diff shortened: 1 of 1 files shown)


laurelbean:
There's quite a few warnings for these tests. I already added those as issues.

stories/__tests__/SQFormCheckboxGroup.stories.test.js
```diff
@@ -0,0 +1,172 @@
...
+  Default: SQFormCheckboxGroup,
+  WithValidation: SQFormCheckboxGroupWithValidation
+} = composeStories(stories);
+
+const SHOPPING_LIST_OPTIONS = [
+  {label: 'Voltfruit', value: 'voltfruit'},
+  {label: 'Sunshroom', value: 'sunshroom'},
+  {label: 'Bokoblin Guts', value: 'bokoblin guts'},
+  {label: 'Lynel Hoof', value: 'lynel hoof'},
+  {label: 'Stealthfin Trout', value: 'stealthfin trout'}
```
lucas-homer, line 17: 🧝‍♀️ love it

lucas-homer (approved):

laurelbean:
Added in another test scenario for testing non-empty initial value

stories/__tests__/SQFormCheckboxGroup.stories.test.js
```diff
@@ -0,0 +1,191 @@
...
+      const checkboxLabel = screen.getByLabelText(
+        SHOPPING_LIST_OPTIONS[0].label
+      );
+
+      expect(checkboxLabel).toBeChecked();
+
+      const nextCheckboxLabel = screen.getByLabelText(
+        SHOPPING_LIST_OPTIONS[1].label
+      );
+      expect(nextCheckboxLabel).not.toBeChecked();
```
codyarose, line 48: I think this one could be a bit more explicit in checking which checkboxes are or are not checked. The first and second checkboxes are being tested, but we don't know for sure that the other checkboxes in the group are behaving as expected. 

So, one option would be to get all of the checkboxes, filter it down to only the checkboxes that are 'checked', then make an assertion on what should be left in that array.

```js
const initialValues = {
  shoppingList: [SHOPPING_LIST_OPTIONS[0].value]
};

render(<SQFormCheckboxGroup SQFormProps={{initialValues}} size="auto" />);

const checkboxes = screen.getAllByRole('checkbox');

const targetCheckbox = screen.getByLabelText(
  SHOPPING_LIST_OPTIONS[0].label
);

expect(checkboxes.filter(checkbox => checkbox)).toEqual(
  expect.arrayContaining([targetCheckbox])
);
```

stories/__tests__/SQFormCheckboxGroup.stories.test.js
```diff
@@ -0,0 +1,191 @@
...
+      render(<SQFormCheckboxGroup size="auto" shouldUseSelectAll />);
+
+      const selectAllCheckbox = screen.getByLabelText('All');
+      expect(selectAllCheckbox).toBeInTheDocument();
+      expect(selectAllCheckbox).not.toBeChecked();
+    });
+
+    it('should check all checkboxes when select all checkbox checked', () => {
+      render(<SQFormCheckboxGroup size="auto" shouldUseSelectAll />);
+
```
codyarose, line 95: Here's another one where getting all the checkboxes then filtering it by 'checked' would well. 
You could:

- get all the checkboxes
- expect the checkboxes array filtered by `checked` to have length `0`
- click the select all checkbox
- expect the checkboxes array filtered by `checked` to be the shopping list length + 1 (for the select all checkbox)

stories/__tests__/SQFormCheckboxGroup.stories.test.js
```diff
@@ -0,0 +1,191 @@
...
+      expect(cb1).toBeChecked();
+      expect(cb2).toBeChecked();
+      expect(cb3).toBeChecked();
+      expect(cb4).toBeChecked();
+      expect(cb5).toBeChecked();
+    });
+
+    it('should uncheck all checkboxes when select all checkbox is unchecked', () => {
+      render(<SQFormCheckboxGroup size="auto" shouldUseSelectAll />);
+
```
codyarose, line 121: Same idea here

codyarose (changes requested):

stories/__tests__/SQFormCheckboxGroup.stories.test.js
laurelbean: I have tried a few different ways to do this, but I can't find any attribute or way to actually check if the checkbox is checked. It worked in the above test, so I don't know why it won't work for this one.

codyarose: Ah, I missed the checked attribute in the filter in my previous comment. Updated that - it should have been `.filter(checkbox => checkbox.checked)`

stories/__tests__/SQFormCheckboxGroup.stories.test.js
```diff
@@ -0,0 +1,194 @@
...
+      const selectAllCheckbox = screen.getByLabelText('All');
+      const cb1 = screen.getByLabelText(SHOPPING_LIST_OPTIONS[0].label);
+      const cb2 = screen.getByLabelText(SHOPPING_LIST_OPTIONS[1].label);
+      const cb3 = screen.getByLabelText(SHOPPING_LIST_OPTIONS[2].label);
+      const cb4 = screen.getByLabelText(SHOPPING_LIST_OPTIONS[3].label);
+      const cb5 = screen.getByLabelText(SHOPPING_LIST_OPTIONS[4].label);
+
+      userEvent.click(selectAllCheckbox);
```
20BBrown14, line 132: I thnk it'd be better if we used initialValues to set the checked state for the test instead of checking them in the test.

We have a test else where that the initialValues is doing what we need so we can use it here with confidence. That way we don't have to check that each checkbox is checked in this test since this particular test doesn't really care about that. Instead it should only care about unchecking the select all.

stories/__tests__/SQFormCheckboxGroup.stories.test.js
```diff
@@ -0,0 +1,194 @@
...
+
+    it('should render an extra checkbox when adding select all capability', () => {
+      render(<SQFormCheckboxGroup size="auto" shouldUseSelectAll />);
+
+      const selectAllCheckbox = screen.getByLabelText('All');
+      expect(selectAllCheckbox).toBeInTheDocument();
+      expect(selectAllCheckbox).not.toBeChecked();
+    });
+
+    it('should check all checkboxes when select all checkbox checked', () => {
```
20BBrown14, line 96: What do we think about having a test that tests the 'selectall' checkbox in a situation where some of the options (not all) are selected to ensure that the select all isn't simply toggling the checkbox and actually doing what it should.

laurelbean: Good idea! I'll add a scenario like that

stories/__tests__/SQFormCheckboxGroup.stories.test.js
laurelbean: I have attempted this, but the select all does not seem to play nice with `initalValues`. I can't find a way to set it. I checked the html and it's showing no value (`value >`) so I think that's probably why.

20BBrown14: Hm, not sure I follow, sorry. Although if you've found an issue then it's surely one we'll want to fix in the component.

You can set the select all in initialValues by using `selectAll_[groupNameHere]: true`

laurelbean:
Thanks for all the feedback so far! New version is up

stories/__tests__/SQFormCheckboxGroup.stories.test.js
```diff
@@ -96,55 +96,97 @@ describe('SQFormCheckboxGroup Tests', () => {
...
-      expect(cb1).toBeChecked();
-      expect(cb2).toBeChecked();
-      expect(cb3).toBeChecked();
-      expect(cb4).toBeChecked();
-      expect(cb5).toBeChecked();
+      const checkedCheckboxes = checkboxes.filter(checkbox => checkbox.checked);
+      expect(checkedCheckboxes).toEqual(
+        expect.arrayContaining([selectAllCheckbox])
+      );
+      expect(checkedCheckboxes).toHaveLength(6);
```
codyarose, line 115: Looks good!
Nitpick: for the length it may be more future proof to do `SHOPPING_LIST_OPTIONS.length + 1`, just in case someone later on adds or removes options from that array

stories/__tests__/SQFormCheckboxGroup.stories.test.js
```diff
@@ -96,55 +96,97 @@ describe('SQFormCheckboxGroup Tests', () => {
...
-      expect(cb2).toBeChecked();
-      expect(cb3).toBeChecked();
-      expect(cb4).toBeChecked();
-      expect(cb5).toBeChecked();
+
+      const updatedCheckboxes = checkboxes.filter(checkbox => checkbox.checked);
+      expect(updatedCheckboxes).toEqual(
+        expect.arrayContaining([selectAllCheckbox])
+      );
+      expect(updatedCheckboxes).toHaveLength(6);
```
codyarose, line 150: Nitpick: Same thing here on the list length

stories/__tests__/SQFormCheckboxGroup.stories.test.js
```diff
@@ -96,55 +96,97 @@ describe('SQFormCheckboxGroup Tests', () => {
...
+      );
+
+      const checkboxes = screen.getAllByRole('checkbox');
+
+      const selectAllCheckbox = screen.getByLabelText('All');
+      const checkedCheckboxes = checkboxes.filter(checkbox => checkbox.checked);
+      expect(checkedCheckboxes).toEqual(
+        expect.arrayContaining([selectAllCheckbox])
+      );
+      expect(checkedCheckboxes).toHaveLength(6);
```
codyarose, line 179: Nitpick: Same thing here on the list length

stories/__tests__/SQFormCheckboxGroup.stories.test.js
```diff
@@ -96,55 +96,97 @@ describe('SQFormCheckboxGroup Tests', () => {
...
-      expect(cb3).not.toBeChecked();
-      expect(cb4).not.toBeChecked();
-      expect(cb5).not.toBeChecked();
+      const uncheckedCheckboxes = checkboxes.filter(
+        checkbox => !checkbox.checked
+      );
+      expect(uncheckedCheckboxes).toEqual(
+        expect.arrayContaining([selectAllCheckbox])
+      );
+      expect(uncheckedCheckboxes).toHaveLength(6);
```
codyarose, line 189: Nitpick: Same thing here on the list length

codyarose (approved):
Nice job on these!

PR #287, opened 2021-05-26
test: 💍 SQFormMultiSelect

lucas-homer:
✅ Closes: 270

### DIFF
--- a/stories/SQFormMultiSelect.stories.js
+++ b/stories/SQFormMultiSelect.stories.js
@@ -23,8 +23,8 @@ export default {
 };
 
 const MOCK_FRIENDS_OPTIONS = [
-  {label: 'Joe', value: random(10 + Math.ceil(Math.random() * 20))},
-  {label: 'Jane', value: random(10 + Math.ceil(Math.random() * 20))},
+  {label: 'Joe', value: 1},
+  {label: 'Jane', value: 2},
   {label: 'Jack', value: random(10 + Math.ceil(Math.random() * 20))},
   {label: 'Jill', value: random(10 + Math.ceil(Math.random() * 20))},
   {label: 'John', value: random(10 + Math.ceil(Math.random() * 20))},
--- a/stories/SQFormMultiSelect.stories.js
+++ b/stories/SQFormMultiSelect.stories.js
@@ -47,18 +47,24 @@ const defaultArgs = {
   children: MOCK_FRIENDS_OPTIONS
 };
 
-export const SQFormMultiSelect = args => (
-  <div style={{minWidth: 250}}>
-    <SQFormStoryWrapper initialValues={{[defaultArgs.name]: []}}>
-      <SQFormMultiSelectComponent
-        name={defaultArgs.name}
-        label={defaultArgs.label}
-        {...args}
-        size={args.size !== 'auto' ? Number(args.size) : args.size}
-      />
-    </SQFormStoryWrapper>
-  </div>
-);
+export const SQFormMultiSelect = args => {
+  const {SQFormProps, ...rest} = args;
+  return (
+    <div style={{minWidth: 250}}>
+      <SQFormStoryWrapper
+        initialValues={{[defaultArgs.name]: []}}
+        {...SQFormProps}
+      >
+        <SQFormMultiSelectComponent
+          name={defaultArgs.name}
+          label={defaultArgs.label}
+          {...rest}
+          size={rest.size !== 'auto' ? Number(rest.size) : rest.size}
+        />
+      </SQFormStoryWrapper>
+    </div>
+  );
+};
 SQFormMultiSelect.storyName = 'SQFormMultiSelect';
 SQFormMultiSelect.args = defaultArgs;
 
--- a/stories/SQFormMultiSelect.stories.js
+++ b/stories/SQFormMultiSelect.stories.js
@@ -43,7 +43,8 @@ const MOCK_FRIENDS_OPTIONS = [
 
 const defaultArgs = {
   label: 'Friends',
-  name: 'friends'
+  name: 'friends',
+  children: MOCK_FRIENDS_OPTIONS
 };
 
 export const SQFormMultiSelect = args => (
@@ -54,9 +55,7 @@ export const SQFormMultiSelect = args => (
         label={defaultArgs.label}
         {...args}
         size={args.size !== 'auto' ? Number(args.size) : args.size}
-      >
-        {MOCK_FRIENDS_OPTIONS}
-      </SQFormMultiSelectComponent>
+      />
     </SQFormStoryWrapper>
   </div>
 );
--- a/stories/__tests__/SQFormMultiSelect.stories.test.js
+++ b/stories/__tests__/SQFormMultiSelect.stories.test.js
@@ -39,14 +39,29 @@ describe('Tests for SQFormMultiSelect', () => {
     expect(expandButton).toHaveTextContent(initialDropdownValue);
   });
 
-  it('should show empty value "- -" if provided initial value not in options', async () => {
-    render(<SQFormMultiSelect size="auto" initialValue={{friends: [0]}} />);
+  it('should show console warning if provided initial value not in options', async () => {
+    const consoleWarnSpy = jest
+      .spyOn(console, 'warn')
+      .mockImplementation(() => {});
 
-    const expandButton = await waitFor(() =>
-      screen.findByRole('button', {name: /friends/i})
+    render(
+      <SQFormMultiSelect
+        size="auto"
+        SQFormProps={{
+          initialValues: {friends: [0]}
+        }}
+      />
     );
 
-    expect(expandButton).toHaveTextContent(initialDropdownValue);
+    await waitFor(() =>
+      expect(consoleWarnSpy).toHaveBeenCalledWith(
+        expect.stringMatching(
+          /the requested display value.*could not be found/i
+        )
+      )
+    );
+
+    consoleWarnSpy.mockRestore();
   });
 
   it('should show empty list if no options passed', async () => {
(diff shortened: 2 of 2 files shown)


stories/__tests__/SQFormMultiSelect.stories.test.js
```diff
@@ -0,0 +1,142 @@
...
+    const expandButton = await waitFor(() =>
+      screen.findByRole('button', {name: /friends/i})
+    );
+    await waitFor(() =>
+      expect(expandButton).toHaveTextContent(initialDropdownValue)
+    );
+  });
+
+  it('should show empty value "- -" if provided initial value not in options', async () => {
+    render(<SQFormMultiSelect size="auto" initialValue={{friends: [0]}} />);
```
codyarose, line 46: `initialValues` isn't a prop on `SQFormMultiSelect`, but I think I see what you were going for. In some of the other tests you can pass in different `initialValues` to the underlying `SQForm` because the story was set up to make that possible with `SQFormProps`. This story does not have that set up because it existed before that solution was created.

Check out the stories and tests for `SQFormAutocomplete` to get an idea of how to set that up. Feel free to ping me if you have any questions - the storybook template and args structure is kind of weird.

After that, a test for checking what should happen when an initial value is provided that _is_ in the options would be good to add.

stories/__tests__/SQFormMultiSelect.stories.test.js
```diff
@@ -0,0 +1,142 @@
...
+
+    consoleWarnSpy.mockRestore();
+  });
+
+  it('should update when options are selected', async () => {
+    render(<SQFormMultiSelect size="auto" />);
+
+    const expandButton = await screen.findByRole('button', {name: /friends/i});
+
+    expect(expandButton).toHaveTextContent('- -');
```
codyarose, line 81: Nitpick: `initialDropdownValue` could be used here

stories/__tests__/SQFormMultiSelect.stories.test.js
```diff
@@ -0,0 +1,142 @@
...
+    userEvent.click(option2);
+
+    await waitFor(() => expect(expandButton).toHaveTextContent('Jack, Jill'));
+  });
+
+  it('should not open dropdown options if field is disabled', async () => {
+    render(<SQFormMultiSelect size="auto" isDisabled={true} />);
+
+    const expandButton = await screen.findByRole('button', {name: /friends/i});
+    expect(expandButton).toHaveAttribute('aria-disabled');
```
codyarose, line 98: Check out the `SQFormDropdown` test called `should not be selectable if it is disabled` for a good example of how to test that it's truly disabled. I think the same methods should apply to the multiselect

stories/__tests__/SQFormMultiSelect.stories.test.js
```diff
@@ -0,0 +1,142 @@
...
+  it('should highlight field if required but no value selected', async () => {
+    const validationSchema = {
+      friends: Yup.string().required('Required')
+    };
+
+    render(
+      <SQFormMultiSelect
+        size="auto"
+        isRequired
+        validationSchema={validationSchema}
```
codyarose, line 124: Same as my previous comment about `initialValues` here. `SQFormProps` needs to be exposed on the story to pass in a schema to `SQForm`

codyarose (changes requested):

stories/__tests__/SQFormMultiSelect.stories.test.js
```diff
@@ -0,0 +1,142 @@
...
+  it('should render a form with a multiselect input and dropdown options', async () => {
+    render(<SQFormMultiSelect size="auto" />);
+
+    const expandButton = await waitFor(() =>
+      screen.findByRole('button', {
+        name: /friends/i
+      })
+    );
+
+    userEvent.click(expandButton);
```
laurelbean, line 25: Since you're performing an action after the render, checking for the options could be moved to its own test. But I can also see how having the options is essential to the component.

stories/__tests__/SQFormMultiSelect.stories.test.js
```diff
@@ -0,0 +1,142 @@
...
+import {composeStories} from '@storybook/testing-react';
+
+import userEvent from '@testing-library/user-event';
+import * as stories from '../SQFormMultiSelect.stories';
+
+const {SQFormMultiSelect} = composeStories(stories);
+
+const initialDropdownValue = '- -';
+
+afterEach(() => {});
```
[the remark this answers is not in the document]
lucas-homer: bingo! updating now...

stories/__tests__/SQFormMultiSelect.stories.test.js
```diff
@@ -0,0 +1,142 @@
...
+        name: /friends/i
+      })
+    );
+
+    userEvent.click(expandButton);
+
+    const optionsList = await screen.findByRole('listbox');
+    await waitFor(() => expect(optionsList).toBeInTheDocument());
+
+    const options = await within(optionsList).findAllByRole('option');
```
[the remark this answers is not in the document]
lucas-homer: ok cool. honestly, at a certain point, I went with the shotgun approach cause I couldn't tell which part actually needed it 😬

stories/__tests__/SQFormMultiSelect.stories.test.js
```diff
@@ -0,0 +1,140 @@
...
+import userEvent from '@testing-library/user-event';
+import * as stories from '../SQFormMultiSelect.stories';
+
+const {SQFormMultiSelect} = composeStories(stories);
+
+const initialDropdownValue = '- -';
+
+describe('Tests for SQFormMultiSelect', () => {
+  it('should render a form with a multiselect input and dropdown options', async () => {
+    render(<SQFormMultiSelect size="auto" />);
```
20BBrown14, line 15: nit: Size defaults to 'auto'

stories/__tests__/SQFormMultiSelect.stories.test.js
```diff
@@ -0,0 +1,140 @@
...
+  });
+
+  it('should display icon and text if field is required', async () => {
+    render(<SQFormMultiSelect isRequired={true} size="auto" />);
+
+    const required = await screen.findByText(/required/i);
+    expect(required).toBeVisible();
+  });
+
+  it('should not display icon and text if field is not required', async () => {
```
20BBrown14, line 106: This test doesn't check if the icon is present as the description says

stories/__tests__/SQFormMultiSelect.stories.test.js
laurelbean: This is an issue with the story, not the component. Currently, if you don't add size to the story then it'll throw a warning.

20BBrown14: Thanks, I wasn't aware!

laurelbean: Cody fixed it on the newest set of stories, but the fix hasn't been ported to the old ones yet.

codyarose: I _think_ I've fixed it. I believe the issue is with storybook changing numbers to strings when it passes the size prop along through `args`, which causes proptypes to throw a fit

stories/__tests__/SQFormMultiSelect.stories.test.js
lucas-homer: this is a good point, and after reviewing, I'm changing the description because I didn't have a good way of checking for the svg anyway, and it didn't seem absolutely necessary for the test to still keep its usefulness

laurelbean:
I think it would be good to add another test for an initial value that is part of the options.

lucas-homer:
> I think it would be good to add another test for an initial value that is part of the options.

done!

stories/__tests__/SQFormMultiSelect.stories.test.js
lucas-homer: > Since you're performing an action after the render, checking for the options could be moved to its own test. But I can also see how having the options is essential to the component.

I'm going to go ahead and split this into two. For extra by-the-bookness

laurelbean (approved):
Looks good to me!

stories/__tests__/SQFormMultiSelect.stories.test.js
lucas-homer: done!

stories/__tests__/SQFormMultiSelect.stories.test.js
```diff
@@ -0,0 +1,189 @@
...
+    userEvent.click(option2);
+
+    await waitFor(() => expect(expandButton).toHaveTextContent('Jack, Jill'));
+  });
+
+  it('should not open dropdown options if field is disabled', async () => {
+    render(<SQFormMultiSelect size="auto" isDisabled={true} />);
+
+    const expandButton = await screen.findByRole('button', {name: /friends/i});
+    expect(expandButton).toHaveAttribute('aria-disabled', 'true');
```
codyarose, line 144: I think the only thing that's missing is checking that the dropdown really won't open when clicked while disabled
```js
userEvent.click(expandButton);

const optionsList = screen.queryByRole('listbox');

expect(optionsList).not.toBeInTheDocument();
```

stories/SQFormMultiSelect.stories.js
```diff
@@ -43,23 +43,28 @@ const MOCK_FRIENDS_OPTIONS = [
...
-        label={defaultArgs.label}
-        {...args}
-        size={args.size !== 'auto' ? Number(args.size) : args.size}
+export const SQFormMultiSelect = args => {
+  const {SQFormProps, ...rest} = args;
+  return (
+    <div style={{minWidth: 250}}>
+      <SQFormStoryWrapper
+        initialValues={{[defaultArgs.name]: []}}
+        {...SQFormProps}
```
codyarose, line 56: 👍

codyarose (changes requested):

stories/__tests__/SQFormMultiSelect.stories.test.js
lucas-homer: > I think the only thing that's missing is checking that the dropdown really won't open when clicked while disabled
> 
> ```js
> userEvent.click(expandButton);
> 
> const optionsList = screen.queryByRole('listbox');
> 
> expect(optionsList).not.toBeInTheDocument();
> ```

awesome feedback. done!

codyarose (approved):

Issue #273, opened 2021-05-21
SQFormCheckboxGroup Failed Prop Type Error


laurelbean:
Loading the SQFormCheckboxGroup story immediately creates 2 failed prop type warnings in the console for `Warning: Failed prop type: Invalid prop 'checked' of type 'string' supplied, expected boolean`

Loom showing issue: https://www.loom.com/share/fa1c8f223b9946c0a218b5cec1a96930

PR #292, opened 2021-06-01
fix: 🐛 SQFormCheckboxGroup no longer has failed prop error

laurelbean:
Changed the initial value of the checkbox group in the story to be an
empty array instead of a string

✅ Closes: #273


SeanGroff:
Since this only effects storybook let's redo this so it's a chore commit. We don't want to bump the npm version for a code change consumers don't receive.

Issue #226, opened 2021-05-04
Form component with a header, scrollable body, and sticky footer


SeanGroff:
We have a common design pattern that began in our Quote Tools and is now the same design we'll use for Ancillary apps where a whole page is a form.

I've created a POC and ran it by the internal FE team. https://codesandbox.io/s/sqform-full-page-prototype-keisd?file=/src/SQFormFullPage.js

![Screen Shot 2021-05-04 at 9 54 46 AM](https://user-images.githubusercontent.com/4367312/117023480-c4c5e780-acbe-11eb-9371-46eee7c85b6b.png)

The component will likely be named something like SQFormScrollCard, tbd.

A/C

* The form should match the designs
* The form should always take up 100% of it's parent elements available height by making the content area grow/shrink
* Content area should scrollable vertically
* The footer should stick to the bottom, and have 3 components from left to right. Reset Button, InfoText, Submit Button

SeanGroff:
I've been working on this as I'm able to. I know 2 projects are waiting on this component. The Srx project will need this as well. I'll have it released this week.
The component is finished and works within Storybook. 
Next, I need to test this component out on the Data Capture project and the MCE to ensure it works for both use cases. 
Will PR after that.

![Screen Shot 2021-05-17 at 5 01 51 PM](https://user-images.githubusercontent.com/4367312/118562029-95d55a00-b731-11eb-9919-1af45c1ae493.png)

SeanGroff:
Progress Update since I thought I'd be finished by now. 

This component worked perfectly on the Data Capture Project.
Consuming this component on the MCE uncovered some bugs, some of which I still need to figure solutions out for:

Issue 1 - MCE Doesn't specify a static height anywhere. Not a bad thing at all. However, `overflow-y: scroll` will not work unless a static height is specified. Currently, the consumer will have to wrap this component in an element with a static height. 

Issue 2 - The `validationSchema` prop causes an error if we pass it a POJO on the MCE. Currently, we have to pass it a Yup.object().shape() which SQFormScrollable component already does for us.

Issue 3 - When providing a subheader, the content must indent `20px` more.

Further remaining work:

* Write stories in Storybook.
* Write tests (This can be handled in a separate issue that doesn't block this work)

SeanGroff:
I collaborated with @gdoohen1 to bring him up to speed on this feature. He's going to pick up where I left off and get this across the finish line 🏁 
Assigning issue to Grant.

[...]

PR #294, opened 2021-06-01
feat: 🎸 ScrollableCard that always maintains full height

gdoohen1:
✅ Closes: #226 

Hey there, this is my first PR to SQForm so apologies if I miss anything.

Almost all of this was done by @SeanGroff so I can't take much credit for it.

I documented the changes I made on the issue itself https://github.com/SelectQuoteLabs/SQForm/issues/226#issuecomment-849923366


src/components/SQFormScrollableCard/SQFormScrollableCard.js
```diff
@@ -0,0 +1,212 @@
...
+              </Card>
+            </Form>
+          );
+        }}
+      </Formik>
+    </div>
+  );
+}
+
+SQFormScrollableCard.propTypes = {
```
20BBrown14, line 167: Looks like we might be missing some of the prop type definitions

SeanGroff:
Thanks, @gdoohen1 💯 

Can you record a Loom video to demo this working in Storybook and the MCE?
Since I wrote a lot of this code, it wouldn't be fair for me to review this PR.

SeanGroff (changes requested):
The storybook story was a placeholder story I wrote to get something on the page. Can you create a story that shows this component working without a subheader and a story of it WITH a subheader?

stories/SQFormScrollableCard.stories.js
```diff
@@ -0,0 +1,135 @@
...
+    alignItems: 'center'
+  },
+  onSubmit: () => {},
+  validationSchema: {hello: Yup.string().required('Required')}
+};
+
+const Template = args => {
+  return (
+    <SQFormScrollableCard {...args}>
+      <SQFormTextField
```
20BBrown14, line 34: Could we add different fields in here to see how they look/work in this context?

stories/SQFormScrollableCard.stories.js
```diff
@@ -0,0 +1,135 @@
+import React from 'react';
```
20BBrown14, line 1: Do we need to add tests for this new component?

SeanGroff: > Do we need to add tests for this new component?

Good question. If a test exists for a component, then we should add/modify the tests to account for the incoming code.
If a test does not yet exist for a component or a brand new component, we can handle writing the test in a separate issue until we all are more comfortable writing tests.

20BBrown14: Awesome. We should open an issue then to ensure we don't forget.

gdoohen1:
[Here is a loom](https://www.loom.com/share/aaf49c69a0f54b4d8da7755e5b029bf8) demoing the ScrollableCard in the senior-age-ui. If you want to test this yourself pull down branch `scrollable-card-test`, delete your `package-lock` and `node_modules` and run `npm i`.

src/components/SQFormScrollableCard/SQFormScrollableCard.js
gdoohen1: Good catch, added!

stories/SQFormScrollableCard.stories.js
gdoohen1: I did add some more stories to show off some of the more important props of the component but I didn't add any variance of fields since all this component does is put them in a `<Form />` so I don't think it would add any value as far as this component is concerned.

stories/SQFormScrollableCard.stories.js
```diff
@@ -0,0 +1,185 @@
...
+  initialValues: {hello: ''},
+  muiGridProps: {
+    spacing: 2,
+    alignItems: 'center'
+  },
+  onSubmit: () => {},
+  validationSchema: {hello: Yup.string().required('Required')}
+};
+
+const Template = (args = {}) => {
```
codyarose, line 32: This default object assignment needs to be removed to allow the Storybook Controls to work

codyarose (changes requested):

codyarose (approved):

SeanGroff:
> Here is a loom

Thanks for the loom, super helpful!

SeanGroff (approved):

PR #298, opened 2021-06-02
feat: 🎸 fix test for icon button

saadurrehman-selectquote:
feat: 🎸 fix test for icon button

fix test for icon button

✅ Closes: 297

image
image

### DIFF
--- a/stories/__tests__/SQFormIconButton.stories.test.js
+++ b/stories/__tests__/SQFormIconButton.stories.test.js
@@ -68,7 +68,11 @@ describe('SQFormIconButton Tests', () => {
       const svg = within(iconButton).getByTitle(/Form Submission/i);
 
       expect(iconButton).toBeInTheDocument();
-      expect(svg).toHaveStyle('color: var(--color-palmLeaf)');
+      expect(svg).toHaveStyle({
+        color: 'var(--color-teal)',
+        width: '2.5rem',
+        height: '2.5rem'
+      });
     });
 
     it('should call a function when clicked', () => {


SeanGroff:
For context, I asked Saad to make the conventional commit a `feat` because his previous PR didn't follow the conventional commit pattern so the feature was never published to npm.

stories/__tests__/SQFormIconButton.stories.test.js
```diff
@@ -68,7 +68,11 @@ describe('SQFormIconButton Tests', () => {
       const svg = within(iconButton).getByTitle(/Form Submission/i);
 
       expect(iconButton).toBeInTheDocument();
-      expect(svg).toHaveStyle('color: var(--color-palmLeaf)');
+      expect(svg).toHaveStyle({
+        color: 'var(--color-teal)',
+        width: '2.5rem',
+        height: '2.5rem'
+      });
```
SeanGroff, line 75: @codyarose @laurelbean Is this testing implementation details? Thoughts?

saadurrehman-selectquote: we had to add all styles because it requires all elements of style
image
https://github.com/testing-library/jest-dom#tohavestyle

codyarose: @SeanGroff Yeah, this is all implementation details. If we need to test that the color is correct, then we pretty much have to test implementation details.

However, I have a feeling that there's a better way to do this, but I need to do some research on it. Essentially I'm thinking we can get the css-in-js value from `makeStyles` then snapshot test that.

SeanGroff: https://jestjs.io/docs/snapshot-testing

This looks like the route we'd want to take to properly test this.
However, I don't want to hold this up. I'm going to merge this for now and create a new issue to explorer snapshot testing situations like this where we want to ensure the correct styles apply based on a components props.

SeanGroff (approved):

Issue #301, opened 2021-06-08
Migrate components/LoadingIcon/* to Typescript


caleboleary:
Converting the contents of LoadingIcon folder to TS - ref https://github.com/SelectQuoteLabs/SQForm/issues/244

PR #302, opened 2021-06-08
refactor: 💡 convert LoadingIcon to typescript Ref #244

caleboleary:
✅ Closes: #301

I think we're still developing our style for this. 

- I've set the component file to .TSX rather than .TS. 
- I've named the props interface as componentNameProps. 
- I've placed the interface at the bottom of the file, mirroring what we've done with proptypes in the past. 
- I've left the comment in that was providing more descriptions to the proptypes. 
- I've set the return type as `JSX.Element`. 
 
Please discuss in this PR if you disagree or have opinions on any of those patterns. (Or, if we've declared contrary styles elsewhere and I'm blind and have missed them =])

Here's a screenshot of the loading icon working in storybook after this change:
Screen Shot 2021-06-08 at 5 09 00 PM


SeanGroff:
I agree with the opinions you listed. I've always declared the Props interface towards the top of the file. Unlike prop types. Not sure why, just how I've always seen it done.

Since we have a hard rule of one component per file. You're safe to just name the interface Props.

src/components/LoadingIcon/LoadingIcon.tsx
```diff
@@ -97,9 +96,9 @@ const LoadingIcon = ({height = '10rem'}) => {
   );
 };
 
-LoadingIcon.propTypes = {
+interface LoadingIconProps {
   /** The height of the loading icon */
-  height: PropTypes.string
-};
+  height: string;
```
codyarose, line 101: This prop should be marked as optional since it's given a default value on line 41

codyarose (changes requested):

codyarose:
I also put the interface just above the component, rather than at the bottom like prop-types; I mostly prefer it that way because it's collocated to where the interface is used. I think that makes it easier to reference the component's prop types because it's right above where the props are listed, if it's at the bottom the interface gets further and further away as the component grows.

Agreed on the interface naming as well; for some reason lately I've been naming them `ComponentNameProps`, but if it's a single file component and it's not being exported, I can't think of any valid reason not to just name it `Props`.

caleboleary:
@codyarose good catch, thank you

I went ahead and moved the interface up since that seems to be the consensus forming here.

I've not yet changed ComponentNameProps to Props, but if that's what we settle on, I'm happy to.

20BBrown14 (approved):

SeanGroff:
@caleboleary We need to fix the lint warnings. Those can be seen in the Files Changed tab above. I also posted in our private channel about someone adjusting our lint script and pre-commit tool to run against `ts` and `tsx` files. Currently, only `js` files are being ran through the linter.

caleboleary:
@SeanGroff I'm a bit confused about the type warnings on js files? What should I be doing about those?

SeanGroff:
> @SeanGroff I'm a bit confused about the type warnings on js files? What should I be doing about those?

Convert them to TS duh! J/k, we'll need to configure our linter to only run the TypeScript lint rules against ts & tsx files.

SeanGroff:
@caleboleary Where are we at with resolving the lint warnings?
I think this SO post shows how to handle filetype overrides.
https://stackoverflow.com/questions/60773362/how-to-lint-js-and-ts-files-with-different-eslint-configs-at-the-same-time

caleboleary:
@SeanGroff I'll be honest, I thought someone else was already looking into it haha. If no one is, I will give it a go, thanks for the link

SeanGroff:
> @SeanGroff I'll be honest, I thought someone else was already looking into it haha. If no one is, I will give it a go, thanks for the link

No worries, I just happened to notice this PR again.
Yea if you could include the linting config fixes to this PR that would be 💯

.eslintrc.json
```diff
@@ -30,10 +27,42 @@
...
+      "parser": "@typescript-eslint/parser",
+      "parserOptions": {
+        "ecmaFeatures": {"jsx": true},
+        "ecmaVersion": 2018,
+        "sourceType": "module",
+        "project": "./tsconfig.json"
+      },
+      "plugins": ["@typescript-eslint"],
+      "rules": {
+        "no-useless-computed-key": "off",
```
caleboleary, line 51: I don't really know what I'm doing here. Copied all of the rules from above because it sounded like a good idea? The rest of this overrides section came from stackoverflow. @codyarose @SeanGroff please look this over =]

.eslintrc.json
```diff
@@ -30,10 +27,42 @@
...
+  "overrides": [
+    {
+      "files": ["**/*.ts", "**/*.tsx"],
+      "env": {"browser": true, "es6": true, "node": true},
+      "extends": [
+        "eslint:recommended",
+        "plugin:@typescript-eslint/eslint-recommended",
+        "plugin:@typescript-eslint/recommended"
+      ],
+      "globals": {"Atomics": "readonly", "SharedArrayBuffer": "readonly"},
```
SeanGroff, line 41: I'm not familiar with these globals? What are they or can we delete them and everything still works?

.eslintrc.json
```diff
@@ -30,10 +27,42 @@
...
+      "parser": "@typescript-eslint/parser",
+      "parserOptions": {
+        "ecmaFeatures": {"jsx": true},
+        "ecmaVersion": 2018,
```
SeanGroff, line 45: will this accept `esnext` instead of 2018?
Guessing 2018 was around the time the SO post was created

.eslintrc.json
SeanGroff: I believe you want to copy rules that are TypeScript related here.
I think we need to move them all back accept 
```
"no-use-before-define": "off",
 "@typescript-eslint/no-use-before-define": ["error"]
 ```

SeanGroff (changes requested):

.eslintrc.json
caleboleary: Doesn't look like it: 
`Value is not accepted. Valid values: 3, 5, 6, 2015, 7, 2016, 8, 2017, 9, 2018, 10, 2019, 11, 2020, 12, 2021.`

SeanGroff: 2021 it is 🚀

SeanGroff (approved):

codyarose (approved):

Issue #174, opened 2021-04-01
Boolean Valued Dropdown options


hvilander:
Right now SQFormDropDown accepts strings and numbers for values of options. Since there are a ton of yes/no dropdowns it may be good to accept boolean values.

The docs should be updated maybe with a validation example. When I asked about why we don't accept booleans someone mentioned the false value would break validation. 

https://github.com/SelectQuoteLabs/SQForm/blob/f2070904b3f4cb556bed70d2bfb265116fa0cc2c/src/components/SQForm/SQFormDropdown.js#L129

hvilander:
@codyarose I intended to get back around to this, I have not yet.

PR #305, opened 2021-06-16
feat: 🎸 support for boolean valued dropdown

hvilander:
Adds support for boolean valued dropdowns; modifies _getHasValues on
useForm to account for boolean values

✅ Closes: #174

Please watch the [Loom](https://www.loom.com/share/5a258dd945c343158fdabaee439d1d9c) as it brings up a potential for some more changes.

### DIFF
--- a/src/components/SQForm/SQFormDropdown.js
+++ b/src/components/SQForm/SQFormDropdown.js
@@ -138,7 +138,11 @@ SQFormDropdown.propTypes = {
   children: PropTypes.arrayOf(
     PropTypes.shape({
       label: PropTypes.string,
-      value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+      value: PropTypes.oneOfType([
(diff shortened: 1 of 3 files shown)


SeanGroff (approved):
Nice catch!
I'm not really concerned with the submitting empty string scenario.

src/components/SQForm/SQFormDropdown.js
```diff
@@ -138,7 +138,11 @@ SQFormDropdown.propTypes = {
   children: PropTypes.arrayOf(
     PropTypes.shape({
       label: PropTypes.string,
```
20BBrown14, line 140: `value` and `label` should be required here.

We reuse this propType in so many places, I almost wish we had it defined as a constant somewhere.

20BBrown14 (approved):
The dropdown being able to submit with an empty string, on a non-required field, would be expected.

Issue #303, opened 2021-06-15
New Autocomplete, Multiselect, free solo component


20BBrown14:
I'm proposing we add a new component that is autocomplete-able, multiselectable, and free solo. We have a use case in our project where we could have a list of > 100 options that is best filtered with autocomplete. The use case allows for multiple selects _and_ the list of options is not extensive we need an option to allow users to input their own option.

The "free solo" discussion has been had in #249 but this would be a brand new component.

An example of it from MUI is here https://material-ui.com/components/autocomplete/#multiple-values

New name: SQFormMultiValue (???)

PR #306, opened 2021-06-16
feat: 🎸 Add SQFormMultiValue component

20BBrown14:
Add new component, SQFormMultiValue

Loom: https://www.loom.com/share/e2405d5fbe1a42b7a7d0b457132c5593

Sorry for the loom length :/ the last half is just code, though, if you care about that.

✅ Closes: #303

**When this is approved I'll add an issue to add tests for this component**

Screen Shot 2021-06-17 at 8 57 34 AM


hvilander:
Listening to your loom, what happens if I type in a string that is equal to the `value` of an option.

Example: options = { red: 1, blue: 2} and I type in 2.

20BBrown14:
> Listening to your loom, what happens if I type in a string that is equal to the `value` of an option.
> 
> Example: options = { red: 1, blue: 2} and I type in 2.

Good question. So for your example, nothing. BEcause 2 and "2" are different. But if you had { red: 'color-red', blue: 'color-blue' } and you typed in "color-blue" you'll end up with just "Blue" in the options

seen here: https://www.loom.com/share/8f331a81f57c4bf5bb533cbc0c03d126

src/components/SQForm/SQFormMultiValue.js
```diff
@@ -0,0 +1,271 @@
...
+
+// Adapter for react-window
+const ListboxVirtualizedComponent = React.forwardRef(
+  function ListboxVirtualizedComponent(props, ref) {
+    const {children, ...listboxProps} = props;
+    const LIST_MAX_VIEWABLE_ITEMS = 8;
+    const LIST_OVERSCAN_COUNT = 5;
+    const items = React.Children.toArray(children);
+    const ITEM_COUNT = items.length;
+    const ITEM_SIZE = 36;
```
hvilander, line 36: Should these constants live with the one on line 10?

hvilander (approved):

stories/SQForm.stories.js
```diff
@@ -197,6 +211,13 @@ export const BasicForm = () => {
         >
           {CHECKBOX_GROUP_OPTIONS}
         </SQFormCheckboxGroup>
+        <SQFormMultiValue
+          name="favoriteColors"
+          label="Your Favorite Colors"
+          size={12}
```
Chris-Boyle, line 217: Size 12 is a little extreme.  I don't think we will be doing that large of a field in any of our apps so we probably shouldn't do that in our Storybook as well.  We want to keep the stories as close to real world examples as possible.

stories/SQForm.stories.js
```diff
@@ -29,7 +29,8 @@ import {
   SQFormRadioButtonGroup,
   SQFormCheckboxGroup,
   SQFormMaskedTextField,
```
Chris-Boyle, line 31: [nit] move `SQFormMultiValue` above the MASKS to make the imports look more alphabetical. (I know they all aren't perfect)

Chris-Boyle:
What happens if your selected options take up more width than the field?

package.json
```diff
@@ -134,7 +134,7 @@
     "prettier": "^1.18.2",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "react-scripts": "^4.0.0",
+    "react-scripts": "^3.4.3",
```
SeanGroff, line 137: Blocker - we lose things like React Fast Refresh with this downgrade.
We have full control of our linter. Fix the specific issue, preferably in another PR.
No need for this to get in the way of the SQFormMultiValue PR

src/components/SQForm/SQFormMultiValue.js
```diff
@@ -0,0 +1,271 @@
...
+    const height = React.useMemo(() => {
+      if (ITEM_COUNT > LIST_MAX_VIEWABLE_ITEMS) {
+        return LIST_MAX_VIEWABLE_ITEMS * ITEM_SIZE;
+      }
+      return (
+        items.reduce((acc, _item) => acc + ITEM_SIZE, 0) + 2 * LISTBOX_PADDING
+      );
+    }, [ITEM_COUNT, items]);
+
+    const getItemSize = React.useCallback(() => ITEM_SIZE, []);
```
SeanGroff, line 47: I know this might be copy/pasta from the MUI docs example, but if this only returns `ITEM_SIZE` does this need to be a function?

vs passing `ITEM_SIZE` to the `itemSize` prop on line 59?

stories/SQForm.stories.js
```diff
@@ -299,6 +328,14 @@ export const FormWithValidation = () => {
         >
           {CHECKBOX_GROUP_OPTIONS}
         </SQFormCheckboxGroup>
+        <SQFormMultiValue
+          name="favoriteColors"
+          label="Your Favorite Colors"
+          size={12}
```
SeanGroff, line 334: Storybook is great for what I call UI chaos testing.
I question what happens if we change this to size `6` then select all of the options. What happens?

SeanGroff (changes requested):

stories/SQForm.stories.js
SeanGroff: +1
I've never realized how wide that form gets if the right side bar in Storybook is hidden.
Branden all up in here with the ultra wide lol.

Can we make this a 4 or something? To try and get a realistic view of how it'll behave in our apps.
Or maybe we add a max-width to the form wrapper so it doesn't grow so wide.

src/components/SQForm/SQFormMultiValue.js
20BBrown14: tbh idk, i copied this from the autocomplete component. i can try to clean it up though.

stories/SQForm.stories.js
20BBrown14: Good question And so is

> What happens if your selected options take up more width than the field?

It wraps the options and grows vertically when the options are more than allowed in the width. if a single option is longer than the width of the field it truncates the option label.

Screen Shot 2021-06-17 at 8 51 02 AM

Chris-Boyle: @20BBrown14  what is that little x all by itself on the right there?  I didn't see that in the Loom.  Does that clear all the fields?  
image

src/components/SQForm/SQFormMultiValue.js
20BBrown14: Turns out this does need to be a function, actually. `getItemSize` is being passed to `VariableSizeList` from react-window for the `itemSize` prop which requires a function.

stories/SQForm.stories.js
20BBrown14: Screen Shot 2021-06-17 at 8 57 34 AM

src/components/SQForm/SQFormMultiValue.js
SeanGroff: These constants are only used in this closure so they were scoped to this closure

stories/SQForm.stories.js
20BBrown14: I pointed that out in the loom as a way to clear the field at about 43 seconds

SeanGroff: > I pointed that out in the loom as a way to clear the field at about 43 seconds
> 
> edit: and is also how SQFormAutocomplete works so our users and devs should be familiar.

Yep. It's a great feature to have! I think you'll want to run the clear button by Nemeth. My hunch is he's going to want it to always stick to the bottom regardless of the number of rows so it's always in the same spot near the inputs bottom border. The default from MUI sort of looks likes its floating in no mans land.

20BBrown14: Nemeth wanted it on the bottom, this is what i was able to get 
Screen Shot 2021-06-17 at 11 07 36 AM
Screen Shot 2021-06-17 at 11 07 22 AM

Chris-Boyle:
We are missing component documentation for this component.  Adding the validation example and reasoning would be good in the documentation.

20BBrown14:
> We are missing component documentation for this component. Adding the validation example and reasoning would be good in the documentation.

I'm not sure what you're referring. Is there some documentation for the components I wish i knew about several months ago?

Chris-Boyle:
We do notes on some components.  We used to do it a lot more.  I am not sure where the stance is but it is nice to have for things that need a little more explanation.
https://github.com/SelectQuoteLabs/SQForm/tree/master/notes
Here is an example
https://5f4431386ea00a00220d495c-qfuecrbuoq.chromatic.com/?path=/docs/components-sqformdatepicker--date-picker-before-2024

SeanGroff (approved):
Fantastic work @20BBrown14!

SeanGroff:
Great job building this component @20BBrown14 , accepting PR feedback, and turning the feedback into a polished solution 💯  

I stamped my approval, pending the removal of the package-lock.json that's still there.

Chris-Boyle (approved):

20BBrown14:
> We do notes on some components. We used to do it a lot more. I am not sure where the stance is but it is nice to have for things that need a little more explanation.
> https://github.com/SelectQuoteLabs/SQForm/tree/master/notes
> Here is an example
> https://5f4431386ea00a00220d495c-qfuecrbuoq.chromatic.com/?path=/docs/components-sqformdatepicker--date-picker-before-2024

Ah, that's fair. I added a notes page that explains some stuff. Validation would probably be the trickiest thing. I also added component stories, which I left out originally.

20BBrown14:
Thanks for all the feedback and approvals. Merging since everything seems to be addressed 🎉 🌮

Issue #308, opened 2021-06-17
SQFormMultiValue should accept "custom" options as initialValues


20BBrown14:
Affected Components: SQFormMultiValue

I talked with Chris Nemeth and discussed some of the behavior of SQFormMultiValue. We landed on two major changes from current functionality.

1. When a "custom" option is entered by a user we _should not_ add that to the list of dropdown options
2. We should allow "custom" options to be entered as initialValues to this field. The usecase for this is one where, as a dev, I'm getting a list of options an API and I have a list of items previously selected by the user. When I show the user this field I want to prepopulate it with the previously selected options, including the custom ones. Since the custom options are not in the provided options array the component will throw a run time error. We want to resolve that in this issue.

PR #309, opened 2021-06-17
fix: 🐛 SQFormMultiValue accepts custom initial values

20BBrown14:
SQFormMultiValue now accepts custom initial values as well as avoids
adding custom options to list of options displayed in the autocomplete
dropdown

✅ Closes: #308

Loomeroni: https://www.loom.com/share/9a73405ca68f4d00831abedf2e5e5173

### DIFF
--- a/src/components/SQForm/SQFormMultiValue.js
+++ b/src/components/SQForm/SQFormMultiValue.js
@@ -128,16 +128,18 @@ function SQFormMultiValue({
       return acc;
     }, []);
 
-    const nonOptionValues = initialValue.filter(value => {
-      return !displayValues.includes(value);
-    });
+    const newCustomOptions = initialValue.reduce((acc, value) => {
+      if (displayValues.includes(value)) {
+        return acc;
+      }
 
-    const newCustomOptions = nonOptionValues.map(newCustomOptionValue => {
-      return {
-        label: newCustomOptionValue,
-        value: newCustomOptionValue
-      };
-    });
+      acc.push({
+        label: value,
+        value
+      });
+
+      return acc;
+    }, []);
 
     setCustomOptions(previousCustomOptions => {
       return [...previousCustomOptions, ...newCustomOptions];
--- a/src/components/SQForm/SQFormMultiValue.js
+++ b/src/components/SQForm/SQFormMultiValue.js
@@ -123,10 +123,9 @@ function SQFormMultiValue({
       return;
     }
 
-    const displayValues = displayOptions.reduce((acc, option) => {
-      acc.push(option.value);
-      return acc;
-    }, []);
+    const displayValues = displayOptions.map(option => {
+      return option.value;
+    });
 
     const newCustomOptions = initialValue.reduce((acc, value) => {
       if (displayValues.includes(value)) {
--- a/src/components/SQForm/SQFormMultiValue.js
+++ b/src/components/SQForm/SQFormMultiValue.js
@@ -5,6 +5,7 @@ import Autocomplete from '@material-ui/lab/Autocomplete';
 import {VariableSizeList} from 'react-window';
 import {useField, useFormikContext} from 'formik';
 import {useForm} from './useForm';
+import {usePrevious} from '@selectquotelabs/sqhooks';
 
 // MUI uses px, a numeric value is needed for calculations
 const LISTBOX_PADDING = 8; // px
@@ -97,7 +98,7 @@ function SQFormMultiValue({
 }) {
   const classes = useStyles();
   const clearButtonClasses = clearButtonStyles();
-  const [{value: fieldValue}] = useField(name);
+  const [{value: fieldValue}, {initialValue}] = useField(name);
   const {setFieldValue, setTouched, touched} = useFormikContext();
   const {
     fieldState: {isFieldError},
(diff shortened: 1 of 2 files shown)


src/components/SQForm/SQFormMultiValue.js
```diff
@@ -109,8 +110,39 @@ function SQFormMultiValue({
...
+    const nonOptionValues = initialValue.filter(value => {
+      return !displayValues.includes(value);
+    });
+
+    const newCustomOptions = nonOptionValues.map(newCustomOptionValue => {
+      return {
+        label: newCustomOptionValue,
+        value: newCustomOptionValue
+      };
+    });
```
hvilander, line 140: I am finding some of the naming a bit awkward. 
Both 131 and 135 are new custom values, i.e. they are values that are not in the list of predefined options, correct? 

Also you could maybe combine the filter and map into another reduce. Might clean it up a bit and reduce the number of loops. The use case you are building this for I don't think will have a huge set, but other consumers might be glad to cut a loop out.

SeanGroff: > I am finding some of the naming a bit awkward.
> Both 131 and 135 are new custom values, i.e. they are values that are not in the list of predefined options, correct?
> 
> Also you could maybe combine the filter and map into another reduce. Might clean it up a bit and reduce the number of loops. The use case you are building this for I don't think will have a huge set, but other consumers might be glad to cut a loop out.

I WOULD recommend taking the time to reduce (unintentional pun) the number of iterations. Whenever we step into the realm of an autocomplete, we offer an autocomplete because there's a large number of options. So many, it's much faster to type than to scroll.

We have windowing implemented which is the biggest perf gain, but if we have a dropdown of 10000 options, do we want to perform 20000 iterations (map + filter), or 10000 iterations (reduce)?

20BBrown14: That's definitely a good call out and an easy fix and implemented.

SeanGroff (approved):

src/components/SQForm/SQFormMultiValue.js
```diff
@@ -109,8 +110,41 @@ function SQFormMultiValue({
...
+    // Don't run this effect every time displayOptions changes
+    // This also prevents infinite rerenders
+    if (JSON.stringify(previousInitialValue) === JSON.stringify(initialValue)) {
+      return;
+    }
+
+    const displayValues = displayOptions.reduce((acc, option) => {
+      acc.push(option.value);
+      return acc;
+    }, []);
```
hvilander, line 129: [nit] Maybe I am crazy, doesn't a map work better here or at least more specific? We start with one array and want to end up with another array of the same size.

20BBrown14: lol ya, you're right. that 5pm brain really doing my dirty on this PR. I remember my thinking was "I want to reduce this object to a single value, .reduce() will work!"

src/components/SQForm/SQFormMultiValue.js
```diff
@@ -109,8 +110,41 @@ function SQFormMultiValue({
...
+
+    const newCustomOptions = initialValue.reduce((acc, value) => {
+      if (displayValues.includes(value)) {
```
hvilander, line 132: [question] It looks like we loop over displayOptions in line 126 to create an array, displayValues which we loop over on 132 to see if the value is in the array. It would cut another loop to just check here: 
``` 
if (displayOptions.some(({value: displayValue}) => displayValue === value) 
```

the names are awkward there since values is overloaded. Not sure if I am making any sense?

hvilander: Since all of these, correct me if I am wrong here, are the values that the user has added it probably is 10s not 10,000. So it probably doesn't matter.

hvilander (approved):

src/components/SQForm/SQFormMultiValue.js
20BBrown14: I thought about that but wouldn't that mean that for every iteration of `reduce` we're be running the `.some` loop resulting in _many_ more iterations per execution?

hvilander: I think some is as bad as includes? Maybe I am wrong.

20BBrown14: I did a simple test on my machine in chrome and found these results with 100001 items and looking for the last item in the array with .some and .includes. I think the biggest thing with `.some()` is that it uses a callback with generates some overhead.
Screen Shot 2021-06-18 at 10 52 20 AM

hvilander: My original thought was you could cut the map out. If you used `.some()`. If it is trivial, try your test again first mapping an array of objects in to an array of values. Perhaps you already considered that and these number reflect that?

20BBrown14: I hadn't done that before. I made those changes and ran the tests. I accidently created an infinite loop though so my chrome tab crashed and couldn't screenshot the results. The includes test _was_ slightly slower, about 0.3 - 0.5ms on average.

codyarose: I haven't read through all of this, but FWIW I've looked into the `includes` vs `some` thing before, and I believe `includes` is significantly faster than `some`. `some` is good when you need to check the value against some other arbitrary value because of the callback. Then if you get into "big data" territory a `for` loop is likely faster than `includes`

Issue #88, opened 2021-02-04
SQFormDialog content body margin


SeanGroff:
The SQDialog Modal needs specific interior margins set by default.
Interior Content Margins of the scrollable content area needs to be `20PX` on the top, left, bottom, and right sides of the content area.

UPDATE: As of 6/21/2021, Nemeth has clarified we should change the existing `padding: 16px 24px` of the `DialogContent` component to be `padding: 20px` (20 pixels of padding for the Top, Left, Bottom, and Right)

This should be a BREAKING change as consumers who previously added a `div` to control the padding will need to fix their existing implementation.

Comp: https://www.sketch.com/s/b1684560-808a-4f20-960d-fd3969080240/a/GmoOLYY#Inspector

20BBrown14:
Looking at this it seems that the content already has margins for the "scrollable" area. Or perhaps I'm missing something? Screenshots below from storybook. Although not of the same size asked for here.

<img width="1078" alt="Screen Shot 2021-02-26 at 11 02 19 AM" src="https://user-images.githubusercontent.com/29528409/109331220-33e81100-7822-11eb-8d07-2c1e84c78139.png">
<img width="1065" alt="Screen Shot 2021-02-26 at 11 02 13 AM" src="https://user-images.githubusercontent.com/29528409/109331223-35b1d480-7822-11eb-9061-bc8056b252cd.png">
<img width="1064" alt="Screen Shot 2021-02-26 at 11 01 33 AM" src="https://user-images.githubusercontent.com/29528409/109331224-35b1d480-7822-11eb-9f7d-739aa35b69a6.png">
<img width="1075" alt="Screen Shot 2021-02-26 at 11 01 27 AM" src="https://user-images.githubusercontent.com/29528409/109331225-35b1d480-7822-11eb-9fa8-48e30fe21af5.png">

Chris-Boyle:
The current padding is 8px, 24px
![image](https://user-images.githubusercontent.com/20329437/122820645-733edf80-d2a1-11eb-8004-56f532a05159.png)
I will set a padding style on the DialogContent to have the padding be 20px on all sides
          <DialogContent style={{overflowY: 'visible', padding: '20px'}}>

PR #310, opened 2021-06-21
feat: 🎸 adding 20px padding to DialogContent used by the SQFormDialog component

Chris-Boyle:
Adding 20px padding to DialogContent used by the SQFormDialog component.

BREAKING CHANGE: 🧨 Changed SQFormDialog content padding to 20px from 8px 24px

✅ Closes: #88

Loom: https://www.loom.com/share/6ead5c0298f24e4f80533f2d3e05d1af

### DIFF
--- a/src/components/SQFormDialog/SQFormDialogInner.js
+++ b/src/components/SQFormDialog/SQFormDialogInner.js
@@ -44,6 +44,12 @@ const useActionsStyles = makeStyles({
     borderTop: ({palette}) => `1px solid ${palette.divider}`
   }
 });
+const useDialogContentStyles = makeStyles({
+  root: {
+    overflowY: 'visible',
+    padding: '20px'
+  }
+});
 
 function SQFormDialogInner({
   cancelButtonText,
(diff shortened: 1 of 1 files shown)


20BBrown14:
The breaking change section should include what the breaking change actually was, it'll make its way into the changelog

Chris-Boyle:
> The breaking change section should include what the breaking change actually was, it'll make its way into the changelog

This commit tool is so silly sometimes.  I wish it was more intuitive and user friendly.

20BBrown14:
> > The breaking change section should include what the breaking change actually was, it'll make its way into the changelog
> 
> This commit tool is so silly sometimes. I wish it was more intuitive and user friendly.

I agree. Fortunately, however, I think you should be able to do a `git reset` and then run `npm run commit` again and then do a `git push --force` to fix it. Or you could just do `git commit --amend` and change the text. I'm 99% sure (Maybe Sean knows more) it'll pull the text off the commit, whatever it is, as merge time to generate the changelog. Of course you'll need to `git push --force` after a commit amend.

codyarose (approved):

20BBrown14 (approved):
💥

Issue #311, opened 2021-06-21
Convert SQFormReadOnlyField to Typescript


20BBrown14:
Affected Components: SQFormReadOnlyField
Ref: #244 

Convert SQFormReadOnlyField to typescript 🎉

20BBrown14:
Closed with #312

PR #312, opened 2021-06-21
refactor: 💡 Convert SQFormReadyOnlyField to Typescript

20BBrown14:
Refactored SQFormReadOnlyField to be typescript. Since SQFormReadOnly also uses `useForm` I converted that to typescript as well.

I created a common type, `CommonComponentProps` that we should be able to use in all our components. It defines the `name`, `label`, and `size` props since all our components have those props.

✅ Closes: #311


caleboleary (approved):
If you merge the source branch up into this, I think your "Unchanged files with check annotations" warnings should disappear - I changed some linting settings in another branch that has since been merged.

20BBrown14:
> If you merge the source branch up into this, I think your "Unchanged files with check annotations" warnings should disappear - I changed some linting settings in another branch that has since been merged.

That worked. Thanks!

caleboleary:
> > If you merge the source branch up into this, I think your "Unchanged files with check annotations" warnings should disappear - I changed some linting settings in another branch that has since been merged.
> 
> That worked. Thanks!

Ah well, actually now there's a couple others. Looks like they're mad at your `any`s - thoughts on disabling the `no-explicit-anys` rule?

src/types/CommonComponentProps.ts
```diff
@@ -0,0 +1,9 @@
+import {GridSize} from '@material-ui/core';
+
+interface CommonComponentProps {
```
lucas-homer, line 3: [nit] could we make this slightly more specifically named? something like 'CommonFieldComponentProps'? So we can know that it's a base type for the various form field components as opposed SQForm, SQFormDialog, or anything else. Maybe that's self-evident? Thinking out loud here...

SeanGroff:
> > > If you merge the source branch up into this, I think your "Unchanged files with check annotations" warnings should disappear - I changed some linting settings in another branch that has since been merged.
> > 
> > 
> > That worked. Thanks!
> 
> Ah well, actually now there's a couple others. Looks like they're mad at your `any`s - thoughts on disabling the `no-explicit-anys` rule?

If we were refactoring production code incrementally this would be an option.
Since we are waiting until we are 100% TypeScript, we don't want to disable this rule.

SeanGroff:
@20BBrown14 after pulling the code from the source branch, can you run `npm install`?
These linter errors should occur when you attempt to commit your code, never making it to PR.
Curious if that's working or not

src/components/SQForm/useForm.tsx
```diff
@@ -1,11 +1,45 @@
...
+  FieldHelperProps,
+  FieldInputProps,
+  FieldMetaProps,
+  getIn,
+  useField
+} from 'formik';
 import WarningIcon from '@material-ui/icons/NewReleases';
 import VerifiedIcon from '@material-ui/icons/VerifiedUser';
 
+interface useFormParam {
```
codyarose, line 12: The generally accepted format for type and interface names is PascalCase, it makes it easier to distinguish them from regular variables or functions

src/components/SQForm/useForm.tsx
```diff
@@ -1,11 +1,45 @@
...
+  name: string;
+  isRequired: boolean;
+  onBlur?: React.FocusEventHandler;
+  onChange?: React.ChangeEventHandler;
+}
+
+interface useFormReturn {
+  formikField: {
+    field: FieldInputProps<any>;
```
codyarose, line 21: If we don't know what that type argument will be, `unknown` would be a better option - we lose basically all type safety with `any`.

I have a feeling that we could/should use a generic for the value rather than unknown or any, but I'd have to look more into that later

src/types/CommonComponentProps.ts
codyarose: Maybe BaseFieldProps?

codyarose (changes requested):

20BBrown14:
> @20BBrown14 after pulling the code from the source branch, can you run `npm install`?
> These linter errors should occur when you attempt to commit your code, never making it to PR.
> Curious if that's working or not

I can confirm that I have the latest base branch and the linter on `npm run commit` is letting me use all the `any`s I want.

SeanGroff:
> > @20BBrown14 after pulling the code from the source branch, can you run `npm install`?
> > These linter errors should occur when you attempt to commit your code, never making it to PR.
> > Curious if that's working or not
> 
> I can confirm that I have the latest base branch and the linter on `npm run commit` is letting me use all the `any`s I want.

Thanks for for checking on this.
We'll need to open a separate issue to fix the issue with our pre-commit config for TS

20BBrown14:
> > > @20BBrown14 after pulling the code from the source branch, can you run `npm install`?
> > > These linter errors should occur when you attempt to commit your code, never making it to PR.
> > > Curious if that's working or not
> > 
> > 
> > I can confirm that I have the latest base branch and the linter on `npm run commit` is letting me use all the `any`s I want.
> 
> Thanks for for checking on this.
> We'll need to open a separate issue to fix the issue with our pre-commit config for TS

https://github.com/SelectQuoteLabs/SQForm/issues/317

src/components/SQForm/useForm.tsx
```diff
@@ -14,15 +48,15 @@ function _handleError(name, isRequired) {
...
+function _getIsFulfilled(hasValue: boolean, isError: boolean) {
   if (hasValue && !isError) {
     return true;
   }
 
   return false;
 }
 
-function _getHasValue(meta) {
+function _getHasValue(meta: any) {
```
caleboleary, line 59: @20BBrown14 looks like its throwing another any complaint here

20BBrown14: Does anybody know when these start showing up? I pushed my commit and waited for the build to finish, and a few more minutes and saw no issues here on github. But now they're there so just curious what triggers them.

SeanGroff: This comes from the `ci` GH Action I wrote.
GH has a neat feature where it will annotate the Files Changed tab with these warnings.
How long it takes for their UI to shows them? I have no idea.

codyarose:
> @20BBrown14 after pulling the code from the source branch, can you run `npm install`?
> These linter errors should occur when you attempt to commit your code, never making it to PR.
> Curious if that's working or not

I'm wondering if it's not failing on commit because the `no-explicit-any` rule is a warning by default rather than an `error`

codyarose (approved):

Issue #318, opened 2021-06-25
Update id on components to be name instead of label


Chris-Boyle:
I have two fields that both contain a county value.  One is in the mailing address section and one in the personal information  section. This is causing this warning for having duplicate id values.  In Our ReadOnlyComponent it looks like we are setting the id value to just be the label.  We need to do is make that id more unique.

Name appears to be a better option since the names are pretty unique to their use case.

![image](https://user-images.githubusercontent.com/20329437/123448320-69b6bf80-d5a0-11eb-85a7-3ed9cb141ec9.png)

PR #319, opened 2021-06-25
fix: 🐛 Updated id values to name instead of label for more uniq

Chris-Boyle:
Updated id values to name instead of label for more uniqueness. 

✅ Closes: #318


20BBrown14:
Did you check that the tests weren't relying on those labels? That is, are all the tests still passing? I don't think we have tests in our ci pipeline yet.

Chris-Boyle:
> Did you check that the tests weren't relying on those ids? That is, are all the tests still passing? I don't think we have tests in our ci pipeline yet.

The tests that I have seen so far are depending on the name values.  I can let the UI automation team know these changes are coming just in case though.

20BBrown14 (approved):

SeanGroff (approved):

Issue #314, opened 2021-06-23
Convert SQFormCheckbox to Typescript


caleboleary:
ref #244

codyarose:
Solved by #325

PR #325, opened 2021-06-28
refactor: 💡 Convert SQFormCheckbox to typescript Ref #244

caleboleary:
✅ Closes: #314

Loom of it functioning in storybook https://www.loom.com/share/6bcbc4de7364403eb49ec08f6d3d084a

### DIFF
--- a/src/components/SQForm/SQFormCheckbox.tsx
+++ b/src/components/SQForm/SQFormCheckbox.tsx
@@ -2,20 +2,15 @@ import React from 'react';
 import {Checkbox, CheckboxProps} from '@material-ui/core';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Grid from '@material-ui/core/Grid';
+import BaseFieldProps from '../../types/BaseFieldProps';
 
 import {useForm} from './useForm';
 
-interface SQFormCheckboxProps {
+interface SQFormCheckboxProps extends BaseFieldProps {
   /** Disabled state of the checkbox */
   isDisabled?: boolean;
-  /** Descriptive label text for the checkbox */
-  label: string;
-  /** Unique name of the checkbox element */
-  name: string;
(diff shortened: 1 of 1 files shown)


src/components/SQForm/SQFormCheckbox.tsx
```diff
@@ -28,7 +42,7 @@ function SQFormCheckbox({
       <FormControlLabel
         control={
           <Checkbox
-            checked={field.value}
+            checked={field.value as boolean}
```
codyarose, line 45: Is there a way to know what type the value might be, rather than casting as `boolean`? We lose type safety with type assertions
It looks like the value is an array in SQFormCheckboxGroup, but I'm not sure how it's used elsewhere

src/components/SQForm/SQFormCheckbox.tsx
```diff
@@ -1,19 +1,33 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-import Checkbox from '@material-ui/core/Checkbox';
+import {Checkbox, CheckboxProps} from '@material-ui/core';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Grid from '@material-ui/core/Grid';
 
 import {useForm} from './useForm';
 
+interface SQFormCheckboxProps {
```
20BBrown14, line 8: You can use `BaseFieldProps` not that #311 is merged.

src/components/SQForm/SQFormCheckbox.tsx
caleboleary: Cody and I talked offline - Changed to force it to boolean with `!!` - and did some looking at SQFormCheckboxGroup - that array value is never directly passed down to SQFormCheckbox

20BBrown14 (approved):

codyarose (approved):

PR #330, opened 2021-07-07
docs: ✏️ upgrade storybook to 6.3

### LINKED ISSUE #329: Upgrade Storybook to 6.3
https://storybook.js.org/blog/storybook-6-3/

Main new features are some UI improvements, some new addons like [Measure](https://storybook.js.org/addons/@storybook/addon-measure) and [Outline](https://storybook.js.org/addons/storybook-addon-outline) that help with UI debugging (these would need to be installed separately if we want them), and, most importantly, [support for Vite](https://storybook.js.org/blog/storybook-for-vite/)

codyarose:
Closes #329


stories/__tests__/SQForm.stories.test.js
```diff
@@ -151,7 +152,7 @@ describe('Tests for FormWithValidation', () => {
     //Ten Thousand Options
     userEvent.click(screen.getByLabelText(/ten thousand options/i));
     const firstOption = screen.getAllByRole('option')[0];
-    userEvent.click(screen.getAllByRole('option')[0]);
+    fireEvent.click(screen.getAllByRole('option')[0]);
```
SeanGroff, line 155: Question - I thought `userEvent` was preferred over `fireEvent`?

SeanGroff (approved):

stories/__tests__/SQForm.stories.test.js
codyarose: It is, but it seems there was an update to `userEvent` that allows it to error if the target it's clicking isn't clickable because of things like `display: none` or `pointer-events: none`. Some of our tests ensure that nothing happens when an element is clicked and this error is preventing that. `fireEvent` is the low-level api that `userEvent` is built on, so we can still achieve the results we expect without being blocked by the fancy `userEvent` feature/error

codyarose: I feel like there should be an option to allow `userEvent` to click on an element that's somehow disabled without triggering the error, but I wasn't able to find one

Issue #337, opened 2021-07-14
`useForm` needs generics


codyarose:
The current types for `useForm` are not flexible enough for the way some components utilize it, and this is becoming a blocking issue for getting those components migrated to TS. We likely need type arguments for the field's value and event handlers at minimum.

20BBrown14:
Alright. It's been a hot minute since we've made any progress on this. But I _think_ I might have landed on something that is useable. The magic around generics + react events is still a little unknown to me but TS doesn't seem to complain and the types seem to line up with what I think they should be.

I have this branch linked below with the changes. I cribbed some of the changes from Cody's draft PR for this same issue.

After make the changes to the `useForm` hook I converted SQFormMultiSelect and SQFormDropdown to TS to display the changes working as expect. In the same vein, I updated some of the other components that benefit from using the generics.

Hopefully this will get some traction again and we can move forward with complete the Typescript conversion. I think this was the main blocker.

https://github.com/SelectQuoteLabs/SQForm/tree/issue/337_useForm_generics_revisited

SeanGroff:
Fixed by @20BBrown14 in #493

PR #338, opened 2021-07-14
refactor: 💡 add generics to useForm

codyarose:
This is what I've come up with so far - it seems to be working okay for my tests with SQFormTextField, but not quite right for SQFormDropdown (cc @SeanGroff )

https://www.loom.com/share/579531579466421885d6a3583d68265a

✅ Closes: #337


SeanGroff:
Thanks you @codyarose for taking on this challenge 🚀 ❤️

codyarose:
> Thanks you @codyarose for taking on this challenge 🚀 ❤️ 

It's a challenge for sure. The dropdown change handler is really throwing me off

SeanGroff:
> > Thanks you @codyarose for taking on this challenge 🚀 ❤️
> 
> It's a challenge for sure. The dropdown change handler is really throwing me off

When I was reading about this issue, MUI mentioned the `onChange` for the `<Select>` component is completely different than the `onChange` of the rest of the MUI components.

Should we expose a `handleSelectChange` hook in addition to `onChange` in the `useForm` hook? (If we can't solve this challenge with generics)

codyarose:
> Should we expose a `handleSelectChange` hook in addition to `onChange` in the `useForm` hook? (If we can't solve this challenge with generics)

I've been trying to implement that for a bit now and thought I had it figured out, but still errors...
image
I've used the exact props it expects (`SelectProps['onChange']`) and it still somehow gets 
```ts
((event: ChangeEvent<{ name?: string | undefined; value: unknown; }>, child: ReactNode) => void) | ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement> | undefined
```

I'm not sure how the `ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement>` is getting in there

codyarose:
@SeanGroff The main issue appears to be that Formik expects the `onChange` to be in one shape and the Mui components expect it to be in another shape. `Select` passes along a `child` prop with the `event`, `Checkbox` passes along a `checked` prop with the `event`, etc... while Formik just expects an `event`

I think something like this will work:
```ts
interface UseFormParam<ChangeEventType> {
  name: string;
  isRequired: boolean;
  onBlur?: React.FocusEventHandler;
  onChange?: (
    event: React.ChangeEvent<ChangeEventType>,
    ...args: any[]
  ) => void;
}
```
There's a rule to allow `any[]` on rest parameters, which I think is a valid option because sometimes you can't know exactly what all the parameters will be

codyarose:
https://www.loom.com/share/294836a04eb247c4b43168261b920c16

Issue #335, opened 2021-07-13
Convert SQFormCheckboxGroupItem to Typescript 


abbas-ali123:
Affected Components: SQFormCheckboxGroupItem 
Ref: #244

Convert SQFormCheckboxGroupItem to typescript 🎉

20BBrown14:
Closed with #340

PR #340, opened 2021-07-14
refactor: 💡 Convert SQFormCheckboxGroupItem to Typescript

abbas-ali123:
Loom of functionality in storybook
https://www.loom.com/share/6196c918fcba4802b48729cf341d813c

✅ Closes: #335

### DIFF
--- a/src/components/SQForm/SQFormCheckboxGroupItem.tsx
+++ b/src/components/SQForm/SQFormCheckboxGroupItem.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-import FormControlLabel from '@material-ui/core/FormControlLabel';
+import {FormControlLabel, CheckboxProps} from '@material-ui/core';
 import Checkbox from '@material-ui/core/Checkbox';
 import {makeStyles} from '@material-ui/core/styles';
 import {useForm} from './useForm';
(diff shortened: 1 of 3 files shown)


SeanGroff (approved):

caleboleary (approved):

src/components/SQForm/SQFormCheckboxGroupItem.tsx
```diff
@@ -14,6 +13,23 @@ const useStyles = makeStyles(theme => ({
...
 
+interface SQFormCheckboxGroupItemProps {
+  /** The name of the group this checkbox is a part of */
+  groupName: string;
+  /** Label for the checkbox */
+  label: string;
+  /** Value for the checkbox */
+  value: string | boolean | number;
+  /** Function to call when input value is changed */
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
```
codyarose, line 24: We can use `CheckboxProps` here as well, to get the `onChange` type that `<Checkbox />` expects: 
```ts
onChange: CheckboxProps['onChange'];
````

codyarose (changes requested):

codyarose:
When committing changes via `npm run commit` you'll want to just press `enter` to move passed the `BREAKING CHANGES` question if there are no breaking changes. Anything other than an empty value will trigger a major version update by semver.
You may be able to do something like [this](https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message#amending-older-or-multiple-commit-messages) to remove that breaking change section of the commit message.

src/components/SQForm/SQFormCheckboxGroupItem.tsx
abbas-ali123: @codyarose The useForm component does not support that. I have tried and it gives me this error 
image

codyarose: Ah, right. My mistake, that should be possible after I add the useForm generics.

codyarose (approved):

Issue #349, opened 2021-07-14
convert `useFormButton.js` to TS


jamesgriffith9519:
#244 
component Affected: useFormButton.js
convert component to TS

codyarose:
#350

PR #350, opened 2021-07-14
refactor:  useFormButton to TS

jamesgriffith9519:
#244 
#349 
https://www.loom.com/share/610c2bcfeb1047da97435fd85f8b4e8f

### DIFF
--- a/src/components/SQForm/SQFormIconButton.tsx
+++ b/src/components/SQForm/SQFormIconButton.tsx
@@ -14,7 +14,7 @@ interface SQFormIconButtonProps {
   /** Type of button, defaults to 'submit' */
   type?: string;
   /** Standard React event handler */
-  onClick?: (...event: React.MouseEvent<HTMLButtonElement>[]) => void;
+  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
 }
 
 function SQFormIconButton({
--- a/src/components/SQForm/SQFormIconButton.tsx
+++ b/src/components/SQForm/SQFormIconButton.tsx
@@ -14,7 +14,7 @@ interface SQFormIconButtonProps {
   /** Type of button, defaults to 'submit' */
   type?: string;
   /** Standard React event handler */
-  onClick?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  onClick?: (...event: React.MouseEvent<HTMLButtonElement>[]) => void;
 }
 
 function SQFormIconButton({
--- a/src/components/SQForm/useFormButton.tsx
+++ b/src/components/SQForm/useFormButton.tsx
@@ -16,7 +16,7 @@ type UseFormButtonReturnType = {
 export function useFormButton(
   isDisabled: boolean,
   shouldRequireFieldUpdates: boolean,
-  onClick?: (event: React.ChangeEvent<HTMLInputElement>) => void
+  onClick?: (...event: React.MouseEvent<HTMLButtonElement>[]) => void
 ): UseFormButtonReturnType {
   const {values, initialValues, isValid, ...rest} = useFormikContext();
   const hasFormBeenUpdated = hasUpdated(initialValues, values);
--- a/src/components/SQForm/useFormButton.tsx
+++ b/src/components/SQForm/useFormButton.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
-import {useFormikContext} from 'formik';
-import {useDebouncedCallback} from 'use-debounce';
-import {hasUpdated} from '../../utils';
 import {DebouncedState} from 'use-debounce/lib/useDebouncedCallback';
+import {useDebouncedCallback} from 'use-debounce';
+import {useFormikContext} from 'formik';
+import {hasUpdated} from '../../utils';
 
 type UseFormButtonReturnType = {
   isButtonDisabled: boolean;
--- a/src/components/SQForm/useFormButton.tsx
+++ b/src/components/SQForm/useFormButton.tsx
@@ -3,7 +3,11 @@ import {useFormikContext} from 'formik';
 import {useDebouncedCallback} from 'use-debounce';
 import {hasUpdated} from '../../utils';
 
-export function useFormButton(isDisabled, shouldRequireFieldUpdates, onClick) {
+export function useFormButton(
+  isDisabled: boolean,
+  shouldRequireFieldUpdates: boolean,
+  onClick: <TArgs extends any[], TResult>(...args: TArgs) => TResult
+) {
   const {values, initialValues, isValid, ...rest} = useFormikContext();
   const hasFormBeenUpdated = hasUpdated(initialValues, values);
 
--- a/src/components/SQForm/useFormButton.tsx
+++ b/src/components/SQForm/useFormButton.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import {useFormikContext} from 'formik';
 import {useDebouncedCallback} from 'use-debounce';
 import {hasUpdated} from '../../utils';
+import {DebouncedState} from 'use-debounce/lib/useDebouncedCallback';
 
 type UseFormButtonReturnType = {
   isButtonDisabled: boolean;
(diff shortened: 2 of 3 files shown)


jamesgriffith9519:
This had originally tested good and seemed to be working, however, during the build tests, it's uncovered some issues with a couple of types, specifically the Onclick. I'll be working more to address this tomorrow and ill provide an updated commit!

jamesgriffith9519:
I went ahead and moved this back into Review Requested from Draft after recent commits.

codyarose:
@jamesgriffith9519 When using `npm run commit` on the `BREAKING CHANGES` prompt just press `enter` to move passed it if there are no breaking changes. Anything other than an empty value will trigger semver to introduce a major version bump.
You should be able to amend the commit message to remove the breaking change footer by following [these directions](https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message#amending-older-or-multiple-commit-messages)

src/components/SQForm/useFormButton.tsx
```diff
@@ -3,7 +3,21 @@ import {useFormikContext} from 'formik';
...
+  initialValues: unknown;
+  isValid: boolean;
+  handleClick: (...args: unknown[]) => void;
+  [key: string]: unknown;
+};
+
+export function useFormButton(
+  isDisabled: boolean,
+  shouldRequireFieldUpdates: boolean,
+  onClick?: (event: React.ChangeEvent<HTMLInputElement>) => void
```
codyarose, line 19: I think the MouseEvent type you previously had was correct. The onClick type that Mui expects for the Button component is `React.MouseEventHandler<HTMLButtonElement>`, so the onClick type in SQFormButton and SQFormIconButton need to be changed to match that.

codyarose (changes requested):

src/components/SQForm/useFormButton.tsx
jamesgriffith9519: ahh cool! I was thinking the same thing but since the SQFormIconButton got merged I assumed that was the correct one, ill go ahead and adjust that.

codyarose: Yeah, that was an oversight 🥴

src/components/SQForm/SQFormIconButton.tsx
```diff
@@ -14,7 +14,7 @@ interface SQFormIconButtonProps {
   /** Type of button, defaults to 'submit' */
   type?: string;
   /** Standard React event handler */
-  onClick?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  onClick?: (...event: React.MouseEvent<HTMLButtonElement>[]) => void;
```
codyarose, line 17: Sorry, I should've been more clear in my other comment - we want the MouseEvent type that you had before, but not the event spread and `[]`

```suggestion
  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
```

src/components/SQForm/useFormButton.tsx
```diff
@@ -3,7 +3,21 @@ import {useFormikContext} from 'formik';
...
+  initialValues: unknown;
+  isValid: boolean;
+  handleClick: (...args: unknown[]) => void;
+  [key: string]: unknown;
+};
+
+export function useFormButton(
+  isDisabled: boolean,
+  shouldRequireFieldUpdates: boolean,
+  onClick?: (...event: React.MouseEvent<HTMLButtonElement>[]) => void
```
codyarose, line 19: Same here
```suggestion
  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void
```

src/components/SQForm/SQFormIconButton.tsx
SeanGroff, line 17: To my knowledge, there's only 1 argument (the `event`) on button `onClick` handlers.
I would expect something like this:

```javascript
onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
```

src/components/SQForm/useFormButton.tsx
SeanGroff, line 19: Same here, I'd expect the `onClick` to have a single argument, not multiple

src/components/SQForm/useFormButton.tsx
```diff
@@ -3,7 +3,21 @@ import {useFormikContext} from 'formik';
...
 import {hasUpdated} from '../../utils';
 
-export function useFormButton(isDisabled, shouldRequireFieldUpdates, onClick) {
+type UseFormButtonReturnType = {
+  isButtonDisabled: boolean;
+  hasFormBeenUpdated: boolean;
+  values: unknown;
+  initialValues: unknown;
+  isValid: boolean;
+  handleClick: (...args: unknown[]) => void;
```
codyarose, line 12: Also this type needs to be updated since handleClick has changed. If you hover over the `handleClick` declaration it should show you the correct inferred type. Probably something like this
```suggestion
  handleClick: DebouncedState<(event: React.MouseEventHandler<HTMLButtonElement>)=> void | undefined>;
```

SeanGroff, line 12: Does this need to be `unknown`?
Could this be the return type of `useDebouncedCallback`?

codyarose (changes requested):

src/components/SQForm/useFormButton.tsx
jamesgriffith9519: So actually i've been playing around with this after you mentioned it and here are some of the things i've tried,

First i noticed that useDebouncedCallback has a type of `DebouncedState<(event: any) => undefined>` so if i take an approach similar to what Cody suggested below, like this `DebouncedState<(event: React.MouseEventHandler<HTMLButtonElement>)=> void | undefined>;` Then I actually get an error in the SQFormButton component on the getClickHandler function, something like "expecting 1 arg but got 0 or more" which led me to believe that perhaps the getClickHandler function should also not be spread param but instead a single param.

` const getClickHandler = (...args: unknown[]) => {
    if (type === 'reset') {
      return handleReset;
    } else if (typeof onClick !== 'undefined') {
      return handleClick(...args);
    }
  };`

it would look something like this after i adjust it, `const getClickHandler = (args: MouseEventHandler) => {
    if (type === 'reset') {
      return handleReset;
    } else if (typeof onClick !== 'undefined') {
      return handleClick(args);
    }
  };` Though this seems a little narrow, so i was thinking either sticking with the original implimentation of `handleClick: (...args: unknown[]) => void;` or something like DebouncedState<(event: unknown)=> void | undefined>;
  
  what do you guys think?  @SeanGroff @codyarose

codyarose: How about in `UseFormButtonReturnType`:
```ts
handleClick: DebouncedState<React.MouseEventHandler<HTMLButtonElement>>
```
The `useFormButton` `onClick`:
```ts
onClick?: React.MouseEventHandler<HTMLButtonElement>
```
Then pass a generic into `useDebouncedCallback` so it can infer the `event` type:
```ts
const handleClick = useDebouncedCallback<
  React.MouseEventHandler<HTMLButtonElement>
>(event => onClick && onClick(event), 500, {
  leading: true,
  trailing: false
});
```
---
Over in `SQFormButton` the `onClick` prop is wrong, it should be:
```ts
onClick?: React.MouseEventHandler<HTMLButtonElement>
```
Then `getClickHandler` should probably just have an `event` arg:
```ts
const getClickHandler = (event: React.MouseEvent<HTMLButtonElement>) => {
  if (type === 'reset') {
    return handleReset;
  } else if (typeof onClick !== 'undefined') {
    return handleClick(event);
  }
};
```

codyarose (approved):
Nice work @jamesgriffith9519

src/components/SQForm/useFormButton.tsx
```diff
@@ -2,8 +2,23 @@ import React from 'react';
 import {useFormikContext} from 'formik';
 import {useDebouncedCallback} from 'use-debounce';
 import {hasUpdated} from '../../utils';
+import {DebouncedState} from 'use-debounce/lib/useDebouncedCallback';
```
gdoohen1, line 5: Import order =P

gdoohen1 (approved):

20BBrown14 (approved):

PR #354, opened 2021-07-15
refactor: 💡 Converted SQFormTextarea to TypeScript

shoaibhassan:
ScreenCast of this is working in the storybook.
https://www.screencast.com/t/WwXQJTPY37Zd

### DIFF
--- a/src/components/SQForm/SQFormTextarea.tsx
+++ b/src/components/SQForm/SQFormTextarea.tsx
@@ -3,35 +3,35 @@ import TextField from '@material-ui/core/TextField';
 import Grid from '@material-ui/core/Grid';
 import {useFormikContext, FormikProps} from 'formik';
 import {useForm} from './useForm';
-import {TextareaAutosizeProps, TextFieldProps} from '@material-ui/core'
+import {TextFieldProps} from '@material-ui/core';
 import BaseFieldProps from '../../types/BaseFieldProps';
 import {toKebabCase} from '../../utils';
 
 interface SQFormTextareaProps extends BaseFieldProps {
   /** Placeholder text used inside the input field to provide hints to the user */
-  placeholder: string;
+  placeholder?: string;
   /** Disabled property to disable the input if true */
-  isDisabled: boolean;
+  isDisabled?: boolean;
   /** Required property used to highlight input and label if not fulfilled */
-  isRequired: boolean;
+  isRequired?: boolean;
   /** Custom onBlur event callback */
-  onBlur: React.FocusEventHandler;
+  onBlur?: React.FocusEventHandler<HTMLTextAreaElement>;
   /** Custom onChange event callback */
-  onChange: React.ChangeEventHandler;
+  onChange?: React.ChangeEventHandler<HTMLTextAreaElement>;
   /** Number of rows to display when multiline option is set to true. */
-  rows: number;
+  rows?: number;
   /** Maximum number of rows to display when multiline option is set to true. */
-  rowsMax: number;
+  rowsMax?: number;
   /** Attributes applied to the `textarea` element */
-  inputProps: TextareaAutosizeProps;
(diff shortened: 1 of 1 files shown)


src/components/SQForm/SQFormTextarea.tsx
```diff
@@ -1,12 +1,39 @@
...
+  /** Custom onBlur event callback */
+  onBlur: React.FocusEventHandler;
+  /** Custom onChange event callback */
+  onChange: React.ChangeEventHandler;
+  /** Number of rows to display when multiline option is set to true. */
+  rows: number;
+  /** Maximum number of rows to display when multiline option is set to true. */
+  rowsMax: number;
+  /** Attributes applied to the `textarea` element */
+  inputProps: TextareaAutosizeProps;
```
sohail-panezai, line 26: I don't think the type for inputProps should be TextAreaAutosizeProps because we are not using TextAreaAutosize component inside SQFormTextArea component. I think the type should be `React.InputHTMLAttributes<HTMLInputElement>`  as the component being used is TextField component but would like to see what others say in the matter. @SeanGroff @codyarose

src/components/SQForm/SQFormTextarea.tsx
```diff
@@ -1,12 +1,39 @@
...
 
+interface SQFormTextareaProps extends BaseFieldProps {
+  /** Placeholder text used inside the input field to provide hints to the user */
+  placeholder: string;
+  /** Disabled property to disable the input if true */
+  isDisabled: boolean;
+  /** Required property used to highlight input and label if not fulfilled */
+  isRequired: boolean;
+  /** Custom onBlur event callback */
+  onBlur: React.FocusEventHandler;
```
codyarose, line 18: To be more explicit:
```suggestion
  onBlur: React.FocusEventHandler<HTMLTextAreaElement>;
```

src/components/SQForm/SQFormTextarea.tsx
```diff
@@ -1,12 +1,39 @@
...
+  /** Custom onChange event callback */
+  onChange: React.ChangeEventHandler;
```
codyarose, line 20:
```suggestion
  onChange: React.ChangeEventHandler<HTMLTextAreaElement>;
```

src/components/SQForm/SQFormTextarea.tsx
```diff
@@ -1,12 +1,39 @@
...
 import Grid from '@material-ui/core/Grid';
-import {useFormikContext} from 'formik';
-
+import {useFormikContext, FormikProps} from 'formik';
 import {useForm} from './useForm';
+import {TextareaAutosizeProps, TextFieldProps} from '@material-ui/core'
+import BaseFieldProps from '../../types/BaseFieldProps';
 import {toKebabCase} from '../../utils';
 
+interface SQFormTextareaProps extends BaseFieldProps {
```
codyarose, line 10: Per the original propTypes, `name` and `label` are the only required props, which are included in `BaseFieldProps`, everything else here should be set to optional.

src/components/SQForm/SQFormTextarea.tsx
codyarose: @sohail-panezai is correct; we shouldn't use `TextAreaAutosizeProps` because we're not using that component here, and the type for `inputProps` should be the html textarea element's props/attributes not a React component's props.
`React.InputHTMLAttributes<HTMLTextAreaElement>` should work

codyarose (changes requested):

src/components/SQForm/SQFormTextarea.tsx
shoaibhassan: @codyarose  field those are not required set to Optional

codyarose (approved):

gdoohen1 (approved):

PR #361, opened 2021-07-21
Issue244/loading spinner convert ts

jamesgriffith9519:
Convert LoadingSpinner to TS and create story file
#359 
#244

### DIFF
--- a/src/components/LoadingIcon/LoadingIcon.tsx
+++ b/src/components/LoadingIcon/LoadingIcon.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import {makeStyles} from '@material-ui/core/styles';
 
-const useStyles = makeStyles(theme => ({
+const useStyles = makeStyles({
   '@keyframes bumpit': {
     '0%': {
       transform: 'scale(1)'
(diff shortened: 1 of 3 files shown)


jamesgriffith9519:
Here's a loom video explaining some of the changes i had to work through! https://loom.com/share/69445665604844f78b23a67c930ac088

TLDR, or rather TDLW? 
The loadingIcon component that is being consumed by this component had broken animations, they were not working and the icon was static because the syntax for animations was incorrect for makestyles, so i had to correct that first.

Next there wasn't a story for this so i took a crack at creating one, let me know if i should make adjustments to it! 

lastly, simple conversion from JS to TS for the LoadingSpinner component.

stories/LoadingSpinner.stories.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+};
+
+export const LoadingSpinnerWithoutMessage = () => {
+  return <LoadingSpinner />;
+};
+
+export const LoadingSpinnerWithMessage = args => {
+  return <LoadingSpinner message={args.message ?? 'Loading...'} />;
+};
```
codyarose, line 21: With Storybook Controls, in most cases, I don't think we really need multiple stories for the different options of a component. Just one story with the `message` control should suffice.
Here's how I'd make that change:
```suggestion
export const Default = args => {
  return <LoadingSpinner {...args} />;
};
Default.storyName = 'LoadingSpinner';
```
Since there's only one story we can take advantage of single story hoisting by calling this Default then set the storyName to the name of the component. This will tell Storybook to display the story itself in the sidebar instead of a folder containing the story.

jamesgriffith9519: Ahh super cool, ill make those adjustments!

codyarose (approved):

src/components/LoadingIcon/LoadingIcon.tsx
```diff
@@ -17,16 +17,16 @@ const useStyles = makeStyles({
     }
   },
   orange: {
-    animation: 'bumpit 2s linear 0.66s infinite'
+    animation: `$bumpit 2s linear 0.66s infinite`
```
gdoohen1, line 20: Do these need to be string templates for the `$` keyframe reference to work?

gdoohen1 (approved):

src/components/LoadingIcon/LoadingIcon.tsx
jamesgriffith9519: from what i have read and understand, yes!

There's another syntax that also works, but it's not an exact match to what we are using across our other components so i think using the string templates is the safest approach!

PR #362, opened 2021-07-26
Add SQFormDatePicker tests

### LINKED ISSUE #265: Add tests for SQFormDatePicker
SQFormDatePicker needs tests

laurelbean:
The tests I have here currently are all I could figure out. The testing library is doing really weird things in the interpretation of the DOM for the DatePickers. The textbox doesn't let you type and instead opens the calendar like a button. 

I think we need to pause on any other tests for these components until MUI version 5 comes out and we can try a new version of the date pickers. The testing library might work better with the updated version.

Leaving this as draft for now, since I think this needs some discussion on direction / what we should do.

Closes #265


stories/__tests__/SQFormDatePicker.stories.test.js
```diff
@@ -0,0 +1,142 @@
...
+      date: ''
+    };
+
+    renderDatePicker({SQFormProps: {initialValues}});
+
+    const textField = screen.getByRole('textbox', {name: /date/i});
+    expect(textField).toHaveValue('');
+  });
+
+  it('should render value with non-empty initial value', () => {
```
gdoohen1, line 43: This test seems like a duplicate of the second part of the first test.

stories/__tests__/SQFormDatePicker.stories.test.js
```diff
@@ -0,0 +1,142 @@
...
+    renderDatePicker();
+
+    const textField = screen.getByRole('textbox', {name: /date/i});
+    expect(textField).toBeInTheDocument();
+    expect(textField).toHaveValue('09/22/2020');
+  });
+
+  it('should update when text is entered', () => {
```
gdoohen1, line 51: The content of this test seems like a duplicate of the second test but the name seems different from what's being tested here - are you just using this as a placeholder to fill the test since we can't fully test this until MUI v5?

laurelbean: Ah, good catch. I should probably take this one out for now. I had a bunch of commented out items in here previously, so it doesn't make sense now.

stories/__tests__/SQFormDatePicker.stories.test.js
laurelbean: You're right. It's a duplicate since the story has a default value. I think it might still be worth keeping if I use a different value from the default. Or do you still think it would be duplicate?

gdoohen1: Yeah I think passing in a non-default initial value would a good test to have for sure.

gdoohen1 (approved):

stories/__tests__/SQFormDatePicker.stories.test.js
```diff
@@ -0,0 +1,135 @@
...
+  );
+};
+
+describe('SQFormDatePicker Tests', () => {
+  it('should render a label and input', () => {
+    renderDatePicker();
+
+    const label = screen.getByText('Date');
+    expect(label).toBeInTheDocument();
+    expect(label).toHaveTextContent('Date');
```
codyarose, line 25: I think this `expect` is redundant -- you've used `getByText('Date')` to get the `label`, so it's safe to assume that it also has the text content `'Date'`

codyarose (approved):

PR #367, opened 2021-07-28
refactor: 💡 refactor from js to ts

jamesgriffith9519:
#244 
Convert SQFormResetButtonWithConfirmation to TS
https://www.loom.com/share/e19ecfefef324c0290565e1a77eb3241

### DIFF
--- a/src/components/SQForm/SQFormResetButtonWithConfirmation.tsx
+++ b/src/components/SQForm/SQFormResetButtonWithConfirmation.tsx
@@ -1,8 +1,24 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-import {useDialog} from '@selectquotelabs/sqhooks';
 import {DialogAlert, RoundedButton} from 'scplus-shared-components';
 import {useFormButton} from './useFormButton';
+import {useDialog} from '@selectquotelabs/sqhooks';
+
+type SQFormResetButtonWithConfirmationTypes = {
+  /** The contents of the form button; usually text */
+  children: React.ReactNode;
+  /** Title of the confirmation dialog */
+  confirmationTitle?: string;
+  /** Content of the confirmation dialog */
+  confirmationContent: React.ReactNode;
+  /** Whether the button is disabled */
+  isDisabled?: boolean;
+  /** Title of the button */
+  buttonTitle?: string;
+  /** Callback for reset*/
+  onReset?: () => void;
+
+  variant?: string;
+};
 
 function SQFormResetButtonWithConfirmation({
   children,
@@ -12,13 +28,12 @@ function SQFormResetButtonWithConfirmation({
   confirmationTitle = 'Reset Form',
   variant = 'contained',
   onReset
-}) {
+}: SQFormResetButtonWithConfirmationTypes): JSX.Element {
   const {isDialogOpen, openDialog, closeDialog} = useDialog();
   const {dirty, handleReset} = useFormButton(isDisabled);
 
-  const handlePrimaryButtonClick = () => {
+  const handlePrimaryButtonClick = (): void => {
     handleReset();
-
     // optional function prop consumer can pass to handle any additional side effects when resetting form
     onReset && onReset();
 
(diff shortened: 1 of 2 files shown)


src/components/SQForm/useFormButton.tsx
```diff
@@ -12,11 +12,12 @@ type UseFormButtonReturnType = {
   isValid: boolean;
   handleClick: DebouncedState<React.MouseEventHandler<HTMLButtonElement>>;
   [key: string]: unknown;
+  handleReset: () => void;
 };
 
 export function useFormButton(
   isDisabled: boolean,
-  shouldRequireFieldUpdates: boolean,
+  shouldRequireFieldUpdates?: boolean,
```
jamesgriffith9519, line 20: I noticed this component never had any propTypes to start with that would tell me if these props should be required or optional, i just looked inside of some other components that are consuming this to determine that, some use all 3 props and others. just seem to use IsDiabled, so i went ahead and made this 'shouldRequireFieldUpdates' optional.

src/components/SQForm/SQFormResetButtonWithConfirmation.tsx
```diff
@@ -1,8 +1,24 @@
...
+  confirmationTitle?: string;
+  /** Content of the confirmation dialog */
+  confirmationContent: React.ReactNode;
+  /** Whether the button is disabled */
+  isDisabled?: boolean;
+  /** Title of the button */
+  buttonTitle?: string;
+  /** Callback for reset*/
+  onReset?: () => void;
+
```
mishaqkhan, line 19: [Suggestion] Can we add comment to explain `variant` prop ?

src/components/SQForm/SQFormResetButtonWithConfirmation.tsx
```diff
@@ -1,8 +1,24 @@
...
+  /** Content of the confirmation dialog */
+  confirmationContent: React.ReactNode;
+  /** Whether the button is disabled */
+  isDisabled?: boolean;
+  /** Title of the button */
+  buttonTitle?: string;
+  /** Callback for reset*/
+  onReset?: () => void;
+  /** style variant, "contained" or "outlined"*/
+  variant?: string;
```
codyarose, line 20: `variant` should be limited to "contained" or "outlined" if those are the only options allowed
```suggestion
  variant?: 'contained' | 'outlined';
```

codyarose (changes requested):

src/components/SQForm/SQFormResetButtonWithConfirmation.tsx
jamesgriffith9519: ahh, good catch! a union type would work perfectly!

20BBrown14 (approved):

src/components/SQForm/SQFormResetButtonWithConfirmation.tsx
```diff
@@ -1,8 +1,24 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-import {useDialog} from '@selectquotelabs/sqhooks';
 import {DialogAlert, RoundedButton} from 'scplus-shared-components';
 import {useFormButton} from './useFormButton';
+import {useDialog} from '@selectquotelabs/sqhooks';
+
+type SQFormResetButtonWithConfirmationTypes = {
```
20BBrown14, line 6: I'm not sure it matters too much but I think all the other components so far have used `interface` instead of `type`. Also the names we're using are of hte format `${componentName}Props`

src/components/SQForm/useFormButton.tsx
```diff
@@ -12,11 +12,12 @@ type UseFormButtonReturnType = {
   isValid: boolean;
   handleClick: DebouncedState<React.MouseEventHandler<HTMLButtonElement>>;
   [key: string]: unknown;
+  handleReset: () => void;
```
codyarose, line 15: This `handleReset` should actually come from the `...rest` in `useFormikContext`. I'll be creating an issue and PR for that and a couple other type improvements soon

codyarose: #382

codyarose: Merged #382, so the handleReset should be typed properly and you can remove this addition to useFormButton

codyarose:
This looks good! Could you merge the recent updates from #248? Otherwise the tests are going to fail in CI

codyarose:
@jamesgriffith9519 I've updated #244 with the changes from #385, so this PR needs to pull in those changes and update `SQFormResetButtonWithConfirmation` to use the new `useFormButton` object parameter. After that this should be good to go 🤞

PR #371, opened 2021-07-29
refactor: 💡 SQForm to ts

AamirMaan:
Finally, first component to typescript
https://www.loom.com/share/a29b81ba523f46ccbeea6f9ed580d3ed


src/components/SQForm/SQForm.tsx
```diff
@@ -24,12 +50,15 @@ function SQForm({
...
+  const handleReset = (): void => {
+    document &&
+      document.activeElement &&
+      (document.activeElement as HTMLElement).blur();
   };
 
   const handleSubmit = useDebouncedCallback(
-    (...args) => onSubmit(...args),
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (...args: [any]) => onSubmit(...args),
```
gdoohen1, line 61: I think @jamesgriffith9519 ran into something similar to this the other day and found an alternative to `...args`

src/components/SQForm/SQForm.tsx
```diff
@@ -1,19 +1,45 @@
...
+   *
+   * https://jaredpalmer.com/formik/docs/api/withFormik#handlesubmit-values-values-formikbag-formikbag--void--promiseany
+   * */
+  onSubmit: (event: React.SyntheticEvent) => void;
+  /**
+   * Yup validation schema shape
+   * https://jaredpalmer.com/formik/docs/guides/validation#validationschema
+   * */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  validationSchema?: Record<string, any>;
```
gdoohen1, line 32: Can this be `Record<string, unknown>` instead of using `any` ?

src/components/SQForm/SQForm.tsx
AamirMaan: > I think @jamesgriffith9519 ran into something similar to this the other day and found an alternative to `...args`

I would love to see that, I was stuck with this for a bit, and was nothing seemed to work.

src/components/SQForm/SQForm.tsx
AamirMaan: > Can this be `Record<string, unknown>` instead of using `any` ?

Actually I tried this one, but it was yelling
image

src/components/SQForm/SQForm.tsx
jamesgriffith9519: > > I think @jamesgriffith9519 ran into something similar to this the other day and found an alternative to `...args`
> 
> I would love to see that, I was stuck with this, and nothing seemed to work.

Sorry for the delay! I didin't notice this until i checked my email and found it, here's the file i was working in and I think it will help with this issue that you are having! https://github.com/SelectQuoteLabs/SQForm/pull/350/files#diff-b2fd4208eb74b575ac362adcbdd31266b1ab12b56a0d0b04bd0bb7d61696d7c4

src/components/SQForm/SQForm.tsx
gdoohen1: @AamirMaan It looks like `Yup` has a type for schemas. If you look in `SQForm.stories.js` line 217 and mouse over `validationSchema` you can see what the schemas look like. But in this we don't know exactly what schemas we are getting so instead of `any` you should be able to use `Yup.MixedSchema` instead.

codyarose (changes requested):

src/components/SQForm/SQForm.tsx
```diff
@@ -1,19 +1,50 @@
...
+    values: Values,
+    formikHelpers: FormikHelpers<Values>
+  ) => void | Promise<unknown>;
+  /**
+   * Yup validation schema shape
+   * https://jaredpalmer.com/formik/docs/guides/validation#validationschema
+   * */
+  validationSchema?: Record<
+    keyof Values,
+    Yup.AnySchema<Values[keyof Values] | null | undefined>
```
AamirMaan, line 36: Adding null here, as in most of repos, nullable is being passed to Yup schema

codyarose (approved):

gdoohen1 (approved):

codyarose:
@AamirMaan Please merge the recent updates from #248 and resolve any conflicts

src/components/SQForm/SQForm.tsx
```diff
@@ -1,19 +1,50 @@
...
+  >;
+}
+
+function SQForm<Values>({
```
20BBrown14, line 40: Hm. I've just seen this. Why does SQForm need an argument like this? This is definitely at the limit of my TS knowledge.

As a consumer what would I pass here?

gdoohen1: I think it's for the props `initialValues` and `validationSchema` to be able to use that to define their types.

20BBrown14: cool, I just watched cody's video above.

So when i consume then I don't need to `SQForm<someTypehere>` , TS will infer that type based on what I pass into `initialValues`?

Or do i need a seperate interface, `initialValuesTypes` outlining the types of the my initialValues?

codyarose: It should get the type from whatever you put in `initialValues`. I'm going to take a second look at this because we may not need to have that generic on `SQForm`, which could prevent any future confusion about needing to provide a generic

gdoohen1: Oh I didn't know it inferred the values that's awesome.

AamirMaan: It was new for me, this thing is awesome, I didn't we could pull off something like this using inferring type

codyarose: Alright, I took another look and I do think this is the way to go. With the `Values` generic on SQForm, we infer the type from `initialValues` and pass it to the `SQFormProps` interface so we strongly type the `onSubmit` and `validationSchema` props. You could still provide the `<SQForm>` component with a value for that generic, but `initialValues` is a required prop and linked to that generic, so whatever you define in the generic will have to match the type of whatever is given to `initialValues`. However, there's no added value in doing that -- the consumer won't need to know or care about the generic

[This SO post](https://stackoverflow.com/questions/67869151/infer-generic-type-between-2-props-in-react-typescript) shows the same solution in a more simple context -- that may better illustrate what's going on here if anyone is confused

package.json
```diff
@@ -84,7 +84,8 @@
     "react-window": "^1.8.5",
     "scplus-shared-components": "https://npm-public.selectquotelabs.com/scplus-shared-components/7.3.1",
     "text-mask-addons": "^3.8.0",
-    "use-debounce": "^6.0.0"
+    "use-debounce": "^6.0.0",
+    "yup": "^0.32.9"
```
20BBrown14, line 88: FWIW, there was a breaking change in Yup 0.31 that was "array().required() will no longer consider an empty array missing and required checks will pass." which we rely on for some of our components, name checkboxgroup. This is why your tests and ci are failing. So I think we need to ugprade yup in a seperate issue which will broke this and #395

20BBrown14 (changes requested):

package.json
20BBrown14: #398 has been merged, so after a rebase/merge this branch's CI failures should be resolved.

AamirMaan: I am trying to merge latest changes from issue/244-EnableTypeScript branch, but it is giving me error which is I am unable to understand/resolve. 
image
Is anyone else facing the same issue while commiting or merging the code or is it only me?

20BBrown14: We've had this discussion before in Slack somewhere. To get around this, because it's an error you can't really fix without changing config and it's not something that we expect to see often, just do `git commit -m "whatever your commit message is here" --no-verify`

DO NOT skip verification otherwise with `--no-verify`.

20BBrown14: Looks like a lot of files got touched by prettier or the linter for some reason

AamirMaan: Yes, I have created a new branch and pushed my changes to that branch "issue/346-EnableTypeScript-SQForm". Maybe creating new pull request and declining this is a good idea.

20BBrown14: uh, idk about that. Seems unnecessary, imo. bUt that's just me.

AamirMaan: Yes, you are right, those changes don't seem harmful. So I am sticking with this PR.

aharpt:
Closing this Pull Request since it has not been worked on for ~5 months and it seems it would take less work to simply close this and open another PR as opposed to work off this current one.

PR #372, opened 2021-07-29
fix: 🐛 gave autocomplete dropdown tooltips and larger size

gdoohen1:
Here is the ticket for these changes https://selectquote.atlassian.net/browse/SSC-134

The issue was the autocomplete's dropdown of options was limited to the field's size in the form which caused issues with readability of the options. I increased the width of the dropdown to be 2X the width of the field but with a limit so that it doesn't run off the screen. I also added tooltips to all the options.

**Loom**: https://www.loom.com/share/609c71187ca84dc6925ea88a5652e846
Note: The loom shows the tooltip text in middle of the row, I've updated it to be on the bottom-left so it now shows directly below the value for very large dropdowns.

### DIFF
--- a/src/components/SQForm/SQFormAutocomplete.js
+++ b/src/components/SQForm/SQFormAutocomplete.js
@@ -347,7 +347,7 @@ SQFormAutocomplete.propTypes = {
   onInputChange: PropTypes.func,
   /** Size of the input given full-width is 12. */
   size: PropTypes.oneOf(['auto', 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
-  /** Lock width to the width of the field in the form */
+  /** Lock width of the dropdown to the width of the field in the form */
   lockWidthToField: PropTypes.bool
 };
 
--- a/src/components/SQForm/SQFormAutocomplete.js
+++ b/src/components/SQForm/SQFormAutocomplete.js
@@ -1,11 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import TextField from '@material-ui/core/TextField';
+import {
+  TextField,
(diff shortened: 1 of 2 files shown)


lucas-homer:
So if I understand correctly, this is now the default behavior. What if we had a flag or something to lock the width to the input width? I can imagine some situations where the dropdown options are not long, and they're positioned on the left of a form and we're going to end up with humongous dropdown windows. Thoughts?

gdoohen1:
> So if I understand correctly, this is now the default behavior. What if we had a flag or something to lock the width to the input width? I can imagine some situations where the dropdown options are not long, and they're positioned on the left of a form and we're going to end up with humongous dropdown windows. Thoughts?

Yeah I think this is a good idea. Our stories I think are a bit atypical in that we have very large sizes for some of the fields, the fields in the actual CRM use `size={2}` for pretty much every field regardless of what it's for, so 2X width there is a bit less drastic but I agree this should definitely be able to be opted out of.

src/components/SQForm/SQFormAutocomplete.js
```diff
@@ -109,9 +201,13 @@ function SQFormAutocomplete({
   onBlur,
   onChange,
   onInputChange,
-  size = 'auto'
+  size = 'auto',
+  lockWidthToField
```
SeanGroff, line 205: Can we default `lockWidthToField` to be false so we don't rely on `undefined`? Once this gets to TypeScript we'd have to make this adjustment anyway

SeanGroff (approved):
Great work @gdoohen1 💯

lucas-homer (approved):

codyarose (approved):

PR #375, opened 2021-08-02
fix(SQFormGuidedWorkflow.js): add padding and update shared component…

shahidmanzoorzekiexperts:
When we updated the version of scplus-shared-components in SQForm then it causes a Padding issue with SQFormGuidedWorkflow component.
https://www.loom.com/share/66e8fc9316124bc4aed2c0e3fa27a2c2

Before:
before

After:
lib

### DIFF
--- a/src/components/SQFormGuidedWorkflow/SQFormGuidedWorkflow.js
+++ b/src/components/SQFormGuidedWorkflow/SQFormGuidedWorkflow.js
@@ -18,6 +18,9 @@ import {GuidedWorkflowProps} from './PropTypes';
 
 const useStyles = makeStyles(() => {
   return {
+    root: {
+      padding: '16px !important' /* material-ui override */
+    },
     panelFooter: {
       display: 'flex',
       justifyContent: 'space-between',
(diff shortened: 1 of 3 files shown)


src/components/SQFormGuidedWorkflow/SQFormGuidedWorkflow.js
```diff
@@ -18,6 +18,9 @@ import {GuidedWorkflowProps} from './PropTypes';
 
 const useStyles = makeStyles(() => {
   return {
+    root: {
+      padding: '16px !important' /* material-ui override */
```
SeanGroff, line 22: The `!important` tag was needed in version 0.x of Material-UI but shouldn't be needed when using `makeStyles`
Can you try this without the `!important` tag please?

SeanGroff (changes requested):

src/components/SQFormGuidedWorkflow/SQFormGuidedWorkflow.js
shahidmanzoorzekiexperts: let me try

shahidmanzoorzekiexperts: I removed !important but not working I tried.

SeanGroff (approved):
Thanks @shahidmanzoorzekiexperts for your contribution!

PR #376, opened 2021-08-02
refactor: 💡 Convert SQFormResetInitialValuesButton to typescript Ref #244

mishaqkhan:
Loom of it functioning in storybook:

https://www.loom.com/share/a02b663efdfa4f2fb9c73c524e4a93cc


20BBrown14 (approved):

src/components/SQForm/SQFormResetInitialValuesButton.tsx
```diff
@@ -1,9 +1,28 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import {FormikValues, FormikState} from 'formik';
 import {useDialog} from '@selectquotelabs/sqhooks';
 import {DialogAlert, TextButton} from 'scplus-shared-components';
 import {useFormButton} from './useFormButton';
 
+interface Props {
```
20BBrown14, line 7: I believe we're using ${componentName}Props for these interfacts

src/components/SQForm/SQFormResetInitialValuesButton.tsx
```diff
@@ -13,14 +32,15 @@ function SQFormResetInitialValuesButton({
...
-}) {
+}: Props): JSX.Element {
   const {isDialogOpen, openDialog, closeDialog} = useDialog();
-  const {values, resetForm} = useFormButton(isDisabled);
+  const {values, resetForm} = useFormButton(isDisabled, false);
 
-  const handlePrimaryButtonClick = () => {
-    resetForm({
+  const handlePrimaryButtonClick = (): void => {
+    //TODO: Remove ResetForm and FormikValues once useFormButton is TSified
```
codyarose, line 40: `useFormButton` has already been converted to TS, but looks like its types can be improved

codyarose: #382

codyarose: Merged #382. Pull that change in and see if the types are working

mishaqkhan: Sure thanks

src/components/SQForm/SQFormResetInitialValuesButton.tsx
```diff
@@ -13,14 +30,15 @@ function SQFormResetInitialValuesButton({
   tooltip = 'Reset Button',
   initialValuesObject = {},
   ...props
-}) {
+}: SQFormResetInitialValuesButtonProps): JSX.Element {
   const {isDialogOpen, openDialog, closeDialog} = useDialog();
-  const {values, resetForm} = useFormButton(isDisabled);
+  const {values, resetForm} = useFormButton(isDisabled, false);
```
codyarose, line 35: > Still getting error on values.

`useFormButton` now accepts a `Values` generic that will tell `useFormikContext` what shape the form's values are in. With that, the `values` that is returned from `useFormButton` will be typed to what is passed in as the generic. 
So, we need to pass in the type of the values object:
```suggestion
  const {values, resetForm} = useFormButton<typeof initialValuesObject>(isDisabled, false);
```
Then `values` will no longer need to be cast as `FormikValues`

codyarose (changes requested):

codyarose (approved):

Issue #379, opened 2021-08-02
Migrate `SQFormMaskedTextField.js` to TS


20BBrown14:

20BBrown14:
Closed with #380

PR #380, opened 2021-08-02
refactor: 💡 Convert SQFormMaskedTextField to TS

20BBrown14:
✅ Closes: #379
ref: #244


src/components/SQForm/SQFormMaskedTextField.tsx
```diff
@@ -1,14 +1,52 @@
...
+  endAdornment?: React.ReactElement;
+  /** Defines the input type for the text field. Must be a valid HTML5 input type */
+  type?: string;
+  /** Props applied to the Input element */
+  InputProps?: InputProps;
+  /** Attributes applied to the `input` element */
+  inputProps?: React.InputHTMLAttributes<HTMLInputElement>;
+  /** Any valid prop for material ui text input child component - https://material-ui.com/api/text-field/#props */
+  muiFieldProps?: TextFieldProps;
+}
```
codyarose, line 33: Since the props for this component are identical to SQFormTextField, other than `mask`, we could just export the `SQFormTextFieldProps` interface and use it here.

```suggestion
import { SQFormTextFieldProps } from './SQFormTextField';

interface SQFormMaskedTextFieldProps extends SQFormTextFieldProps {
  /** Valid mask array; custom or from utils/masks.js */
  mask?: maskProp;
}
```

src/components/SQForm/SQFormMaskedTextField.tsx
```diff
@@ -1,14 +1,52 @@
...
+import {TextFieldProps, InputProps} from '@material-ui/core';
 import MaskedInput from 'react-text-mask';
+import BaseFieldProps from 'types/BaseFieldProps';
+import {maskProp} from '../../types/MaskTypes';
 import SQFormTextField from './SQFormTextField';
 
-function TextFieldMask({inputRef, mask, ...rest}) {
+interface SQFormMaskedTextFieldProps extends BaseFieldProps {
+  /** Valid mask array; custom or from utils/masks.js */
+  mask?: maskProp;
```
codyarose, line 10: Shouldn't `mask` be required? I know it wasn't before, but if the consumer doesn't provide a mask they should just use a TextField, right?

codyarose (changes requested):

src/components/SQForm/SQFormMaskedTextField.tsx
20BBrown14: I figure that maybe there'd be a use case for a conditional mask and instead of conditionally using a whole different component they can just not supply the mask? 🤷

codyarose: Yeah, that makes sense

codyarose (approved):

codyarose:
@20BBrown14 It looks like after that merge you have both the js and ts versions of this component

gdoohen1 (approved):

src/components/SQForm/SQFormMaskedTextField.tsx
```diff
@@ -57,7 +57,7 @@ function SQFormMaskedTextField({
       placeholder={placeholder}
       size={size}
       onBlur={onBlur}
-      onChange={stripNonNumeric ? handleChangeStripNonNumeric : onChange}
```
codyarose: Ah the stripNonNumeric feature was lost in the merge as well

PR #383, opened 2021-08-03
Issue/369 migration sq form date picker

shoaibhassan:
Here is the loom for demo after migration to TypeScript.
https://www.loom.com/share/6af15322b0d948c78f4ba3182cc5da17


src/components/SQForm/SQFormDatePicker.tsx
```diff
@@ -16,6 +21,32 @@ const useStyles = makeStyles(() => ({
...
+  onBlur?: React.FocusEventHandler<HTMLInputElement>;
+  /** Custom onChange event callback */
+  onChange?: (
+    date: React.ChangeEvent<Element | moment.Moment> | unknown
+  ) => void;
+  /** Disable specific date(s) (day: DateIOType) => boolean
+   * This is a predicate function called for every day of the month
+   * Return true to disable that day or false to enable that day
+   */
+  setDisabledDate: (day: unknown) => boolean;
```
hvilander, line 41: do we want to avoid unknown?

shoaibhassan: Nope, I tried to pass the `Moment` type to but it highlighted it as a warning. The default type is `unknown` suggested by linter. 
image

src/components/SQForm/SQFormDatePicker.tsx
```diff
@@ -16,6 +21,32 @@ const useStyles = makeStyles(() => ({
...
+  isDisabled?: boolean;
+  /** Required property used to highlight input and label if not fulfilled */
+  isRequired?: boolean;
+  /** Placeholder text used inside the input field to provide hints to the user */
+  placeholder?: string;
+  /** Custom onBlur event callback */
+  onBlur?: React.FocusEventHandler<HTMLInputElement>;
+  /** Custom onChange event callback */
+  onChange?: (
+    date: React.ChangeEvent<Element | moment.Moment> | unknown
```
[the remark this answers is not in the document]
shoaibhassan: Yeah, the date cannot be an element. 

- [ remove Element from date type]

src/components/SQForm/SQFormDatePicker.tsx
```diff
@@ -41,7 +72,9 @@ function SQFormDatePicker({
     onChange
```
hvilander, line 72: useForm calls onChange differently than we are using it in handleChange. Is there some typing work here that needs done? This was one detail that held me up a bit when I was duplicating this work.

src/components/SQForm/SQFormDatePicker.tsx
20BBrown14: Date should be of type `DateIOType` as seen here https://material-ui-pickers.dev/api/DatePicker
DateIOType is, according to the link above, "DateIOType — date object type of your linked date-io adapter (Moment, DayJS, etc.)"

Is there _any_ way we can get more specific than `unknown`?

src/components/SQForm/SQFormDatePicker.tsx
```diff
@@ -29,7 +60,7 @@ function SQFormDatePicker({
   muiFieldProps = {},
   muiTextInputProps = {},
   isCalendarOnly = false
-}) {
+}: SQFormDatePickerProps): JSX.Element {
```
20BBrown14, line 63: I don't think it matters too much but I believe we prefer `React.ReactElement` over `JSX.Element`

shoaibhassan: We have used `JSX.Element` in the SQForm TypeScript version.

src/components/SQForm/SQFormDatePicker.tsx
shoaibhassan: Yes, they have mentioned `DateIOType ` in docs. but material-UI `picker` is using the `shouldDisableDate?: (day: unknown) => boolean;` type under the hood
image

**Refrence file path:**  node_modules\@material-ui\pickers\_helpers\date-utils.d.ts

src/components/SQForm/SQFormDatePicker.tsx
```diff
@@ -1,9 +1,14 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import Grid from '@material-ui/core/Grid';
 import TextField from '@material-ui/core/TextField';
-import {DatePicker} from '@material-ui/pickers';
-import {makeStyles, ClickAwayListener} from '@material-ui/core';
+import moment from 'moment';
+import {DatePicker, BaseDatePickerProps} from '@material-ui/pickers';
+import BaseFieldProps from '../../types/BaseFieldProps';
+import {
```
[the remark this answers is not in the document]
shoaibhassan: Fixed.

src/components/SQForm/SQFormDatePicker.tsx
codyarose: This prop should be optional

codyarose (changes requested):

src/components/SQForm/SQFormDatePicker.tsx
shoaibhassan: @hvilander `useForm `use `React.ChangeEventHandler` type on `onChange` which is generic type. `useForm `didn't specifies the arguments type for `onChange` .  I am defining types of those arguments specifically and other than that the type of function remains the same. `useForm` is being used in almost every component So I think those component have their different type of arguments in `change. That is why they didn't specify the argument type in `useForm` itself

20BBrown14 (approved):

src/components/SQForm/SQFormDatePicker.tsx
```diff
@@ -1,9 +1,14 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import moment from 'moment';
```
codyarose, line 2: If we're only using the `Moment` type you could simplify things a bit and just destructure that here rather than using `moment.Moment`
```suggestion
import {Moment} from 'moment';
```

codyarose (approved):

Issue #355, opened 2021-07-19
Ability to disable SQFormButton when the `type` is 'reset'


SeanGroff:
We cannot control the `isDisabled` state of SQFormButton when the `type` prop is set to `'reset'`.

Here's the current code

```javascript
  const isSQFormButtonDisabled = React.useMemo(() => {
    if (isResetButton) {
      return !dirty;
    }

    return isButtonDisabled;
  }, [dirty, isButtonDisabled, isResetButton]);
```

We should be allowed to  control the `isDisabled` state of a reset button AND the reset button be disabled whenever the form is not dirty.

Here's a good use case

```javascript
<SQFormButton type="reset" title"Reset" isDisabled={isSubmitting}>Reset</SQFormButton>
```

PR #385, opened 2021-08-03
fix: 🐛 bug in SQFormButton to prioritize isDisabled over state

gdoohen1:
✅ Closes: #355

Fixed a bug in SQFormButton/useFormButton where the `isDisabled` prop was not used when the button was of type `reset`. Also updated useFormButton to accept an object instead of a multiple params since some params were not always provided.

### DIFF
--- a/src/components/SQForm/useFormButton.js
+++ b/src/components/SQForm/useFormButton.js
@@ -8,12 +8,12 @@ export const BUTTON_TYPES = {
   RESET: 'reset'
 };
 
-export function useFormButton(
-  isDisabled,
-  shouldRequireFieldUpdates,
+export function useFormButton({
+  isDisabled = false,
+  shouldRequireFieldUpdates = false,
   onClick,
   buttonType
-) {
+}) {
   const {values, initialValues, isValid, dirty, ...rest} = useFormikContext();
   const hasFormBeenUpdated = hasUpdated(initialValues, values);
 
@@ -48,10 +48,14 @@ export function useFormButton(
     dirty
   ]);
 
-  const handleClick = useDebouncedCallback((...args) => onClick(...args), 500, {
-    leading: true,
-    trailing: false
-  });
+  const handleClick = useDebouncedCallback(
+    (...args) => onClick?.(...args),
+    500,
+    {
+      leading: true,
+      trailing: false
+    }
+  );
 
   return {
     isButtonDisabled,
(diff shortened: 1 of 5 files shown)


src/components/SQForm/useFormButton.js
```diff
@@ -3,26 +3,59 @@ import {useFormikContext} from 'formik';
...
+};
+
+export function useFormButton({
+  isDisabled = false,
+  shouldRequireFieldUpdates = false,
+  onClick,
+  buttonType
+}) {
+  const {values, initialValues, isValid, dirty, ...rest} = useFormikContext();
   const hasFormBeenUpdated = hasUpdated(initialValues, values);
```
gdoohen1, line 18: [Question] Does anyone know if `hasFormBeenUpdated` is ever different from `dirty`? The Formik docs for `dirty` say that it is calculated by deep comparing `initialValues` and `values` but it looks like this is also the implementation of `hasUpdated` so not sure if these two ways of calculating basically the same thing are different/necessary.

SeanGroff: > [Question] Does anyone know if `hasFormBeenUpdated` is ever different from `dirty`? The [Formik docs](https://formik.org/docs/api/formik#dirty-boolean) for `dirty` say that it is calculated by deep comparing `initialValues` and `values` but it looks like this is also the implementation of `hasUpdated` so not sure if these two ways of calculating basically the same thing are different/necessary.

Forgot I wrote this util function awhile back. 
I'm pretty sure we can delete that `util` function which would also allow us to remove the fast-deep-equal dependency. I don't believe it's used anywhere else.

SeanGroff (approved):
Thanks for fixing this issue @gdoohen1 🥇

src/components/SQForm/useFormButton.js
```diff
@@ -3,26 +3,59 @@ import {useFormikContext} from 'formik';
...
+  SUBMIT: 'submit',
+  RESET: 'reset'
+};
+
+export function useFormButton({
+  isDisabled = false,
+  shouldRequireFieldUpdates = false,
+  onClick,
+  buttonType
+}) {
```
codyarose, line 16: I believe this is a breaking change. Consumers will need to switch their params to an object

codyarose (changes requested):

codyarose (approved):
I thought this was a breaking change at first, but then I remembered we don't export `useFormButton` 🥴

PR #396, opened 2021-08-05
refactor: 💡 Migrate SQFormDialog to TS

20BBrown14:
BREAKING CHANGE: 🧨 SQFormDialog's onSave prop is now required. New prop
`shouldDisplaySaveButton` will handle whether the "Save" button is
displayed or not. Previously this was done by the truthiness of
`onSave`.

✅ Closes: #395


package.json
```diff
@@ -84,7 +84,8 @@
     "react-window": "^1.8.5",
     "scplus-shared-components": "https://npm-public.selectquotelabs.com/scplus-shared-components/7.3.1",
     "text-mask-addons": "^3.8.0",
-    "use-debounce": "^6.0.0"
+    "use-debounce": "^6.0.0",
+    "yup": "^0.32.9"
```
20BBrown14, line 88: FWIW, there was a breaking change in Yup 0.31 that was "array().required() will no longer consider an empty array missing and required checks will pass." which we rely on for some of our components, name checkboxgroup. This is why your tests and ci are failing. So I think we need to ugprade yup in a seperate issue which will broke this and #395

codyarose (approved):

PR #401, opened 2021-08-06
refactor: 💡 Convert SQFormRadioButtonGroup to Typescript

hvilander:
✅ Closes: #400

The title says it all.

https://www.loom.com/share/67616ff2a8b34216b5808bf6a1c155e8

### DIFF
--- a/src/components/SQForm/SQFormRadioButtonGroup.tsx
+++ b/src/components/SQForm/SQFormRadioButtonGroup.tsx
@@ -1,12 +1,36 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-import Grid from '@material-ui/core/Grid';
-import RadioGroup from '@material-ui/core/RadioGroup';
-import FormHelperText from '@material-ui/core/FormHelperText';
-import FormControl from '@material-ui/core/FormControl';
-import FormLabel from '@material-ui/core/FormLabel';
+import {
+  Grid,
+  RadioGroup,
+  RadioProps,
+  FormHelperText,
+  FormControl,
+  FormLabel
+} from '@material-ui/core';
 import SQFormRadioButtonGroupItem from './SQFormRadioButtonGroupItem';
+import BaseFieldProps from 'types/BaseFieldProps';
 import {useForm} from './useForm';
+
+interface RadioButtonInputItemProps {
+  value: string | boolean | number;
+  label: string;
+  isDisabled?: boolean;
+  isRowDisplay?: boolean;
+  inputProps?: RadioProps;
+}
+
+interface SQFormRadioButtonGroupProps extends BaseFieldProps {
+  /** Function to call on value change */
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  /** Whether this radio selection is required */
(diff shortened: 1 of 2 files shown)


src/components/SQForm/SQFormRadioButtonGroup.tsx
```diff
@@ -1,13 +1,37 @@
...
+} from '@material-ui/core';
 import SQFormRadioButtonGroupItem from './SQFormRadioButtonGroupItem';
+import BaseFieldProps from 'types/BaseFieldProps';
 import {useForm} from './useForm';
 
+interface RadioButtonInputItemProps {
+  value: string | boolean | number;
+  label: string;
+  isDisabled?: boolean;
+  isRowDisplay?: boolean;
```
20BBrown14, line 18: I don't think the children passed to this array ened to worry about `isRowDisplay`, it's a derived value from `shouldDisplayInRow`

src/components/SQForm/SQFormRadioButtonGroup.tsx
```diff
@@ -1,13 +1,37 @@
...
+  inputProps?: RadioProps;
+}
+
+interface SQFormRadioButtonGroupProps extends BaseFieldProps {
```
20BBrown14, line 22: I don't think we want this interface to extend `BaseFieldProps` since we don't use a label prop, I'm surprised TS wasn't complaining since `label` is not optional on `BaseFieldProps`

20BBrown14 (changes requested):

20BBrown14 (approved):

src/components/SQForm/SQFormRadioButtonGroup.tsx
```diff
@@ -1,13 +1,40 @@
...
+  GridSize
+} from '@material-ui/core';
 import SQFormRadioButtonGroupItem from './SQFormRadioButtonGroupItem';
 import {useForm} from './useForm';
 
+interface RadioButtonInputItemProps {
+  value: string | boolean | number;
+  label: string;
+  isDisabled?: boolean;
+  inputProps?: RadioProps;
```
codyarose, line 18: Sort of a nitpick, but I think we should change this to PascalCase `InputProps` to follow the pattern used in some of our other components. 
`InputProps`: props that are given to the React component
`inputProps`: props (attributes) that are given to the HTML element

This would also call for a change to `SQFormRadioButtonGroupItem`'s prop name, because it's spreading that prop directly on a Mui Radio component instead of in the `inputProps` prop

hvilander: Not sure how to go about tracing down if the inputProps should be spreading into `inputProps`
It does look like MUI's Radio does take that prop `inputProps`
This would kinda come down to how the consumers are using it, right? I mean it seems to be a path to give Radio props specific to Radio. Maybe we don't want to expose that? 

I am going to go with your suggested change, but I have no strong opinion on if we should limit this to only providing inputProps or if this should expose all of RadioProps. 

Radio's docs for convenience https://material-ui.com/api/radio/

Issue #409, opened 2021-08-12
Migrate `SQFormMultiValue.js` to TS


20BBrown14:

20BBrown14:
Closed in #410

PR #410, opened 2021-08-12
refactor: 💡 Migrate SQFormMultiValue to TS

20BBrown14:
✅ Closes: #409

### DIFF
--- a/src/components/SQForm/SQFormMultiValue.tsx
+++ b/src/components/SQForm/SQFormMultiValue.tsx
@@ -1,27 +1,59 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import {Grid, Chip, TextField, makeStyles} from '@material-ui/core';
-import Autocomplete from '@material-ui/lab/Autocomplete';
-import {VariableSizeList} from 'react-window';
+import Autocomplete, {
+  AutocompleteChangeReason,
+  AutocompleteRenderInputParams,
+} from '@material-ui/lab/Autocomplete';
+import {ListChildComponentProps, VariableSizeList} from 'react-window';
 import {useField, useFormikContext} from 'formik';
-import {useForm} from './useForm';
 import {usePrevious} from '@selectquotelabs/sqhooks';
+import {useForm} from './useForm';
+import {BaseFieldProps, Option, optionValue} from 'types';
+import {
+  OuterElementContextInterface,
+  OuterElementTypeProps,
+} from './SQFormAutocomplete';
+
+interface SQFormMultiValueProps extends BaseFieldProps {
+  /** options to select from */
+  children: Option[];
+  /** Whether the field is required */
+  isRequired?: boolean;
+  /** Whether the field is disabled */
+  isDisabled?: boolean;
+  /** Custom handler for input value change */
+  onInputChange?: (
+    event: React.ChangeEvent<HTMLInputElement>,
+    value: optionValue
+  ) => void;
+  onChange?: (
+    event: React.ChangeEvent<HTMLInputElement>,
+    value: (string | Option)[],
+    reason: AutocompleteChangeReason,
+    detail?: string | Option
+  ) => void;
+  /** Custom onBlur event callback */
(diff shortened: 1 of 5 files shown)


src/components/SQForm/SQFormMultiValue.tsx
```diff
@@ -250,76 +288,61 @@ function SQFormMultiValue({
...
+          const endAdornment = Object.assign<
+            Record<string, Record<string, unknown>>,
+            Record<string, Record<string, unknown>>
+          >({}, params.InputProps.endAdornment || {});
+          if (endAdornment && endAdornment.props) {
+            endAdornment.props = {
+              ...(endAdornment.props || {}),
+              className: `${endAdornment.props.className} ${clearButtonClasses.endAdornment}`,
+            };
+          }
```
20BBrown14, line 317: I'm not so sure about this, particularly the type for `Object.assign`.

src/components/SQForm/SQFormMultiValue.tsx
```diff
@@ -1,28 +1,59 @@
...
+import {useForm} from './useForm';
+import {BaseFieldProps, Option, optionValue} from 'types';
+import {
+  OuterElementContextInterface,
+  OuterElementTypeProps,
+} from './SQFormAutocomplete';
+
+interface SQFormMultiValueProps extends BaseFieldProps {
+  /** options to select from */
+  children: Option[];
```
codyarose, line 18: Based on the previous propTypes, everything outside of `BaseFieldProps` should be optional

src/components/SQForm/SQFormMultiValue.tsx
```diff
@@ -250,76 +288,61 @@ function SQFormMultiValue({
...
-            inputProps={{
-              ...params.inputProps,
-              disabled: isDisabled
-            }}
-            FormHelperTextProps={{error: isFieldError}}
-            helperText={!isDisabled && HelperTextComponent}
-          />
-        )}
+        renderInput={(params) => {
+          console.log('params', params);
```
codyarose, line 306: Some leftover console.log and comments here

src/components/SQForm/SQFormMultiValue.tsx
codyarose: Oof, yeah it's kind of hard for me to follow what's going on with that at the moment, but maybe the `InputAdornmentProps` type could help here?

codyarose (changes requested):

src/components/SQForm/SQFormMultiValue.tsx
20BBrown14: So I was able to come up with a different solution that is far easier to read and gets the job done

src/components/SQForm/SQFormMultiValue.tsx
20BBrown14: Yep, everything except Children. It should actually be required even if the array is empty.

20BBrown14:
Sorry it took so long to update. This is ready for review again with all pending change requests completed.

20BBrown14:
🎉  All CI checks passing, and the changes I want are in the PR. Finally ready.

src/components/SQForm/SQFormMultiValue.tsx
```diff
@@ -115,6 +147,7 @@ function SQFormMultiValue({
   React.useEffect(() => {
     // Don't run this effect every time displayOptions changes
     // This also prevents infinite rerenders
+    // Maybe use deep-equal? This won't work if key order changes
```
hvilander, line 150: Was this comment left in intentionally? Seem like it might be a note for you.

hvilander (approved):

src/components/SQForm/SQFormMultiValue.tsx
20BBrown14: ya, as a note in case we run into issues later on related to this `useEffect`.

SeanGroff (approved):
Awesome!

PR #425, opened 2021-08-16
Issue/408 checkbox stories test to ts

hvilander:
Migrates the tests and stores for checkbox groups to Typescript.

### DIFF
--- a/stories/__tests__/SQFormCheckboxGroup.stories.test.tsx
+++ b/stories/__tests__/SQFormCheckboxGroup.stories.test.tsx
@@ -25,7 +25,7 @@ describe('SQFormCheckboxGroup Tests', () => {
       const group = screen.getByRole('group', {name: /shopping list/i});
       expect(group).toBeInTheDocument();
 
-      const checkboxes = screen.getAllByRole('checkbox');
+      const checkboxes = screen.getAllByRole('checkbox') as HTMLInputElement[];
       expect(checkboxes).toHaveLength(SHOPPING_LIST_OPTIONS.length);
     });
 
@@ -34,9 +34,9 @@ describe('SQFormCheckboxGroup Tests', () => {
         shoppingList: [SHOPPING_LIST_OPTIONS[0].value],
       };
 
-      render(<SQFormCheckboxGroup SQFormProps={{initialValues}} size="auto" />);
+      render(<SQFormCheckboxGroup initialValues={initialValues} size="auto" />);
 
-      const checkboxes = screen.getAllByRole('checkbox');
+      const checkboxes = screen.getAllByRole('checkbox') as HTMLInputElement[];
 
       const targetCheckbox = screen.getByLabelText(
         SHOPPING_LIST_OPTIONS[0].label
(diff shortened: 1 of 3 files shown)


stories/__tests__/SQFormCheckboxGroup.stories.test.tsx
```diff
@@ -25,25 +25,25 @@ describe('SQFormCheckboxGroup Tests', () => {
       const group = screen.getByRole('group', {name: /shopping list/i});
       expect(group).toBeInTheDocument();
 
-      const checkboxes = screen.getAllByRole('checkbox');
+      const checkboxes = screen.getAllByRole('checkbox') as HTMLInputElement[];
```
20BBrown14, line 28: This is interesting that we have to do this cast. Could you help explain why this was necessary?

hvilander: Probably by some ignorance, here was my thought process:
The getAllByRole method returns an HTMLElement. The test accesses the `checked` property of this element. Since `checked` is not on the HTMLElement but is on the HTMLInputElement which, my understanding is the more correct type for these check boxes. It _felt_ save to use the more specific type and it gave me access to the property I wanted. 

I am very typescript green so I may have done the dumb here, happy to learn a better way.

20BBrown14: Seems like this is a common issue and i found this SO thread about it. https://stackoverflow.com/questions/53195381/react-testing-library-with-typescript-set-an-inputs-value

I wonder if anybody has any opinions on this. We could use the help function as described in the answer instead of casting and throw an error or something if the element isn't the type we except so that the test fails? Not sure here. Casting all of our components in tests just seems like the worst. Definitely super weird that the the testing library functions don't utilize generics.

hvilander: I am cool either way. If the cast is bad and we don't have the values that HTMLInputElement provides when we assert our test, the test will just fail too. 

Since the cast just necessary for the test when using getAllByRole it doesn't feel too bad to me, but I will defer to another opinion.

hvilander: After some further investigation I am going to lean towards the cast. I played around and got to the snippet below. I don't feel like we are really gaining that much since getAllByRole('checkbox') will only return an html input element. The simple cast I think is kinda unique here since we need to use the checked property which for many of our tests we will not have to do. 

I am inclined to leave it as it is. 

```
    it('should check all checkboxes when select all checkbox checked', () => {
      render(<SQFormCheckboxGroup size="auto" shouldUseSelectAll />);


      const checkboxes = screen.getAllByRole('checkbox');
      expect(checkboxes).toHaveLength(6)


      const checkboxesAsInputs: HTMLInputElement[]  = checkboxes.map((checkbox) => {
        if(checkbox instanceof HTMLInputElement) {
          return checkbox as HTMLInputElement
        }

        throw new Error('this went so wrong I am not sure what message is useful')
      })

      const filteredCheckboxes = checkboxesAsInputs.filter(
        (checkbox) => checkbox.checked
      );
      expect(filteredCheckboxes).toEqual([]);
      expect(filteredCheckboxes).toHaveLength(0);

      const selectAllCheckbox = screen.getByLabelText('All');

      userEvent.click(selectAllCheckbox);

      const checkedCheckboxes = checkboxesAsInputs.filter(
        (checkbox) => checkbox.checked
      );
      expect(checkedCheckboxes).toEqual(
        expect.arrayContaining([selectAllCheckbox])
      );
      expect(checkedCheckboxes).toHaveLength(SHOPPING_LIST_OPTIONS.length + 1);
    });
    ```

20BBrown14 (approved):

PR #426, opened 2021-08-16
refactor: 💡 Migrate SQFormTextField stories to TS

20BBrown14:
Migrate SQFormTextField.stories.js and SQFormTextField.stories.test.js
to TS

✅ Closes: #424

### DIFF
--- a/stories/SQFormTextField.stories.tsx
+++ b/stories/SQFormTextField.stories.tsx
@@ -1,10 +1,24 @@
 import React from 'react';
 import * as Yup from 'yup';
+import {Story, Meta} from '@storybook/react';
 
 import {SQFormTextField as SQFormTextFieldComponent} from '../src';
-import getSizeProp from './utils/getSizeProp';
+import getSizeProp, {gridOptions} from './utils/getSizeProp';
 import {createDocsPage} from './utils/createDocsPage';
-import {SQFormStoryWrapper} from './components/SQFormStoryWrapper';
+import {
+  SQFormStoryWrapper,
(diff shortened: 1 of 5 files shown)


stories/SQFormTextField.stories.tsx
```diff
@@ -1,10 +1,24 @@
...
+import getSizeProp, {gridOptions} from './utils/getSizeProp';
 import {createDocsPage} from './utils/createDocsPage';
-import {SQFormStoryWrapper} from './components/SQFormStoryWrapper';
+import {
+  SQFormStoryWrapper,
+  SQFormStoryWrapperProps,
+} from './components/SQFormStoryWrapper';
+import {SQFormTextFieldProps} from '../src/components/SQForm/SQFormTextField';
+
+type SQFormTextFieldStoryType = Story<
```
20BBrown14, line 14: So this definitely seems weird plus will probably need to be made reuseable in some way.

Basically our stories needs to accept the props for the component that they're for minus the size prop as that should be coming from storybook args.
They also needs to accept all the SQFormSToryWrapper props so we can supply initialValues and validation schema etc.

codyarose: Yeah, the story wrapper is creating a bit of a mess with types. I've been wanting to revisit that and see if there's another way to do it without the story wrapper types muddying the component types, while also allowing us to easily change SQForm props for the tests

hvilander (approved):

codyarose (approved):

Issue #429, opened 2021-08-17
SQFormAutoComplete dynamic width is not working and causing unexpected behavior


Chris-Boyle:
In this PR https://github.com/SelectQuoteLabs/SQForm/pull/372 we introduced some drop down custom width logic.  Since consuming it in Senior we are seeing some unexpected behavior with the SQFormAutoComplete drop downs. 
Autocomplete width goes wider than the modal width causing a horizontal scroll
![image](https://user-images.githubusercontent.com/20329437/129743446-07d7c527-518b-4bc9-88d4-f8b139f1390c.png)

Senior issues and Storybook issues: https://www.loom.com/share/0e70d6f614b84cafb7b3e3d51340ff23

I can also see some unexpected behavior in [Storybook](https://5f4431386ea00a00220d495c-plvviwxlft.chromatic.com/?path=/story/forms-sqform--basic-form).
In our storybook this story should use the `lockwidth` prop to make the drop down with not wider than the field.  On initial render it is not working
![image](https://user-images.githubusercontent.com/20329437/129742498-6fc9da92-03ee-4649-b97a-e376375874b3.png)
The next time the drop down is clicked it does

PR #430, opened 2021-08-17
fix: 🐛 fixed autocomplete initial width issues and placement

gdoohen1:
Autocomplete should now have the correct width on first render and the popover should more reliably render below the field.
✅ Closes: #429

I was not able to replicate the issues mentioned in #429 as far as the positioning of the popover being in really weird places. Since the popover is positioned absolutely the only thing I can think of as to why it happens sometimes and not others is that some ancestor has some css that is throwing off the absolute positioning of the popover. So I added relative positioning to the grid item that contains the autocomplete and the popover so hopefully that lets the absolutely positioned popover render properly. Also, the default for lockWidthToField is now `true`, I think that maintains backwards compatibility better. An additional advantage to this however, is that when lockWidthToField is true we can re-enable react portal for the popover which should eliminate the absolute position issues in those cases entirely.

### DIFF
--- a/src/components/SQForm/SQFormAutocomplete.js
+++ b/src/components/SQForm/SQFormAutocomplete.js
@@ -276,7 +276,7 @@ function SQFormAutocomplete({
         name={name}
         style={{width: '100%'}}
         disableListWrap
-        disablePortal
+        disablePortal={!lockWidthToField}
         classes={autocompleteClasses}
         ListboxComponent={ListboxVirtualizedComponent}
         // Note: basewidth is not camel cased because React doesn't like it here
--- a/src/components/SQForm/SQFormAutocomplete.js
+++ b/src/components/SQForm/SQFormAutocomplete.js
@@ -18,27 +18,6 @@ const LISTBOX_PADDING = 8; // px
 
 const EMPTY_OPTION = {label: '- -', value: ''};
 
-const useStyles = makeStyles({
-  listbox: {
-    width: 'initial !important',
-    overflowX: 'hidden !important',
-
-    '& ul': {
-      padding: 0,
-      margin: 0
-    }
-  },
-  popper: {
-    borderRadius: '4px',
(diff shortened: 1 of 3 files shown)


Chris-Boyle:
@gdoohen1 can we get this tested in Senior before we merge just to confirm the actions drop down and the other autocompletes work?  Looking to get some more confidence in this so we don't have to do a trial and error workflow.

gdoohen1:
> @gdoohen1 can we get this tested in Senior before we merge just to confirm the actions drop down and the other autocompletes work? Looking to get some more confidence in this so we don't have to do a trial and error workflow.

@Chris-Boyle I absolutely agree - I have tried to test SQForm/shared-components in one of our docker based monolith repos before but I can never get docker to pull in the new .tgz file. If you know how to configure that though I would definitely love to do that though.

Chris-Boyle:
> > @gdoohen1 can we get this tested in Senior before we merge just to confirm the actions drop down and the other autocompletes work? Looking to get some more confidence in this so we don't have to do a trial and error workflow.
> 
> @Chris-Boyle I absolutely agree - I have tried to test SQForm/shared-components in one of our docker based monolith repos before but I can never get docker to pull in the new .tgz file. If you know how to configure that though I would definitely love to do that though.

And we have the steps documented finally!
https://selectquote.atlassian.net/wiki/spaces/SE/pages/2865332240/How+to+test+local+SQForm+changes+in+a+local+application. 
I did see some weird width when there was no drop down options in the list:
image
The options also seemed blurry on my external monitor but not on my Mac.  I would be interested to know if anyone else sees this
image

gdoohen1:
I noticed the package-lock was out of date, but I think it broke the pipeline with the mui theme import so I reverted that commit and the pipeline seems good now.

Chris-Boyle (approved):

codyarose (approved):

SeanGroff (approved):

PR #431, opened 2021-08-18
Migrate SQFormDateTimePicker to Typescript

hvilander:


src/components/SQForm/SQFormDateTimePicker.tsx
```diff
@@ -1,19 +1,37 @@
...
+  /** Disabled property to disable the input if true */
+  isDisabled?: boolean;
+  /** Required property used to highlight input and label if not fulfilled */
+  isRequired?: boolean;
+  /** Placeholder text used inside the input field to provide hints to the user */
+  placeholder?: string;
+  /** Custom onBlur event callback */
+  onBlur?: React.FocusEventHandler<HTMLInputElement>;
+  /** Custom onChange event callback */
+  onChange?: React.ChangeEventHandler;
```
hvilander, line 21: This seems to work, but I don't believe it is the correct type.

src/components/SQForm/SQFormDateTimePicker.tsx
```diff
@@ -1,19 +1,37 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import moment from '@material-ui/pickers/adapter/moment';
```
codyarose, line 2: Is this the same thing as `import moment from 'moment'`?

hvilander: It is not the same, as the adapter is not the same as the moment library itself. MUI supplies the adapter I believe for typing. I want to find some docs to be more sure.

hvilander: It looks like I am off base. I am not sure where I pulled my original line from. I will update it.

hvilander:
Please hold off on merging this. When I came back to this after letting it sit for awhile I am seeing a bug somewhere in validation.

src/components/SQForm/SQFormDateTimePicker.tsx
hvilander: The SQFormDatePicker did it differently and seems to work fine. I am going to align with that.

20BBrown14 (approved):

src/components/SQForm/SQFormDateTimePicker.tsx
```diff
@@ -1,10 +1,27 @@
...
 import {ClickAwayListener, makeStyles} from '@material-ui/core';
 import {useForm} from './useForm';
+import BaseFieldProps from '../../types/BaseFieldProps';
+import {ParsableDate} from '@material-ui/pickers/constants/prop-types';
+
+interface SQFormDateTimePickerProps extends BaseFieldProps {
+  /** Disabled property to disable the input if true */
+  isDisabled?: boolean;
+  /** Required property used to highlight input and label if not fulfilled */
+  isRequired?: boolean;
```
20BBrown14, line 15: I think we've gone away with this prop on our components since it should be automatic now

hvilander: quite right, removing it.

PR #438, opened 2021-08-26
Issue 363 - Add Tests for SQFormGuidedWorkflow

### LINKED ISSUE #363: SQFormGuidedWorkflow

laurelbean:
Added a set of tests for SQFormGuidedWorkflow.

Closes #363

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+### [5.6.3](https://github.com/SelectQuoteLabs/SQForm/compare/v5.6.2...v5.6.3) (2021-10-08)
+
+
+### Bug Fixes
+
+* 🐛 Only shows green fulfilled icon if field updated by user ([a9747fb](https://github.com/SelectQuoteLabs/SQForm/commit/a9747fbc9962591c7c8cbc5990cb2ecece6d2b1a)), closes [#459](https://github.com/SelectQuoteLabs/SQForm/issues/459)
+
 ### [5.6.2](https://github.com/SelectQuoteLabs/SQForm/compare/v5.6.1...v5.6.2) (2021-10-06)
 
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@selectquotelabs/sqform",
-  "version": "5.6.2",
+  "version": "5.6.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@selectquotelabs/sqform",
-  "version": "5.6.2",
+  "version": "5.6.3",
   "description": "SQ Form",
   "main": "dist/bundle.cjs.js",
   "private": false,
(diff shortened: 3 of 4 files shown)


stories/__tests__/SQFormGuidedWorkflow.stories.test.js
```diff
@@ -0,0 +1,249 @@
...
+    const interested = screen.getByText('Interested');
+    userEvent.click(interested);
+
+    expect(nextButton).toBeEnabled();
+  });
+
+  it('should close curent expanded card and open the next one when next button is clicked', async () => {
+    render(<SQFormGuidedWorkflow />);
+
+    const introText = screen.queryByText(/hi, bob smith, my name is/i);
```
codyarose, line 78: I believe this one should be `getByText` if we're testing that it's visible/exists

stories/__tests__/SQFormGuidedWorkflow.stories.test.js
```diff
@@ -0,0 +1,249 @@
...
+    const outcome = screen.getByRole('button', {name: /outcome/i});
+    userEvent.click(outcome);
+
+    const interested = screen.getByText('Interested');
+    userEvent.click(interested);
+
+    const nextButton = screen.getByRole('button', {name: /form submission/i});
+    userEvent.click(nextButton);
+
+    await waitFor(
```
codyarose, line 122: I feel like these `waitFor`s should be changed to `await findByText`s, but it seems like the timeout isn't working properly with that...
I'm still not very familiar with this component, but maybe some `await`ing for elements to be visible/not visible between those button clicks might help with the timing?

stories/__tests__/SQFormGuidedWorkflow.stories.test.js
```diff
@@ -0,0 +1,249 @@
...
+      () =>
+        expect(
+          screen.getByText(/stuff about policy cancellation/i)
+        ).toBeVisible(),
+      {timeout: 5000}
+    );
+
+    const policyText = screen.getByText(/stuff about policy cancellation/i);
+    expect(policyText).toBeVisible();
```
codyarose, line 131: I think this is made redundant by the `expect` on L124-126

stories/__tests__/SQFormGuidedWorkflow.stories.test.js
laurelbean: I used `queryByText` so that I could reuse it again on line 101 to test that it's NOT there. If I use `getByText` and it's not there then it will error out and fail the test.

stories/__tests__/SQFormGuidedWorkflow.stories.test.js
laurelbean: This line does have a wait for visible. I'm not sure I understand your comment. I can try using findByText instead. It's been a while so I can't remember if I tried that or not.
The big reason for the wait here is that part of the story includes a 3 second timer to demonstrate using a loading spinner for this particular model. And also the transitions between modules usually needs a wait of some kind.

laurelbean: Using `await findByText` doesn't seem to be working ☹️

laurelbean: I dug into this a little more and it seems like `findByText` is immediately returning since the text does exist in the DOM. However, it's not visible, which is what we want to check for. `waitFor` allows me to wait for visibility.

Issue #347, opened 2021-07-14
SQFormScrollableCard with Popover Menu for multiple forms


lucas-homer:
We need a version of the SQFormScrollableCard that enables showing a kabob menu to switch between multiple forms in the card content area. This means each menu-item/form will have its own `initialValues` and `validationSchema` and the Reset and Submit buttons will apply only to the in-view menu-item/form.

lucas-homer:
Thinking it might be easier (or at least it makes more sense in my head...) to make this a separate component? It'd be a kind of wrapper that uses a prop `menuOptions` for the popover feature, and each menu option maps to an item in a list of form config objects (containing initialValues, validationSchema, onSubmit, etc)

lucas-homer:
SQFormMultiViewCard?

PR #442, opened 2021-09-03
feat: 🎸 SQFormScrollableCardsMenuWrapper

lucas-homer:
A wrapper to map CardPopoverMenu items to instances of
SQFormScrollableCard to put many forms "into" one card.

✅ Closes: #347

[Made a loom with further explanation. ](https://www.loom.com/share/c323f1756ce94852acc0471693e04a27)

Tl;Dw -- Admin View project needs a way to use the CardPopoverMenu to flip between multiple instances of SQFormScrollableCard. This new component solves that problem, and my solution ended up largely motivated by making it easier to incorporate into existing code, so let me know what you think!

### DIFF
--- a/stories/SQFormScrollableCardsMenuWrapper.stories.js
+++ b/stories/SQFormScrollableCardsMenuWrapper.stories.js
@@ -18,31 +18,11 @@ export default {
 };
 
 export const sqFormScrollableCardsMenuWrapper = () => {
-  const menuItems = [
-    {
-      label: 'Details',
-      value: 'details'
-    },
-    {
-      label: 'Permissions',
-      value: 'permissions'
-    }
-  ];
-
   return (
     <div style={{width: '100%', height: '100%'}}>
-      <SQFormScrollableCardsMenuWrapper
-        title="[selected user]"
-        menuItems={menuItems}
-      >
-        <ScrollableDetails
-          label="Details" // maps to menuItems[0].label
-          value="details" // maps to menuItems[0].value
-        />
-        <ScrollablePermissions
-          label="Permissions" // maps to menuItems[1].label
-          value="permissions" // maps to menuItems[1].value
-        />
+      <SQFormScrollableCardsMenuWrapper title="[selected user]">
+        <ScrollableDetails label="Details" value="details" />
+        <ScrollablePermissions label="Permissions" value="permissions" />
       </SQFormScrollableCardsMenuWrapper>
     </div>
   );
--- a/stories/SQFormScrollableCardsMenuWrapper.stories.js
+++ b/stories/SQFormScrollableCardsMenuWrapper.stories.js
@@ -36,10 +36,12 @@ export const sqFormScrollableCardsMenuWrapper = () => {
         menuItems={menuItems}
       >
         <ScrollableDetails
-          id="details" // maps to menuItems[0].value
+          label="Details" // maps to menuItems[0].label
+          value="details" // maps to menuItems[0].value
         />
         <ScrollablePermissions
-          id="permissions" // maps to menuItems[1].value
+          label="Permissions" // maps to menuItems[1].label
+          value="permissions" // maps to menuItems[1].value
         />
       </SQFormScrollableCardsMenuWrapper>
     </div>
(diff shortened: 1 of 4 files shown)


stories/SQFormScrollableCardsMenuWrapper.stories.js
```diff
@@ -0,0 +1,153 @@
...
+
+  return (
+    <SQFormScrollableCard
+      isHeaderDisabled={true}
+      initialValues={initialValues}
+      onSubmit={handleSubmit}
+      shouldRequireFieldUpdates={true}
+      validationSchema={validationSchema}
+      height={calculatedHeight}
+      title="notApplicableBecauseHeaderDisabled" // bug in SQFormScrollableCard bc it errors if no title prop even though it doesn't render its cardheader
```
SeanGroff, line 93: I probably approved and merged the PR for this, but anyways, I don't think we need the `isHeaderDisabled` prop :(
If there's a `title` prop render a header, otherwise, don't render a header.

lucas-homer: That's fair, but outside scope of this PR/issue. Want to open an issue for that?

stories/SQFormScrollableCardsMenuWrapper.stories.js
```diff
@@ -0,0 +1,153 @@
...
+  ];
+
+  return (
+    <div style={{width: '100%', height: '100%'}}>
+      <SQFormScrollableCardsMenuWrapper
+        title="[selected user]"
+        menuItems={menuItems}
+      >
+        <ScrollableDetails
+          id="details" // maps to menuItems[0].value
```
SeanGroff, line 39: I think we can avoid passing in menuItems and an `id` prop that matches an array item from menuItems.
We can absorb the menuItems pattern into the `children`

```javascript
      <SQFormScrollableCardsMenuWrapper
        title="[selected user]"
      >
      <ScrollableDetails label="Details" value="details" />
      {/* etc, etc for each child */}
```

Then inside our SQScrollableCardsWrapper we can build the menuItems array dynamically.

```javascript
const menuItems = React.Children.map(child => {
  return {label: child.props.label, value: child.props.value}
})
```

stories/SQFormScrollableCardsMenuWrapper.stories.js
SeanGroff: > That's fair, but outside scope of this PR/issue. Want to open an issue for that?

Yep, I meant for this to be a single comment and not part of the review. Woopsie

stories/SQFormScrollableCardsMenuWrapper.stories.js
lucas-homer: > I think we can avoid passing in menuItems and an `id` prop that matches an array item from menuItems.
> We can absorb the menuItems pattern into the `children`
> 
> ```js
> // pseudo code
> ```
> 
> ```js
>       <SQFormScrollableCardsMenuWrapper
>         title="[selected user]"
>       >
>       <ScrollableDetails label="Details" value="details" />
>       {/* etc, etc for each child */}
>       </SQFormScrollableCardsMenuWrapper>
> ```
> 
> Then inside our SQScrollableCardsWrapper we can build the menuItems array dynamically.
> 
> ```js
> const menuItems = React.Children.map(child => {
>   return {label: child.props.label, value: child.props.value}
> })
> ```
> 
> * Edit to add obligatory pseudo code warning

I like this idea! Should work well.

lucas-homer: done

stories/SQFormScrollableCardsMenuWrapper.stories.js
```diff
@@ -0,0 +1,155 @@
...
+      label: 'Permissions',
+      value: 'permissions'
+    }
+  ];
+
+  return (
+    <div style={{width: '100%', height: '100%'}}>
+      <SQFormScrollableCardsMenuWrapper
+        title="[selected user]"
+        menuItems={menuItems}
```
SeanGroff, line 36: With the recent changes, we can remove this prop from the story
We can also remove the comments below mentiong what label and value map to within menuItems because the consumer doesn't need to know those implementation details anymore since menuItems is dynamically generated.

SeanGroff (approved):
My latest comment is just story/documentation cleanup.
Code LGTM!

PR #443, opened 2021-09-07
fix: 🐛 fix isSelfBounding issue in SQFormScrollableCard

lucas-homer:
✅ Closes: 436

I know this is probably not the ideal solution to the problem, but SQFormScrollableCard is pretty critical to MIAV and we need to address the issue with `isSelfBounding` not working and causing us to write more custom add-on stuff inside the MIAV. We know the code proposed here works, and consuming `useAutoHeight` with other components seems to work as expected, but we've investigated this issue within SQFormScrollableCard and cannot seem to figure it out. Our short term solution is to go back to what worked before. In the future, we can devote more time to figuring out why `useAutoHeight` doesn't work in this particular case, but this PR gets folks a working SQFormScrollableCard.

### DIFF
--- a/stories/SQFormScrollableCardsMenuWrapper.stories.js
+++ b/stories/SQFormScrollableCardsMenuWrapper.stories.js
@@ -43,27 +43,6 @@ function ScrollableDetails() {
     actions.resetForm();
   };
 
-  /**
-   * TODO: This code can/should be removed once this issue is resolved - https://github.com/SelectQuoteLabs/SQForm/issues/436
-   * Once this problem is resolved, we can remove this code block below and also remove the wrapper container around our SQFormScrollableCard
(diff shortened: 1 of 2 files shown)


stories/SQFormScrollableCardsMenuWrapper.stories.js
```diff
@@ -43,35 +43,13 @@ function ScrollableDetails() {
     actions.resetForm();
   };
 
-  /**
```
lucas-homer, line 46: these changes are ancillary, and after putting it back into SQFormScrollableCard, are unnecessary here

jamesgriffith9519 (approved):

20BBrown14 (approved):

PR #446, opened 2021-09-20
refactor: 💡 remove unnecessary prop - SQFormScrollableCard

jamesgriffith9519:
Recently we had done an enhancement that would allow the user to disable the rendering of a header based on a true/false boolean prop value.

However I realize now that we really didn't need a specific prop for this, rather we could simply check if we have a title set or not. 

Loom: https://www.loom.com/share/9563e54ab7c141b1b9c21fdfd2de4616

### DIFF
--- a/src/components/SQFormScrollableCard/SQFormScrollableCard.js
+++ b/src/components/SQFormScrollableCard/SQFormScrollableCard.js
@@ -7,7 +7,7 @@ import {
   CardContent,
   CardActions,
   Grid,
-  makeStyles
+  makeStyles,
 } from '@material-ui/core';
 import {Formik, Form} from 'formik';
 import {useDebouncedCallback} from 'use-debounce';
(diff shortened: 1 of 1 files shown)


src/components/SQFormScrollableCard/SQFormScrollableCard.js
```diff
@@ -240,7 +240,7 @@ SQFormScrollableCard.propTypes = {
    * */
   validationSchema: PropTypes.object,
   //** Boolean used to determine if title/header is enabled or disabled */
-  isHeaderDisabled: PropTypes.bool
+  isHeaderDisabled: PropTypes.bool,
```
gdoohen1, line 243: I think this can be removed form propTypes.

gdoohen1 (approved):
This does maybe break backwards compatibility/functionality so not sure if we want to be extra cautious and not do this or wait till a major version bump. Although, since `title` is an optional prop, I imagine if `isHeaderDisabled` was passed as true then `title` was omitted so this probably wouldn't cause any change in behavior for consumers. Probably.

src/components/SQFormScrollableCard/SQFormScrollableCard.js
jamesgriffith9519: It looks like the new MenuWrapper component uses the isHeaderDisabled Prop, I think ill hold off on making this change until i look at that new component a bit closer, just incase!

SeanGroff:
Closing this PR now that we've migrated to TypeScript.
I'll leave the issue open so we can look into addressing this in the TS codebase if necessary

Issue #465, opened 2021-10-25
SQFormDialog scroll bar is on the whole modal and not just body


Chris-Boyle:
When consuming the SQFormDialog if the body needs to scroll to view all the components there is a scroll bar on the right.  The scroll bar goes from the top of the dialog to the bottom instead of just on the dialog body/content.  UX has requested we rectify this with the scroll bar just on the the body of the dialog similar to how the SQFormScrollableCard functions.

To reproduce this in SQForm I added a lot of components to our basic dialog story.

![image](https://user-images.githubusercontent.com/20329437/138753937-ff5f3326-5ecb-4a65-be83-8a6e25e80ca3.png)
![image](https://user-images.githubusercontent.com/20329437/138753952-e1f5de44-3827-4e9e-99e2-7657bc7807c1.png)

Loom: https://www.loom.com/share/2a592913dbfb4c958f92958188695df9

PR #469, opened 2021-11-09
fix: 🐛 Change scrollbar to only Dialog body

aharpt:
Changes scrollbar for SQFormDialog component to be just on the body of the modal, instead of the entire modal.

https://www.loom.com/share/660302d94c8d4fce8e446be1b3b6d3a4

✅ Closes: #465

### DIFF
--- a/src/components/SQFormDialog/SQFormDialogInner.js
+++ b/src/components/SQFormDialog/SQFormDialogInner.js
@@ -46,7 +46,7 @@ const useActionsStyles = makeStyles({
 });
 const useDialogContentStyles = makeStyles({
   root: {
-    overflowY: 'visible',
+    maxHeight: '75vh',
     padding: '20px'
   }
 });
@@ -94,6 +94,7 @@ function SQFormDialogInner({
   return (
     <>
       <Dialog
+        scroll="body"
         disableBackdropClick={disableBackdropClick}
         maxWidth={maxWidth}
         open={isOpen}


src/components/SQFormDialog/SQFormDialogInner.js
```diff
@@ -46,7 +46,7 @@ const useActionsStyles = makeStyles({
 });
 const useDialogContentStyles = makeStyles({
   root: {
-    overflowY: 'visible',
+    maxHeight: '75vh',
```
hvilander, line 49: A question, why was it necessary to set a max height here?

hvilander (approved):

src/components/SQFormDialog/SQFormDialogInner.js
Chris-Boyle: I bet it was because without a set height we don't have a scroll area to scroll within.  That is why it likely just scrolled the full parent container.  This is setting a height on the child to make that the scrollable area.  `75vh` not sure how that was calculated though.

aharpt: If the maxHeight is not set, the dialog will simply overflow the page and you will simply have to scroll down the page to see the end of the modal.  See the attached screenshot.

sqFormDialog maxHeight

Chris-Boyle: Are you guessing that 25vh is how much room the header and footer take up?  75vh sounds mysterious.  How did you get that number?  I would also test the scenario that if we don't have enough fields for a scrollbar to show does it still look good.  A loom showing everything you test is best so that there aren't any questions about what was tested.

20BBrown14: Setting dialog conent dimensions based on the view port doesn't seem like the right choice. Ultimately would it be better to set the dialog to be a `#vh` tall and then the content be `#%` of that? While the footer and header will be another `#%`?

As the view port gets smaller the content is always utilize 75% of the view port while the dialog itself may be taking up a smaller amount.

aharpt: Attached is a video of how the dialog looks without any scrollbar needed.  It does not always take up 75% of the browser's viewport since I set this as maxHeight not height.  I can certainly set it to your suggestion @20BBrown14 where I explicitly set the entire dialog, than the content, header, and footer.  Let me know...

https://www.loom.com/share/53ace8a1fbb74bad93202fa02bc7c71d

20BBrown14: You're right, sorry. I misread the style rule. I think setting the maxHeight to be relative to the dialog itself instead of the view port would make more sense though. Hm. SQForm doesn't have a lot of styling decisions made quite yet. I'm definitely interested in getting more input, though.

aharpt: No worries.  I'll just wait for others to give their opinions so we can come to a consensus.

SeanGroff:
Closing this PR now that we've migrated to TypeScript.
I'll leave the issue open so we can look into addressing this in the TS codebase if necessary

Issue #482, opened 2021-12-30
SQFormDialog expose a way to hide the secondary button


Chris-Boyle:
We have a use case where we need to force a user to make a valid selection in the dialog and click the primary SAVE button before we allow the dialog to close.  
To achieve this the UX has asked that we add an enhancement to our current SQFormDialog so that a consumer has the ability to hide the secondary button.

The plan is to add a prop `showSecondaryButton` that will be defaulted to true.  If set to false by a consumer then we will not show the secondary CANCEL button.

PR #483, opened 2021-12-30
Added ability to hide the secondary button in the SQFormDialog

Chris-Boyle:
UX has requested we offer a way to hide the secondary button to enforce users to take an action in a dialog.  

Closes: #482 

Loom: https://www.loom.com/share/3469d74003e5468991249b27f832d4ef

### DIFF
--- a/src/components/SQFormDialog/SQFormDialogInner.js
+++ b/src/components/SQFormDialog/SQFormDialogInner.js
@@ -186,7 +186,7 @@ SQFormDialogInner.propTypes = {
   title: PropTypes.string.isRequired,
   /** Any prop from https://material-ui.com/api/grid */
   muiGridProps: PropTypes.object,
-  /** show/hide the secondary Cancel button.  Defaults to show(true) */
+  /** show or hide the secondary Cancel button.  Defaults to show(true) */
   showSecondaryButton: PropTypes.bool
 };
 
--- a/src/components/SQFormDialog/SQFormDialogInner.js
+++ b/src/components/SQFormDialog/SQFormDialogInner.js
@@ -104,7 +104,7 @@ function SQFormDialogInner({
         maxWidth={maxWidth}
         open={isOpen}
         TransitionComponent={Transition}
-        onClose={handleCancel}
+        onClose={showSecondaryButton ? handleCancel : null}
       >
         <Form>
           <DialogTitle disableTypography={true} classes={titleClasses}>
--- a/src/components/SQFormDialog/SQFormDialogInner.js
+++ b/src/components/SQFormDialog/SQFormDialogInner.js
@@ -33,7 +33,7 @@ const useTitleStyles = makeStyles({
     borderBottom: ({palette}) => `1px solid ${palette.divider}`
   }
 });
-const useActionsStyles = makeStyles({
+const actionStyles = {
   root: {
     display: 'flex',
     justifyContent: 'space-between',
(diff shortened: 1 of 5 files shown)


aharpt (approved):

SeanGroff (changes requested):
Code looks good.
Commit doesn't follow the conventional commit pattern which means this code wouldn't be released to NPM

20BBrown14 (approved):
Approved pending conventional commit changes

Chris-Boyle:
> Code looks good. Commit doesn't follow the conventional commit pattern which means this code wouldn't be released to NPM

Ahhh yes.  It has been a while.  Let me spin up an enhancement/new feature commit

SeanGroff:
@Chris-Boyle This is a reminder to myself to communicate this to the frontend-devs channel so everyone knows going forward.

Whenever we enhance a component we need to add a test(s) for our enhancement.

For this component: https://github.com/SelectQuoteLabs/SQForm/blob/master/stories/__tests__/SQFormDialog.stories.test.js

Looking at the existing tests, i'm a little concerned about a bug I didn't think about. I believe if the user hits the escape key the modal will still close even if we have this prop set to false.

We should write a test for this, and fix the code if needed.

We should also write a quick test that the close button does NOT render when our new prop is set to false.

SeanGroff (changes requested):
The commit now follows the conventional commit.
Please add the tests to reflect these changes then this work should be ready to merge. IMO

src/components/SQFormDialog/SQFormDialogInner.js
```diff
@@ -98,7 +104,7 @@ function SQFormDialogInner({
         maxWidth={maxWidth}
         open={isOpen}
         TransitionComponent={Transition}
-        onClose={handleCancel}
+        onClose={showSecondaryButton ? handleCancel : null}
```
SeanGroff, line 107: I think we want to provide a `noop` instead of `null` here. 
I haven't looked at the MUI `types` for Dialog, but I don't think `onClose` accepts `null`. I know it will accept `undefined` or `() => {}`

`undefined` probably makes the most sense.

stories/__tests__/SQFormDialog.stories.test.js
```diff
@@ -96,6 +96,31 @@ describe('Tests for Default', () => {
   });
 });
 
+it('should not call onClose on `escape` keydown because cancel is not available', async () => {
```
SeanGroff, line 99: This test looks solid to me. To fully test this code enhancement I think we need the test where we set `showSecondaryButton={false}` and then we expect the Close button to NOT be in the document or visible.

SeanGroff (changes requested):

stories/__tests__/SQFormDialog.stories.test.js
Chris-Boyle: In this test we set `showSecondaryButton={false}` and then query the screen for the button and assert it doesn't exist.  Is that what you are outlining?

SeanGroff: > In this test we set `showSecondaryButton={false}` and then query the screen for the button and assert it doesn't exist. Is that what you are outlining?

I think this is the same question we discussed in Slack. (We wan't to test whether or not the button appears on the screen as its own test)

stories/__tests__/SQFormDialog.stories.test.js
```diff
@@ -96,6 +96,46 @@ describe('Tests for Default', () => {
...
+      isOpen={true}
+      onSave={handleSave}
+      onClose={handleClose}
+      showSecondaryButton={false}
+    />
+  );
+
+  expect(screen.queryByRole('button', {name: /cancel/i})).toBeNull();
+
+  await waitFor(() => {
```
SeanGroff, line 134: I think this snuck in via copy/paste on accident.
There's no user event in this test

SeanGroff (approved):

PR #487, opened 2022-01-04
fix(SQFormGuidedWorkflow.js):  add style prop for container of guided…

shahidmanzoorzekiexperts:
We need to pass a dynamic style to the container of SQFormGuidedWorkflow. 
currently we have <Section style={{padding: '20px'}}>
Which is fixed to 20px. In senior, we need a dynamic padding in our scenario we need it to 0px.
That's why we need a style prop here.


SeanGroff (changes requested):
Code looks good but the comment doesn't follow the conventional commit pattern
Please follow the instructions in the readme and use the script `npm run commit`

src/components/SQFormGuidedWorkflow/SQFormGuidedWorkflow.js
```diff
@@ -176,7 +177,7 @@ function SQFormGuidedWorkflow({
   });
 
   return (
-    <Section style={{padding: '20px'}}>
+    <Section style={containerStyles}>
```
20BBrown14, line 180: How do we feel about `style={{padding: '20px', ...containerStyles}}` which will always give the 20px padding without the consumer having to define it if they want to add _other_ styles. If the consumer provides `padding` it'll override the 20px. But if they provide, for instance, `margin` our current solution will then nuke the padding which would mean the consumer would need to provide that padding as well if they want to keep it.

Plus, then we could make the default value for `containerStyles` to be `{}` which reads a bit easier and could be a bit easier to maintain in the future.

20BBrown14 (changes requested):

src/components/SQFormGuidedWorkflow/SQFormGuidedWorkflow.js
SeanGroff: Agreed 💯 
I didn't notice the default padding was removed.

src/components/SQFormGuidedWorkflow/PropTypes.js
```diff
@@ -108,5 +108,7 @@ export const GuidedWorkflowProps = {
   onError: PropTypes.func.isRequired,
   /** Task Module configuration Object(s) */
   taskModules: PropTypes.arrayOf(PropTypes.shape(TaskModuleProps).isRequired)
-    .isRequired
+    .isRequired,
+  /** An object of css-in-js style properties to be passed*/
```
SeanGroff, line 112: Our Storybook depends on a properly formatted JS Doc comment.
A space is missing between passed and */

20BBrown14 (approved):

PR #489, opened 2022-01-04
feat: 🎸 truncate right card title when text is longer than box

jamesgriffith9519:
There was a small ask in the sc3 admin project to truncate some of the titles and I noticed this component was the one that we are using in quite a few of our views. I added a few lines of CSS that take care of truncating the text if it's larger than the container. There were a few other ways to handle this with other CSS properties but they were producing some weird behaviors and I found using the WebKit properties seemed to work best for me. 

I'm curious what others think and perhaps we can find a better way to handle it! 

https://www.loom.com/share/c33a75f2aac946f89c3c503d55841d9e


20BBrown14 (approved):
🔥

PR #492, opened 2022-01-05
fix: 🐛 add style prop for container of guided workflow

shahidmanzoorzekiexperts:
We need to pass a dynamic style to the container of SQFormGuidedWorkflow. currently we have <Section style={{padding: '20px'}}>
Which is fixed to 20px. In senior, we need a dynamic padding in our scenario we need it to 0px.
That's why we need a style prop.


SeanGroff (approved):
Thanks!

Issue #394, opened 2021-08-05
Migrate `SQFormMultiSelect.js` to TS


20BBrown14:
In-progress on https://github.com/SelectQuoteLabs/SQForm/tree/issue/394

20BBrown14:
Blocked by #337

Issue #332, opened 2021-07-12
Migrate `SQFormDropdown.js` to TS


SeanGroff:
Convert SQFormDropdown to TypeScript

20BBrown14:
Closed with #493

PR #493, opened 2022-01-05
refactor: 💡 Add generics to useForm hook

20BBrown14:
Added generics to useForm hook. Also converted SQFormMultiSelect to TS
and SQFormDropdown to TS, and updated other components that use the hook
as proof that the generics work

✅ Closes: #337
✅ Closes: #394
✅ Closes: #332 
See #338 for conversation, code, and loom videos that explain the situation _super_ well and was largely the inspiration for the changes in this PR.

### DIFF
--- a/src/components/SQForm/SQFormCheckboxGroup.tsx
+++ b/src/components/SQForm/SQFormCheckboxGroup.tsx
@@ -25,7 +25,7 @@ interface SQFormCheckboxGroupProps {
   /** Whether a selection in this group is required */
   isRequired?: boolean;
   /** Function to call on value change */
-  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  onChange?: CheckboxProps['onChange'];
   /** Whether to display the group in a row */
   shouldDisplayInRow?: boolean;
   /** Whether to display the select all checkbox */
@@ -48,7 +48,10 @@ function SQFormCheckboxGroup({
   const {
     fieldState: {isFieldError, isFieldRequired},
     fieldHelpers: {handleChange, handleBlur, HelperTextComponent},
-  } = useForm({name, onChange});
+  } = useForm<CheckboxOption['value'][], React.ChangeEvent<HTMLInputElement>>({
+    name,
+    onChange,
+  });
 
   const {setFieldValue} = useFormikContext();
 
--- a/src/components/SQForm/SQFormCheckboxGroupItem.tsx
+++ b/src/components/SQForm/SQFormCheckboxGroupItem.tsx
@@ -3,6 +3,7 @@ import {FormControlLabel, CheckboxProps} from '@material-ui/core';
 import Checkbox from '@material-ui/core/Checkbox';
 import {makeStyles} from '@material-ui/core/styles';
 import {useForm} from './useForm';
+import {Option} from 'types';
 
 const useStyles = makeStyles((theme) => ({
   checkboxGroupItem: {
@@ -21,7 +22,7 @@ interface SQFormCheckboxGroupItemProps {
   /** Value for the checkbox */
   value: string | boolean | number;
   /** Function to call when input value is changed */
-  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  onChange: CheckboxProps['onChange'];
   /** Whether this group item is part of a group displayed in a row */
   isRowDisplay?: boolean;
   /** Whether the checkbox is disabled */
@@ -42,7 +43,10 @@ function SQFormCheckboxGroupItem({
   const {
     formikField: {field},
     fieldHelpers: {handleChange},
-  } = useForm({name: groupName, onChange});
+  } = useForm<
+    Option['value'][] | Option['value'],
+    React.ChangeEvent<HTMLInputElement>
+  >({name: groupName, onChange});
 
   const classes = useStyles();
 
--- a/src/components/SQForm/SQFormDropdown.tsx
+++ b/src/components/SQForm/SQFormDropdown.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import FormControl from '@material-ui/core/FormControl';
-import Select from '@material-ui/core/Select';
+import Select, {SelectProps} from '@material-ui/core/Select';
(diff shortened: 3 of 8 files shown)


src/components/SQForm/SQFormCheckboxGroupItem.tsx
```diff
@@ -3,6 +3,7 @@ import {FormControlLabel, CheckboxProps} from '@material-ui/core';
 import Checkbox from '@material-ui/core/Checkbox';
 import {makeStyles} from '@material-ui/core/styles';
 import {useForm} from './useForm';
+import {Option} from 'types';
```
hvilander, line 6: Can `import type` be used here?

SeanGroff:
@codyarose is a required reviewer for this PR

src/components/SQForm/SQFormRadioButtonGroup.tsx
```diff
@@ -47,7 +48,15 @@ function SQFormRadioButtonGroup({
...
-  } = useForm({name, onChange});
+  } = useForm<
+    RadioButtonInputItemProps['value'],
+    React.ChangeEvent<HTMLInputElement>
+  >({
+    name,
+    onChange,
+  });
+
+  console.log('field', field);
```
aharpt, line 59: Just curious if this is supposed to be left in?

src/components/SQForm/SQFormDropdown.tsx
```diff
@@ -16,9 +15,29 @@ import {
   getUndefinedValueWarning,
 } from '../../utils/consoleWarnings';
 import {EMPTY_LABEL} from '../../utils/constants';
+import {BaseFieldProps, Option} from 'types';
```
aharpt, line 18: Can we do `import type` here as well?

src/components/SQForm/SQFormDropdown.tsx
```diff
@@ -112,15 +133,17 @@ function SQFormDropdown({
           onBlur={handleBlur}
           onChange={handleChange}
           labelId={labelID}
-          renderValue={renderValue}
+          renderValue={(value: unknown) =>
+            renderValue(value as Option['value'])
+          }
```
hvilander, line 138: [nit / suggestion]
Since you are defining renderValue on line 94 in this file. It might make sense to push these two functions together.

```
94. const renderValue = (value: unknown) {
  const optionValue = value as Option['value'];
  ...
  }
  ```
  
  Right now the function definition is kinda split and decreases readability.

File: src/components/SQForm/SQFormMultiSelect.tsx [hunk 5]
```diff
@@ -21,6 +22,31 @@ import {
   getUndefinedChildrenWarning,
   getUndefinedValueWarning,
 } from '../../utils/consoleWarnings';
+import {BaseFieldProps, Option} from 'types';
```
aharpt, line 25: Maybe `import type` here as well?

src/components/SQForm/SQFormDropdown.tsx
20BBrown14: Good call. I did this to try to make TS happy. the `renderValue` prop expects a function who's first param is of `unknown` value. Even though it's entirely possible for us, the consumer, to know exactly what type the param will be. But there isn't a generic we can provide to tell the MUI component what the type is. So we have to cast it. Which kinda sucks. But it makes our _actual_ renderValue function type safe.

I'll smoosh em together for sure. There's another instance of this in another component, MultiSelect I beleive, i will change it there as well.

hvilander (approved):
+1 pending some nit cleanup

src/components/SQForm/SQFormRadioButtonGroup.tsx
codyarose, line 59: Stray console.log 👀

codyarose (approved):
Other than the stray console.log I think this lgtm 👍

src/components/SQForm/SQFormRadioButtonGroup.tsx
20BBrown14: Should we add a rule to eslint to catch these for us? I know I'm not the only one that does this lol

codyarose: Yeah I imagine there's a rule for it. As long as it isn't a pain/blocker while we're actively working on something

20BBrown14: Ya. Our dev environment is super stingy with eslint errors and blocks the build/compile when it seems them. That may be why we don't have that rule in place. I wonder how hard it'd be to setup using a different set of rules in our CI and precommit `npm run lint` checks to catch things like this but them through while actively developing and running `npm run storybook`

lucas-homer:
great work @20BBrown14 thanks for getting us over the hump on this

caleboleary (approved):

src/components/SQForm/SQFormRadioButtonGroup.tsx
SeanGroff: Setting console logs to warn instead of error will prevent the dev environment (storybook) from failing the compilation.
To my knowledge, the `npm run lint` script will only fail on errors, not warnings.

```
// eslint config
  "rules": {
    "no-console": "warn",
    "no-debugger":"warn"
  }
  ```

File: src/components/SQForm/SQFormCheckboxGroup.tsx [hunk 6]
```diff
@@ -25,7 +25,7 @@ interface SQFormCheckboxGroupProps {
   /** Whether a selection in this group is required */
   isRequired?: boolean;
   /** Function to call on value change */
-  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  onChange?: CheckboxProps['onChange'];
```
SeanGroff, line 28: Love this

20BBrown14: I think this was part of the key. Ensuring that the change handlers our components accept were the same as what MUI wants and then informing Formik about the type of the change handler. Instead of trying to keep both libraries happy.

SeanGroff (approved):
Amazing work @20BBrown14

Issue #499, opened 2022-01-06
Fix failed prop type: justify of ForwardRef(Grid) is deprecated


laurelbean:
Related to #458 

The prop `justify` on `<Grid>` components was changed to be `justifyContent`. So all instances of `justify` on a Grid component need updating.
These logs were another one clogging up testing logs.

PR #500, opened 2022-01-06
fix: 🐛 Update deprecated justify prop for Grid

laurelbean:
The justify prop on Grid components was renamed to justifyContent. After
updating that I noticed that SQFormDialogStepper had 2 new error logs.
So I added defaults for missing required props and added a type to the
children PropTypes.

✅ Closes: #499


20BBrown14 (approved):
Gotta love cleaning up tech debt

SeanGroff (approved):

Issue #494, opened 2022-01-06
Our TS type imports should use `import type`


20BBrown14:
While in the process of converting this library to TS we found that `import type` exists (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html). We should update all instances of importing a type to use `import type` and then continue to use while converting and developing the TS library.

related: #244

SeanGroff:
Haha I was just typing this issue our myself. Thanks for adding this!

PR #505, opened 2022-01-06
refactor: 💡 Change all type imports to 'import type'

aharpt:
Change all type imports to 'import type'

https://www.loom.com/share/a3c187517a15441bb2f17a829946098a

✅ Closes: #494


aharpt:
In a couple of files there are trailing commas added or added optional parentheses around a single parameter on an arrow function.  I did not add these but were unintentionally added by my prettier VSCode extension (which I have since uninstalled).  If people would prefer, I can do a follow-up commit to undo those changes.

src/components/SQForm/SQFormDateTimePicker.tsx
```diff
@@ -5,8 +5,8 @@ import TextField from '@material-ui/core/TextField';
 import {DateTimePicker, BaseDateTimePickerProps} from '@material-ui/pickers';
 import {ClickAwayListener, makeStyles} from '@material-ui/core';
 import {useForm} from './useForm';
-import BaseFieldProps from '../../types/BaseFieldProps';
-import {ParsableDate} from '@material-ui/pickers/constants/prop-types';
+import type BaseFieldProps from '../../types/BaseFieldProps';
```
lucas-homer, line 8: [question] don't we need to import this as if it's a named export (`{ BaseFieldProps }`) instead of default?

src/components/SQForm/SQFormDatePicker.tsx
```diff
@@ -8,7 +8,7 @@ import {
   ClickAwayListener,
   InputBaseComponentProps,
 } from '@material-ui/core';
-import BaseFieldProps from '../../types/BaseFieldProps';
+import type BaseFieldProps from '../../types/BaseFieldProps';
```
lucas-homer, line 11: [nit/suggestion] it'd be snazzy to move these to use absolute import paths. `'../../types/'` vs `'types/'`

lucas-homer (changes requested):
thanks for tacking this one @aharpt ! The prettier changes aren't too bad, but otoh maybe that'd mean it's easy to undo them. Going forward it looks like you need to make sure your Pretter extension respects the project's eslintrc, which it should, or at very least set `trailingComma: false` in your own extension's settings

src/components/SQForm/SQFormDateTimePicker.tsx
aharpt: @lucas-homer Thanks for the feedback!  BaseFieldProps actually is the default export from the imported file.

20BBrown14: I think Lucas was mentioning that we should change this import to import from the `types` index file instead of the `BaseFieldProps` file itself.

aharpt: Ah. Thanks for clarifying @20BBrown14.

lucas-homer (approved):

hvilander (approved):

src/components/SQForm/SQFormCheckbox.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import {Checkbox, CheckboxProps} from '@material-ui/core';
```
20BBrown14, line 2: So CheckboxProps is a type that we'd like to use `import type` on. However, we can't do `import type {Checkbox, CheckboxProps} from '@material-ui/core';` since `Checkbox` isn't a type and we need it at run time.

So we'll need two import statements from the same the module. One for type and one for the component like below.

```
import {Checkbox} from '@material-ui/core';
import type {CheckboxProps} from '@material-ui/core';
```

ESlint might complain at you for duplicate types. If it does, I found this github issue with the solution https://github.com/eslint/eslint/issues/13669#issuecomment-690085156

Would you be able to comb all the files again and split up type and non-types imports? And if ESLint complains then include the above solution in your PR as well?

20BBrown14 (changes requested):

src/components/SQForm/SQFormCheckbox.tsx
aharpt: @20BBrown14 Will do this.  Thanks for the feedback.

aharpt:
Do NOT merge this pull request at this time.  I have some bug that is causing the DateTimePickerwithCalendarInput to break.  I will update when this issue has been resolved.

SeanGroff:
> Do NOT merge this pull request at this time. I have some bug that is causing the DateTimePickerwithCalendarInput to break. I will update when this issue has been resolved.

When a PR is not ready to merge you can convert the PR to a draft

aharpt:
PR can now be merged.  There is an issue with the SQFormDatePickerWithCalendarOnlyInput component but we have confirmed it has nothing to do with my changes.

SeanGroff (approved):

PR #516, opened 2022-01-07
refactor: 💡 SQFormDialogStepper to TypeScript

aharpt:
Migrate SQFormDialogStepper to TypeScript

Loom video showing storybook working and detailing certain changes: https://www.loom.com/share/d58910b24ab74c7dacf59ae08c28ec2c

✅ Closes: #506


src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
```diff
@@ -12,66 +11,119 @@ import {
...
+  /** The current open/closed state of the Dialog */
+  isOpen: boolean;
+  /** Allows the initial values to be updated after initial render */
+  enableReinitialize?: boolean;
+  /** Determine the max-width of the dialog. The dialog width grows with the size of the screen. Set to false to disable maxWidth. */
+  maxWidth?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | false;
+  /** Callback function invoked when the user clicks on the secondary button or outside the Dialog */
+  onClose: () => void;
+  /** Callback function invoked when user clicks primary submit button */
+  onSubmit: (values: Record<string, unknown>, helpers: unknown) => void;
```
20BBrown14, line 112: I think we want to use generics for this component like we did with SQFormDialog. So that the form values will equal exactly what they're supposed to.

I do think we might need to get a little furth and have the generic `extend Record<string, unknown>` so we can index it properly in our comopnents maybe? That may not be necessary, however.

using a generic would affect the `onSubmit` function, `initialValues`, and looks like `setValues`

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
```diff
@@ -12,66 +11,119 @@ import {
...
+  /** Form Entity Object */
+  initialValues: Record<string, string>;
+  /** Callback function that is called when a step is completed to pass back the current state values to the consumer */
+  setValues?: (values: Record<string, unknown>) => void;
+  /** Any prop from https://material-ui.com/api/grid */
+  muiGridProps?: GridProps;
+  /** Any prop from https://material-ui.com/api/dialog/#props */
+  dialogProps?: DialogProps;
+  /** Optional styling on the dialog */
+  contentStyle?: Record<string, unknown>;
```
20BBrown14, line 124: We can likely get the type for this from `DialogContentProps['style']` to ensure we are only giving keys that are actual styles and have proper values

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
```diff
@@ -84,17 +136,17 @@ export function SQFormDialogStepper({
   title,
   enableReinitialize = false,
   muiGridProps = {},
-  dialogProps = {},
+  dialogProps = {open: isOpen},
```
20BBrown14, line 139: What was the need for this change?

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
```diff
@@ -113,14 +165,17 @@ export function SQFormDialogStepper({
     setCompleted([...completed, activeStep]);
   };
 
-  const handleStep = step => () => {
+  const handleStep = (step: number) => () => {
     const nextStep = step;
-    if ([nextStep].some(step => completed.includes(step))) {
+    if ([nextStep].some((step) => completed.includes(step))) {
```
20BBrown14, line 170: Im confused by this. wouldn't this just be the same as `if(completed.includes(step))`?

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
```diff
@@ -113,14 +165,17 @@ export function SQFormDialogStepper({
...
       setActiveStep(step);
     }
   };
 
-  const handleSubmit = async (values, helpers) => {
+  const handleSubmit = async (
+    values: Record<string, unknown>,
+    helpers: unknown
```
20BBrown14, line 177: The type for these helps we can get from formik. `FormikBag` is the type and it'll take two generics. The first will likely just be `unknown` and the second will be the generic this component takes in indicating the values of the form mentioned in a previous comment.

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
```diff
@@ -180,7 +235,6 @@ export function SQFormDialogStepper({
           TransitionComponent={Transition}
           disableBackdropClick={disableBackdropClick}
           maxWidth={maxWidth}
-          open={isOpen}
```
20BBrown14, line 183: Why remove this?

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
```diff
@@ -192,13 +246,13 @@ export function SQFormDialogStepper({
             <Divider />
             {steps.length > 1 && (
               <Grid container classes={stepperClasses}>
-                <Stepper activeStep={activeStep} classes={classes}>
+                <Stepper
+                  activeStep={activeStep}
+                  classes={classes}
+                  alternativeLabel
```
20BBrown14, line 252: nit: can we do `alternativeLabel={true}` here. It's the preferred way to pass a boolean as prop

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
```diff
@@ -192,13 +246,13 @@ export function SQFormDialogStepper({
...
+                >
                   {steps.map((child, index) => (
-                    <Step
-                      key={`${child.props.label}-${index}`}
-                      classes={classes.label}
-                      alternativeLabel
-                    >
+                    <Step key={`${child.props.label}-${index}`}>
```
20BBrown14, line 255: Can you explain why the other two props were removed from this component?

20BBrown14 (changes requested):

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
20BBrown14: I see why you decided to make this change. I have an issue with it, though. You removed passing `open` to the component below. What this means is that if the consumer of this component passes in a value for `dialogProps` it'll overwrite the default and as such they won't have `open`. However, TS won't let them do that if TS is being used to consume this library. So it'd be confusing for the user to pass `open` as a prop and in this dialogProps prop.

If the consumer is using JS alone, no TS, then they can pass whatever they like here and they might not get `open`.

With that being said, I think we should remove the default from this prop, pass `isOpen` to the component that needs it directly. And do a check before spread `dialogProps`. Say `<Component ...(dialogProps || {})>`

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
aharpt: I can look into it.  I actually did not modify this at all.  The parentheses were added to the function parameter to adhere to our current prettier config.

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
aharpt: Makes sense @20BBrown14.  Will change this. Thanks.

aharpt: I guess we won't be able to use `DialogProps` then to type the dialogProps since it requires the `open` property.

src/components/SQFormDialogStepper/SQFormDialogStepper.tsx
```diff
@@ -113,14 +170,16 @@ export function SQFormDialogStepper({
     setCompleted([...completed, activeStep]);
   };
 
-  const handleStep = step => () => {
-    const nextStep = step;
-    if ([nextStep].some(step => completed.includes(step))) {
+  const handleStep = (step: number) => () => {
```
aharpt, line 173: This extra code seems completely unnecessary to me.

20BBrown14 (approved):

lucas-homer (approved):

SeanGroff (approved):

PR #536, opened 2022-01-13
refactor: 💡 Move SQFormReadonly stories and tests to typescript

hvilander:
Moves stories and tests for read only field to typescript.


stories/SQFormReadOnlyField.stories.tsx
```diff
@@ -1,12 +1,22 @@
 import React from 'react';
+import {Story, Meta} from '@storybook/react';
 
 import {SQFormReadOnlyField as SQFormReadOnlyFieldComponent} from '../src';
+import type {SQFormReadOnlyFieldProps} from 'components/SQForm/SQFormReadOnlyField';
 import {SQFormStoryWrapper} from './components/SQFormStoryWrapper';
 import {createDocsPage} from './utils/createDocsPage';
 import getSizeProp from './utils/getSizeProp';
-import markdown from '../notes/SQFormReadOnlyField.md';
+import type {gridOptions} from './utils/getSizeProp';
```
hvilander, line 9: Is there any particular reason why this type is not capitalized?
I am happy to update it here.

SeanGroff: > Is there any particular reason why this type is not capitalized? I am happy to update it here.

Yea that's be nice. It's also a little odd importing a `type` called `gridOptions` from a utils folder named `getSizeProp`. I don't think this type would be very discoverable.

Might be worth moving to a `types` folder

20BBrown14: I believe we made in the _very early_ TS adoption process. We were just babies back then 🍼

stories/SQFormReadOnlyField.stories.tsx
```diff
@@ -46,3 +56,5 @@ WithInitialValue.args = {
 WithInitialValue.parameters = {
   controls: {exclude: 'initialValue'},
 };
+
+export default meta;
```
SeanGroff, line 60: There should be a newline at the end of the file

20BBrown14 (approved):

stories/SQFormReadOnlyField.stories.tsx
hvilander: :shakes fist at vscode:

src/components/SQForm/SQFormRadioButtonGroup.tsx
```diff
@@ -56,8 +56,6 @@ function SQFormRadioButtonGroup({
     onChange,
   });
 
-  console.log('field', field);
-
```
hvilander, line 60: Found this stray log. Removed it.

PR #537, opened 2022-01-13
refactor: 💡 Migrate SQForm Stories to TypeScript

aharpt:
Migrate SQForm Stories to TS

Loom video details the reasoning behind some of my changes.  Please watch.  

https://www.loom.com/share/28f28f77f2cd4069b796c887c0a5a65e

✅ Closes: #530

### DIFF
--- a/src/components/SQForm/SQFormReadOnlyField.tsx
+++ b/src/components/SQForm/SQFormReadOnlyField.tsx
@@ -8,8 +8,8 @@ import {toKebabCase} from '../../utils';
 
 export interface SQFormReadOnlyFieldProps extends BaseFieldProps {
   muiFieldProps?: TextFieldProps;
-  InputProps: TextFieldProps['InputProps'];
-  inputProps: TextFieldProps['inputProps'];
+  InputProps?: TextFieldProps['InputProps'];
+  inputProps?: TextFieldProps['inputProps'];
 }
 
 function SQFormReadOnlyField({
--- a/stories/SQForm.stories.tsx
+++ b/stories/SQForm.stories.tsx
@@ -401,9 +401,7 @@ export const formWithInclusionlist = (): JSX.Element => {
             }
           }
         >
-          {(
-            arrayHelpers: FieldArrayRenderProps
-          ): JSX.Element | Array<JSX.Element> => {
+          {(arrayHelpers: FieldArrayRenderProps): JSX.Element => {
             const {values} = arrayHelpers.form;
             return (
               <Grid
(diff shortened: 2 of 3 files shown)


src/components/SQForm/SQFormReadOnlyField.tsx
```diff
@@ -8,8 +8,8 @@ import {toKebabCase} from '../../utils';
 
 export interface SQFormReadOnlyFieldProps extends BaseFieldProps {
   muiFieldProps?: TextFieldProps;
-  InputProps: TextFieldProps['InputProps'];
-  inputProps: TextFieldProps['inputProps'];
+  InputProps?: TextFieldProps['InputProps'];
+  inputProps?: TextFieldProps['inputProps'];
```
aharpt, line 12: Forgot to mention something about this changes in the video.  There are multiple instances in `SQForm.stories.tsx` that we render a `SQFormReadOnlyField `without this two props defined.  I figured instead of going everywhere and adding in `InputProps={undefined}` or something, I would just make them optional.

stories/SQForm.stories.tsx
```diff
@@ -478,15 +481,15 @@ export const basicFormWithMultiSelect = () => {
         >
           {MOCK_FRIENDS_OPTIONS}
         </SQFormMultiSelect>
-        <Grid item size={2} style={{alignSelf: 'flex-end'}}>
+        <Grid item style={{alignSelf: 'flex-end'}}>
```
aharpt, line 484: Also wanted to mention this.  `size` is not a prop on `Grid` so this was literally doing nothing.  There are more instances of this below.

stories/SQForm.stories.tsx
```diff
@@ -400,7 +401,9 @@ export const formWithInclusionlist = () => {
             }
           }
         >
-          {(arrayHelpers) => {
+          {(
+            arrayHelpers: FieldArrayRenderProps
+          ): JSX.Element | Array<JSX.Element> => {
```
20BBrown14, line 406: This function is actually always returning just `JSX.Element`, right? One element and it's a grid.

20BBrown14 (commented):
Fantastic loom! I appreciate that you went through why some of the changes were made. 

I also looked into the `Yup.lazy as never` thing and couldn't find anything more than you could. Seems like it might just be a bug in how Yup is creatin those types. We should definitely be able to use it the way we are without casting or TS errors. The example in the docs even does it the way we are doing it. https://github.com/jquense/yup/tree/82979729a0b7cc6df5ac3540522fb79f6260ee1b#lazyvalue-any--schema-lazy

stories/SQForm.stories.tsx
lucas-homer: +1

lucas-homer (approved):
small technicality re: that inclusion list return type, but lgtm!

stories/SQForm.stories.tsx
aharpt: Correct.  Just changed it. Thanks.

SeanGroff (approved):

PR #546, opened 2022-01-14
Issue/346 migrate SQForm to TypeScript

aharpt:
Migrate SQForm.js to TypeScript

https://www.loom.com/share/778c7f418c0947b1a4b8a0010d56043f

✅ Closes: #346


src/components/SQForm/SQForm.tsx
```diff
@@ -1,25 +1,55 @@
...
 });
 
-function SQForm({
+interface SQFormProps<Values> {
+  /** Form Input(s) */
+  children: React.ReactNode;
+  /** Bool to pass through to Formik. https://formik.org/docs/api/formik#enablereinitialize-boolean */
+  enableReinitialize?: boolean;
+  /** Form Entity Object */
+  initialValues: Record<string, Values>;
```
20BBrown14, line 23: The type here would just be `Values` right?  I think we might want to do `Values extends Record<string, unknown>` but unsure. `Values` will definitely need to be a type like `{someKey: string, anotherKey: number}` where the keys are names of fields in the form. Basically initialValues. We can enforce that with `extends`

src/components/SQForm/SQForm.tsx
```diff
@@ -1,25 +1,55 @@
...
+  muiGridProps?: GridProps;
+  /**
+   * Form Submission Handler | @typedef onSubmit: (values: Values, formikBag: FormikBag) => void | Promise<any>
+   * IMPORTANT: If onSubmit is async, then Formik will automatically set isSubmitting to false on your behalf once it has resolved.
+   * This means you do NOT need to call formikBag.setSubmitting(false) manually.
+   * However, if your onSubmit function is synchronous, then you need to call setSubmitting(false) on your own.
+   *
+   * https://jaredpalmer.com/formik/docs/api/withFormik#handlesubmit-values-values-formikbag-formikbag--void--promiseany
+   * */
+  onSubmit: (
```
20BBrown14, line 34: I think we should change `onSubmit` to be optional with this TS conversion. We have the use case where SQForm is used for read-only forms and thus `onSubmit` isn't used. I'd be interested to hear thoughts on this, though.

src/components/SQForm/SQForm.tsx
aharpt: Let me look into this.  I was having some issues with the stories not liking `Values` here since it could not guarantee what it was passing was of type `Values` but let me investigate further.

20BBrown14: I could see a case where we're trying to index it or use it like an object and TS can't be guaranteed that `Values` is a type that allows for indexing or being used like an object. If oyu do `Value extends Record<string, unknown>` then TS won't allow a generic to be passed that isn't an object. I'd be interested to know what the issue is if it isn't that.

src/components/SQForm/SQForm.tsx
aharpt: Yeah I'm good with that as well now that I think about it.  I have personally had multiple times I have passed `onSubmit` an empty function `(() => {})` just to make `SQForm` happy.

src/components/SQForm/SQForm.tsx
aharpt: I'll try that and see where I can get.

aharpt: @20BBrown14 After trying to get this to work using your idea, I am still having errors, and am not sure how to resolve them using your suggested implementation.  Here is a loom that shows the issue I am running into.  If you or anyone has ideas of how to get around this let me know, thanks.
https://www.loom.com/share/41e6d770d38d4dd489eb5382a2a88ec4.

src/components/SQForm/SQForm.tsx
aharpt: I am curious for additional thoughts on this.  It seems to me that it is legitimate to make the `onSubmit` prop optional, but would like some additional opinions before I make this change.

src/components/SQForm/SQForm.tsx
20BBrown14: @aharpt thanks for giving it a go. After look at your loom I tried some stuff out and landed on this solution that I've described in this loom https://www.loom.com/share/fbee45c8f27446679bf05b9d5151b738

aharpt: @20BBrown14 Very cool!  I have made the changes.  Thanks for the help!

src/components/SQForm/SQForm.tsx
caleboleary: Can confirm I've had to send in noops to onSubmit several times before and agree making it optional makes sense.

aharpt: Sounds good @caleboleary.  I will make the change.  Thanks for the input.

src/components/SQForm/SQForm.tsx
```diff
@@ -34,17 +64,25 @@ function SQForm({
   // HACK: This is a workaround for: https://github.com/mui-org/material-ui-pickers/issues/2112
   // Remove this reset handler when the issue is fixed.
   const handleReset = () => {
-    document && document.activeElement && document.activeElement.blur();
+    document &&
+      document.activeElement &&
+      (document.activeElement as HTMLElement).blur();
```
hvilander, line 69: [tiny nit] Not sure if prettier or what is going on here, but the indentation is funny.

hvilander (approved):

caleboleary (approved):

src/components/SQForm/SQForm.tsx
```diff
@@ -34,17 +64,25 @@ function SQForm({
...
   };
 
   const handleSubmit = useDebouncedCallback(
-    (...args) => onSubmit(...args),
+    (values: Values, formikHelpers: FormikHelpers<Values>) => {
+      if (typeof onSubmit === 'undefined') {
```
20BBrown14, line 74: If we're keeping onSubmit being optional we could potentially do `onSubmit && onSubmit(values, formikHelpers)`, right?

src/components/SQForm/SQForm.tsx
```diff
@@ -1,25 +1,55 @@
...
+  muiGridProps?: GridProps;
+  /**
+   * Form Submission Handler | @typedef onSubmit: (values: Values, formikBag: FormikBag) => void | Promise<any>
+   * IMPORTANT: If onSubmit is async, then Formik will automatically set isSubmitting to false on your behalf once it has resolved.
+   * This means you do NOT need to call formikBag.setSubmitting(false) manually.
+   * However, if your onSubmit function is synchronous, then you need to call setSubmitting(false) on your own.
+   *
+   * https://jaredpalmer.com/formik/docs/api/withFormik#handlesubmit-values-values-formikbag-formikbag--void--promiseany
+   * */
+  onSubmit?: (
```
SeanGroff, line 34: I know my JS Doc comment above specifies the type. Instead of manually typing `onSubmit` can we use the one from Formik's types?

src/components/SQForm/SQForm.tsx
SeanGroff: I disagree with making `onSubmit` optional. We are promoting a use of this Form component it's not really meant for. 

Formik itself makes the `onSubmit` required as well:

https://github.com/jaredpalmer/formik/blob/e677bea8181f40e6762fc7e7fb009122384500c6/packages/formik/src/types.tsx#L215

To fill the hole in our library we need to build the actual feature that would cause a use to use SQForm without supplying `onSubmit`.

https://github.com/SelectQuoteLabs/SQForm/issues/561

aharpt: Ok, sounds good @SeanGroff.  I will revert this change since this will be addressed in #561.

src/components/SQForm/SQForm.tsx
aharpt: @SeanGroff Are you saying to update the JS Doc comment?  From Fromik, the type for `onSubmit` is `(values: Values, formikHelpers: FormikHelpers<Values>) => void | Promise<any>` which is exactly what I am using except my return type is `void | Promise<unknown>` to avoid eslint yelling at my for an unexpected any.

SeanGroff: @aharpt Sorry for not being clear. So when I wrote that JS Doc comment I stole the Type from their TypeScript definition.
Does Formik expose an `onSubmit` type that we could use instead of mirror their return type?

Example:

```typescript
`onSubmit: FormikOnSubmit` // or whatever the name of their type is for their onSubmit
```

20BBrown14: I think
```
import type {FormikConfig} form 'formik'
...
onSubmit: FormikConfig<Values>['onSubmit']
```

Would be the better option that way we can get our generics in there and ensure we have type safety on our `onSubmit` params

aharpt: Will try that.  Thanks @20BBrown14 and @SeanGroff.

SeanGroff: Looks great, thank you @aharpt

SeanGroff (approved):

Issue #554, opened 2022-01-18
SQFormStepper requires every field to be filled to enable Next/Submit button


BabarKhanSQ:
While working with SQFormStepper, I came across an issue where the next button was not getting enabled even all the **required** fields are filled. The stepper required every field to be filled in order to proceed further. 
Here are my few findings regarding the issue: https://www.loom.com/share/347b61155d5f47cd8a4321467678e4f0

BabarKhanSQ:
:3 Well I worked on it a bit and found the solution as well.

https://www.loom.com/share/6c2580b175de4a24afdcdc995202f0ec

PR #569, opened 2022-01-18
fix: 🐛 modified selectValues logic for required fields only

BabarKhanSQ:
modified selectValues logic for required fields only      ✅ Closes: #554


SeanGroff (changes requested):
The CI failed due to the code not being formatted with Prettier

SeanGroff (approved):

Issue #584, opened 2022-01-20
SQFormDatePicker assigning null on field clearance causing application to break


BabarKhanSQ:
While working on applicaiton I figured out that SQFormDatePicker is assigning value null/undefined to the datepicker field which is causing the application to break. The optional operater did prevent the app from breaking but we were unable to bypass the check.

Here are my findings along with solution: https://www.loom.com/share/f49d9a04a13141b794c46403ea6abace

PR #585, opened 2022-01-20
fix: 🐛 Added ? operator and modified date clearance logic

BabarKhanSQ:
Added optional chaining operator with values[step].length and a
condition that will prevent from assigning null to date

✅ Closes: #584

### DIFF
--- a/src/components/SQForm/SQFormDatePicker.js
+++ b/src/components/SQForm/SQFormDatePicker.js
@@ -44,10 +44,10 @@ function SQFormDatePicker({
     if (date) {
       helpers.setValue(date);
       onChange && onChange(date);
-      return;
+    } else {
+      helpers.setValue('');
+      onChange && onChange('');
     }
-    helpers.setValue('');
-    onChange && onChange('');
   };
 
   const [isCalendarOpen, setIsCalendarOpen] = React.useState(false);
--- a/src/components/SQForm/SQFormDatePicker.js
+++ b/src/components/SQForm/SQFormDatePicker.js
@@ -41,8 +41,13 @@ function SQFormDatePicker({
   });
 
   const handleChange = (date) => {
-    helpers.setValue(date);
-    onChange && onChange(date);
+    if (date) {
+      helpers.setValue(date);
+      onChange && onChange(date);
+      return;
+    }
+    helpers.setValue('');
+    onChange && onChange('');
   };
 
   const [isCalendarOpen, setIsCalendarOpen] = React.useState(false);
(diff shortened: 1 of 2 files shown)


src/components/SQForm/SQFormDatePicker.js
```diff
@@ -41,8 +41,13 @@ function SQFormDatePicker({
   });
 
   const handleChange = (date) => {
-    helpers.setValue(date);
-    onChange && onChange(date);
+    if (date) {
+      helpers.setValue(date);
+      onChange && onChange(date);
+      return;
```
20BBrown14, line 47: I vote `if/else` here instead of a `return` in the `if`

src/components/SQForm/SQFormDatePicker.js
```diff
@@ -41,8 +41,13 @@ function SQFormDatePicker({
...
+    }
+    helpers.setValue('');
```
20BBrown14, line 49: DatePicker doesn't place nice with empty strings, apaprently. There's a comment on L64 that an empty string won't reset the DatePicker we have to use null. Are these changes breaking anything? And how is this different than what we started with? if you clear the date does it try to set null before? And now it's an empty string?

These changes might work for the guidedworkflow but do they work with other forms? Do all the stories where we use this component still operate as expected with validation and all?

20BBrown14 (changes requested):

src/components/SQForm/SQFormDatePicker.js
BabarKhanSQ: 1) Are these changes breaking anything => For now I have tested this component inside the SQForm repo only and it seems to be working fine for all the scenarios that I checked. So, I didn't come across any scenario where it is breaking anything, Made it required, then removed the required field, seems like working good here.

2) How is this different than we started with => Right now our condition in file "SQFormDialogStepper.js" LINE 145 used to be like 
```
const stepValues = currentStepKeys.every((step) => {
        return !!values[step]
        }
```

Which even when gets a null will declare it false and invalidate the schema anyway (Also a reason why it needed every field to be required)

Changing it to what we have right now  
```
 const stepValues = currentStepKeys.every((step) => {
        if (
          values[step]?.length === 0 &&
          validationSchema.fields[step]?.tests[0]?.OPTIONS?.name === 'required'
        ) {
          return false;
          }
```
 Added a validation check for it being required, Now when we pass null to it making it  `if(null.length==0)` which invalidate the process. By putting an optional chaining operator we are preventing the application from breaking but it still invalidate the process. By passing an Empty string will ensure that it will have a property of length which can be equal to zero and will proceed as expected.

3)  if you clear the date does it try to set null before => Yes it does.
4) Changes work with other forms => The stories that we have (With validation and loading stepper) It seems to be working fine.

BabarKhanSQ: Sorry I was over thinking it. Made necessary changes required. Tested and verified.

SeanGroff (approved):

PR #600, opened 2022-01-27
SSC-141 SSC-144 SSC-149 dialog spacing paddings

Chris-Boyle:
Updated the padding and colors of the RoundButton to meet UX designs.  

✅ Closes: #601 

image
image
image

### DIFF
--- a/.storybook/styles/root.css
+++ b/.storybook/styles/root.css
@@ -7,6 +7,7 @@
   --color-srtsGreen: #5a7e44;
   --color-teal: #017788;
   --color-lightTeal: #34a1b6;
+  --color-darkTeal: #004b5b;
   --color-frenchBleu: #0066ff;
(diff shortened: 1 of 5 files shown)


Chris-Boyle:
> Could you open an issue for tracking purposes and explain what sparked the need for these changes, please.

Added an issue #601 and put it in the PR description.

jamesgriffith9519 (approved):

20BBrown14 (approved):

20BBrown14:
In the future, I think something like this would be a bug fix commit. But a feature is fine, so long as it's not a breaking change.

Chris-Boyle:
> In the future, I think something like this would be a bug fix commit. But a feature is fine, so long as it's not a breaking change.

This is an enhancement to our current design.  I didn't think enhancements were bugs.  I thought they fell into the feature commit type more.

PR #611, opened 2022-02-02
refactor: 💡 Migrate SQFormGuidedWorkflow to TS

20BBrown14:
BREAKING CHANGE: 🧨 SQFormGuidedWorkflow prop 'actionButton' renamed to 'actions'. Removed
props 'isPanelExpanded' and 'expandPanel'

✅ Closes: #567


SeanGroff (changes requested):
Looks like the tests are failing.
As a side note, instead of running jest watch mode, give the Jest VS Code extension a try. It makes viewing the failing tests easier than trying to scroll through a terminal window

20BBrown14:
> Looks like the tests are failing. As a side note, instead of running jest watch mode, give the Jest VS Code extension a try. It makes viewing the failing tests easier than trying to scroll through a terminal window

Apparently I never learn my lesson about these darn tests. i always forget about them but I guess that's why we have a CI, right? To be fair, though, I'm not changing functionality so the tests shouldn't be failing.

SeanGroff (approved):

PR #613, opened 2022-02-02
refactor: 💡 Migrate LoadingSpinner stories to TS

aharpt:
Migrate LoadingSpinner stories to TypeScript

https://www.loom.com/share/e858a5d8768c4111a37af63f3048c552

✅ Closes: #612

### DIFF
--- a/stories/LoadingSpinner.stories.tsx
+++ b/stories/LoadingSpinner.stories.tsx
@@ -7,12 +7,15 @@ export default {
   component: LoadingSpinner,
   parameters: {
     docs: {
-      page: createDocsPage()
-    }
-  }
+      page: createDocsPage(),
+    },
+  },
 };
 
-export const Default = args => {
-  return <LoadingSpinner {...args} />;
+export const Default = (
+  args: Record<string, unknown> & {message?: string}
+): JSX.Element => {
+  const {message} = args;
+  return <LoadingSpinner message={message} />;
 };
 Default.storyName = 'LoadingSpinner';


stories/LoadingSpinner.stories.tsx
```diff
@@ -7,12 +7,15 @@ export default {
...
-  }
+      page: createDocsPage(),
+    },
+  },
 };
 
-export const Default = args => {
-  return <LoadingSpinner {...args} />;
+export const Default = (
+  args: Record<string, unknown> & {message?: string}
```
lucas-homer, line 16: maybe a nit, but looking at the LoadingSpinner proptypes, it's a string, so instead of making `message` optional, we could default the value to empty string (`{message: string = ""}`) otherwise looks good to me.

lucas-homer (approved):

20BBrown14 (approved):

stories/LoadingSpinner.stories.tsx
aharpt: Hmm, when I try to do that I get "A type literal property cannot have an initializer" error since I am essentially would be attempting to initialize the value of a type (I believe that is what the error is saying).  What I could do is render the `LoadingSpinner` component as
`<LoadingSpinner message={message ?? ""} />` to return an empty string if `message` is `undefined`.

lucas-homer: gosh, sorry... dunno what I was thinking there. yeah your solution is fine. also though the existing way would work too, so please disregard my original comment. more trouble than it's worth

lucas-homer (approved):

Issue #629, opened 2022-02-04
3rd footer button in SQFormDialog


Ariel-Schnur-SQ:
In AQE2 we have a use case for a 3rd footer button. For example, the Add Prescriptions modal requires buttons for Cancel, Save and Close (add prescription and close modal), and Add Another Rx (add prescription and clear form so agent can add another one). For now, we've copy/pasted the SQFormDialog code into our repo and added a button. An option for a 3rd button would be great!

Add Prescription ticket for reference:
https://selectquote.atlassian.net/browse/AQE2-692

Ariel-Schnur-SQ:
Also adding a screenshot for reference. We didn't have a mockup so I'll update this if need be after running it by the designers.

![Screen Shot 2022-02-04 at 2 07 53 PM](https://user-images.githubusercontent.com/98851673/152603271-80278392-9a66-42c5-90df-a704ae7fb95d.png)

PR #631, opened 2022-02-09
feat: 🎸 Add Third button to SQFormDialog

aharpt:
Added third button functionality to SQFormDialog

https://www.loom.com/share/0619324e66cd4f2ca80f9a034ff8dd89

✅ Closes: #629

### DIFF
--- a/src/components/SQFormDialog/SQFormDialogInner.js
+++ b/src/components/SQFormDialog/SQFormDialogInner.js
@@ -65,10 +65,13 @@ function SQFormDialogInner({
   onClose,
   onSave,
   saveButtonText,
+  tertiaryButtonText,
   shouldRequireFieldUpdates = false,
   title,
   muiGridProps,
   showSecondaryButton = true,
+  showTertiaryButton = false,
+  isTertiaryDisabled = false,
 }) {
   const theme = useTheme();
   const titleClasses = useTitleStyles(theme);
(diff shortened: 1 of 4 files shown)


stories/__tests__/SQFormDialog.stories.test.js
```diff
@@ -168,3 +168,40 @@ describe('Tests for WithAutoFocus', () => {
     expect(textfield).toBe(document.activeElement);
   });
 });
+
+describe('Tests for Tertiary Button', () => {
+  it('should find the tertiary button', async () => {
```
SeanGroff, line 173: We always want to speak in terms of what the user would see or experience, rather than the test itself.

'Should display the tertiary button'

src/components/SQFormDialog/SQFormDialogInner.js
```diff
@@ -124,17 +127,57 @@ function SQFormDialogInner({
               showSecondaryButton ? actionsClasses : primaryActionsClasses
             }
           >
-            {showSecondaryButton && (
-              <RoundedButton
-                title={cancelButtonText}
-                onClick={handleCancel}
-                color="secondary"
-                variant="outlined"
+            {showTertiaryButton ? (
```
SeanGroff, line 130: We can take a functional approach to avoid these conditionals.
We can create render functions that contain the logic, we can create separate components that handle this logic and except the necessary props, or we can live this as is.

SeanGroff (approved):

src/components/SQFormDialog/SQFormDialogInner.js
20BBrown14: I'm in the "Create render functions" camp on this one. I found it a little hard to follow what was going on in the diff.

stories/__tests__/SQFormDialog.stories.test.js
```diff
@@ -168,3 +168,40 @@ describe('Tests for WithAutoFocus', () => {
...
+        isOpen={true}
+        onSave={handleSave}
+        onClose={handleClose}
+        showTertiaryButton={true}
+        tertiaryButtonText={tertiaryButtonText}
+        title="Test"
+      />
+    );
+
+    await screen.findByText(tertiaryButtonText);
```
laurelbean, line 186: There are no expects in this test, so nothing is actually being tested. Probably want to add a check for if the button is visible and maybe with the expected button text.

aharpt: @laurelbean I added the tests, but if you could double-check them I would appreciate it.  Pretty new to writing tests...

laurelbean: It's looking better!

The 2 lines below are essentially doing the same thing, checking for the existence of the button. So you don't need both of them. Also, unless you're expecting to wait for loading or an animation, you don't need to `waitFor` an expect. 

```
expect(
   screen.getByRole('button', {name: tertiaryButtonText})
).toBeInTheDocument();

await waitFor(() => {
    expect(screen.getByText(tertiaryButtonText)).toBeInTheDocument();
});```

If you wanted to check for the text of the button, then I'd save the `getByRole` to a variable, and check that for existence and for text. Something like:
```
const tertiaryButton = screen.getByRole('button', {name: tertiaryButtonText});
expect(tertiaryButton).toBeInTheDocument();

expect(tertiaryButton).toHaveTextContext(tertiaryButtonText);
or 
expect(tertiaryButton).toHaveText(tertiaryButtonText);
```

aharpt: @laurelbean Thanks for helping me out with this!  The correct changes should now be part of this PR.  Let me know if anything further needs to be addressed.

20BBrown14 (approved):

Issue #627, opened 2022-02-03
Expose all component prop TS types to consumers


20BBrown14:
Nearing the end of our SQForm TS conversion we should ensure that we're exposing all of our component props types to any of the library's consumers much like Formik or MUI does.

We should figure out the best way to do this. Discussion can be had on this issue here.

PR #635, opened 2022-02-11
refactor: 💡 Rename poorly named exposed types

20BBrown14:
Renamed some of the types that were poorly named like "Props" or
"HeaderProps" to things that are more descriptive, less likely have to
collisions, and are more easily identifiable.

✅ Closes: #627

### DIFF
--- a/src/components/SQForm/SQFormButton.tsx
+++ b/src/components/SQForm/SQFormButton.tsx
@@ -3,7 +3,7 @@ import {RoundedButton} from 'scplus-shared-components';
 import {useFormButton, BUTTON_TYPES} from './useFormButton';
 import type {ButtonType} from './useFormButton';
 
-export interface Props {
+export interface SQFormButtonProps {
   children: React.ReactNode;
   isDisabled?: boolean;
   shouldRequireFieldUpdates?: boolean;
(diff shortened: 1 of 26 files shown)


src/components/SQForm/SQFormButton.tsx
```diff
@@ -19,7 +19,7 @@ function SQFormButton({
   title,
   type = 'submit',
   onClick,
-}: Props): JSX.Element {
+}: SQFormButtonProps): JSX.Element {
```
SeanGroff, line 22: Just want to call out in our consumer apps `Props` is sufficient as long as we stick to our rule of one component per file. It's not so sufficient here because we re-use the props in Storybook.

SeanGroff (approved):

src/components/SQForm/SQFormButton.tsx
20BBrown14: My main concern was that this type is exposed to consumers and simply have a type named "Props" isn't sufficient.

SeanGroff: Nice catch. That is indeed a very valid concern!

Issue #316, opened 2021-06-23
Change validationSchema to accept POJO OR Yup.object().shape()


SeanGroff:
With TypeScript and Clean Architecture, we can now leverage TypeScript and Yup to create a model we can share across the Backend and Frontend.

```javascript
// Example Model
export const ActionSchema = yup.object().shape({
  key: yup.number().positive().required(),
  title: yup.string().required(),
  CommentFlag: yup.boolean().required().default(false),
  AttemptFlag: yup.boolean().required().default(false),
  ContactFlag: yup.boolean().required().default(false),
  AddUser: yup.string(),
  AddDate: yup.date(),
  ChangeUser: yup.string(),
  ChangeDate: yup.date(),
  // etc
  });

// Generates a TypeScript type based on the Yup validation values in the schema
export type Action = yup.InferType<typeof ActionSchema>;
```

Then on the client, we can use the ActionSchema (or a Partial of it, with `Yup.reach()`) and use it as the `validationSchema` for the form.

We can use the `Action` type to add type safety to out `initialValues` object. Or create a Partial of the Action type based on the properties we need.

```javascript
import {ActionSchema} from '@models/actions/action'
import type {Action} from '@models/actions/action`

const intialValues: Action = {...}
<SQForm validationSchema={ActionSchema} initialValues={initialValues} onSubmit={noop)>...</SQForm>
```

This requires a breaking change as currently, `SQForm.validationSchema` accepts a POJO. 

The approach i've netted out on after some discussion is to avoid the breaking change for now by accepting either a POJO or `yup.object().shape()`.

If the user passes a Pojo, we display a "The ability to pass a plain old javascript object will be deprecated in a future MAJOR version upgrade. Please change the POJO to a Yup.object() to avoid" warning in the console.

In the future, force the standardization of using `yup.object().shape()` as the single way to use the `validationSchema` prop.

PR #648, opened 2022-03-22
refactor: 💡 SQForm components now only accept Yup schemas

20BBrown14:
SQForm components no only accept objects created with `Yup.object()` and
no longer accepts plain javascript objects

BREAKING CHANGE: 🧨 Plain javascript objects are no longer accepted for validation schema

Loom: https://www.loom.com/share/1459283335134f689166163ef83e75f8

✅ Closes: #316

### DIFF
--- a/src/components/SQForm/SQForm.tsx
+++ b/src/components/SQForm/SQForm.tsx
@@ -1,12 +1,12 @@
 import React from 'react';
 import Grid from '@material-ui/core/Grid';
+import {setLocale} from 'yup';
+import type {AnyObjectSchema} from 'yup';
 import {Formik, Form} from 'formik';
 import {useDebouncedCallback} from 'use-debounce';
-import {object as YupObject, setLocale} from 'yup';
 import {useInitialRequiredErrors} from '../../hooks/useInitialRequiredErrors';
 import type {GridProps} from '@material-ui/core';
-import type {FormikHelpers, FormikConfig} from 'formik';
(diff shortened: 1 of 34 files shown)


src/components/SQForm/SQForm.tsx
```diff
@@ -1,20 +1,20 @@
 import React from 'react';
 import Grid from '@material-ui/core/Grid';
+import {setLocale} from 'yup';
+import type {AnyObjectSchema} from 'yup';
```
SeanGroff, line 4: I need to install a TS imports sorter, but since I haven't, do you mind ensure `type` imports are last? (Unless there's a CSS file, then CSS files go last)

src/components/SQForm/SQForm.tsx
```diff
@@ -1,20 +1,20 @@
...
+import type {FormikHelpers, FormikConfig, FormikValues} from 'formik';
 
 setLocale({
   mixed: {
     required: 'Required',
   },
 });
 
-export interface SQFormProps<Values extends Record<string, unknown>> {
+export interface SQFormProps<Values extends FormikValues> {
```
SeanGroff, line 17: Nice work hardening this `type` as well as the `validationSchema` `type`

src/components/SQForm/SQForm.tsx
20BBrown14: Every file will need some attention. We can address that in #649 to avoid bloat in this already massive PR.

PR #651, opened 2022-03-29
Issue/650 no title in sq form scrollable card

hvilander:
https://www.loom.com/share/4d63cdf5b5134fa98565ea3d6e278943

I think this is really straightforward. We may have an additional issue with padding, but this resovles the core of the title not showing.


hvilander:
I looked at the padding on the current master branch deployed story book. It appears the same, which maybe is a chance for enhancement. I figure that is outside the scope of this quick fix.

20BBrown14 (approved):

PR #654, opened 2022-03-30
fix: 🐛 Padding was being hard coded by scplus theme

hvilander:
Due to some theme changes in scplus shared components the height of the header was fixed to 48px. Given that we expect different title variants into SQFormScrollable card. This modifies the css class applied to the header over the theme and gives the header 'auto' for height. This looks to restore the previous padding behavior. 

I have listed this as a breaking change as it does change how the layout will work. However, if this is incorrect I am happy to modify / remake this PR. 

image

image

Here is a loom of the issue when it is broken: https://www.loom.com/share/d051b256c278458cabf0eec76fbc6334


laurelbean (approved):

SeanGroff (changes requested):
To no fault of @hvilander's, requirements have changed/become more clear.
We've discussed the path forward over Slack.

Marking this review with "Request Changes" only so someone doesn't come in and merge this PR yet

SeanGroff:
I converted this PR to a Draft to signify this is a Work in Progress.
Once the changes are applied to this draft, we can convert it back to a PR

Issue #657, opened 2022-04-05
Remove isRequired as a prop from components


20BBrown14:
SQFormAutocomplete is still accepting `isRequired` as a prop and it shouldn't be. We removed that in a previous major version and this one got left behind

SQFormTextField, SQFormTextarea, SQFormMaskedTextField, SQFormDatePicker are also accepting this prop

https://github.com/SelectQuoteLabs/SQForm/blob/issue/244-EnableTypeScript/src/components/SQForm/SQFormAutocomplete.tsx#L24

We should also take this time to make sure we've removed all instances of it.

PR #662, opened 2022-04-06
refactor: 💡 Removes isRequired which has become dead code

SeanGroff:
✅ Closes: #657


20BBrown14 (approved):
So many more leftover than I thought 😓

Issue #561, opened 2022-01-18
SQFormReadOnly Component


SeanGroff:
We commonly have the need in our applications to render Read Only Forms. 
An example would be in the Senior CRM, the Personal Info form is a Read Only form. To edit the fields the agent needs to click the Edit button to bring up a seperate edit form dialog.

To make this work, we have to stub in unnecessary config to SQForm such as initialValues and onSubmit.
This is a gap in our library, we should make a solution where the user has no config other than transforming the data to the expected shape to render the proper fields with Labels and Values.

```
<SQFormReadOnly>{readOnlyFields}<SQFormReadOnly>
```

Since there is no validation on a read-only form we can also compact the vertical spacing between rows of inputs. Currently there's a `21px` bottom margin applied to all form inputs to accommodate the validation message that appears beneath the input.
This bottom margin is not needed for a read-only form and should be overwritten so the rows are more compact, to show more info on the screen.

PR #665, opened 2022-04-08
feat: 🎸 Read only form component

SeanGroff:
✅ Closes: #561

Note: The Birthday field is a Masked Input

Screen Shot 2022-04-11 at 8 52 33 AM

### DIFF
--- a/src/components/SQForm/SQFormReadOnly.tsx
+++ b/src/components/SQForm/SQFormReadOnly.tsx
@@ -2,15 +2,16 @@ import React from 'react';
 import Grid from '@material-ui/core/Grid';
 import {Formik, Form} from 'formik';
 import SQFormReadOnlyField from './SQFormReadOnlyField';
+import SQFormMaskedReadOnlyField from './SQFormMaskedReadOnlyField';
 import {noop} from '../../utils';
 import type {CSSProperties} from 'react';
 import type {GridProps} from '@material-ui/core';
 import type {FormikValues} from 'formik';
-import type {SQFormReadOnlyFieldProps} from './SQFormReadOnlyField';
+import type {SQFormMaskedReadOnlyFieldProps} from './SQFormMaskedReadOnlyField';
 
 export interface SQFormReadOnlyProps<Values extends FormikValues> {
   /** Form Input(s) */
-  readOnlyFields: SQFormReadOnlyFieldProps[];
+  readOnlyFields: SQFormMaskedReadOnlyFieldProps[];
   /** Form Entity Object */
   initialValues: Values;
   /** Any prop from https://material-ui.com/api/grid */
@@ -38,13 +39,16 @@ function SQFormReadOnly<Values extends FormikValues>({
               spacing={muiGridProps.spacing ?? 2}
             >
               {readOnlyFields.map((readOnlyField, index) => {
-                return (
-                  <SQFormReadOnlyField
-                    key={index}
-                    muiFieldProps={noBottomMarginStyle}
-                    {...readOnlyField}
-                  />
-                );
+                const props = {
+                  key: index,
+                  muiFieldProps: noBottomMarginStyle,
+                  ...readOnlyField,
+                };
+
+                if (readOnlyField?.mask) {
+                  return <SQFormMaskedReadOnlyField {...props} />;
+                }
+                return <SQFormReadOnlyField {...props} />;
               })}
             </Grid>
           </Form>
--- a/src/components/SQForm/SQFormReadOnly.tsx
+++ b/src/components/SQForm/SQFormReadOnly.tsx
@@ -1,0 +1,57 @@
+import React from 'react';
+import Grid from '@material-ui/core/Grid';
+import {Formik, Form} from 'formik';
+import SQFormReadOnlyField from './SQFormReadOnlyField';
+import {noop} from 'utils';
+import type {CSSProperties} from 'react';
+import type {GridProps} from '@material-ui/core';
+import type {FormikValues} from 'formik';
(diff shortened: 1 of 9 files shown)


src/components/SQForm/SQFormReadOnly.tsx
```diff
@@ -0,0 +1,57 @@
...
+        return (
+          <Form>
+            <Grid
+              {...muiGridProps}
+              container
+              spacing={muiGridProps.spacing ?? 2}
+            >
+              {readOnlyFields.map((readOnlyField, index) => {
+                return (
+                  <SQFormReadOnlyField
```
20BBrown14, line 42: So, thoughts on incorporating the masked read only field? Could likely do it similar but have `readOnlyFields` typed as `SQFormMaskedReadOnlyFieldProps[]` since it extends SQFormReadOnlyFieldProps _and_ since `mask` is optional it should work out.

Then in this component if the mask is provided render a masked field and give it the mask. otherwise, stick with the normal read only field.

20BBrown14 (changes requested):

src/components/SQForm/SQFormReadOnly.tsx
SeanGroff: Good idea 💡 
I didn't think about the masked component.
I'll change the code, then add another Story example and some tests to cover the Mask component

src/components/SQForm/SQFormReadOnly.tsx
```diff
@@ -0,0 +1,61 @@
...
+        return (
+          <Form>
+            <Grid
+              {...muiGridProps}
+              container
+              spacing={muiGridProps.spacing ?? 2}
+            >
+              {readOnlyFields.map((readOnlyField, index) => {
+                const props = {
+                  key: index,
```
20BBrown14, line 43: In general we should avoid using array indexes alone for element keys. Perhaps we could use the name of the field passed in? Another option would be to require the consumer to pass it in by typing `readOnlyField` as `SQFormMaskedReadOnlyFieldProps[] & { key: string}`

20BBrown14 (changes requested):

src/components/SQForm/SQFormReadOnly.tsx
SeanGroff: We've tried using `name` for the `key` in the past but it quickly became a problem as name from our data was frequently not unique. 

In general we should use array indexes as a last resort. This is a read only form so we can't manipulate, delete, shift around the order of the fields so there shouldn't be any ill effect of using `index`.

If we require the type in the TS `type` the consumer is likely just going to use the array index. A complex key doesn't move the bar forward much either, as `name` and `label` can both be repeated as we've seen on the Admin views project.

Using a date time stamp for each and every key is really expensive.

In this case, I think `index` as the last resort is viable

https://reactjs.org/docs/lists-and-keys.html#keys

> We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state. Check out Robin Pokorny’s article for an [in-depth explanation on the negative impacts of using an index as a key](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318). If you choose not to assign an explicit key to list items then React will default to using indexes as keys.

From the above link in the React docs:

"Update: Exception from the rule
Many people asked if they always, always have to generate ids. Others have suggested use cases when using the index as a key seems justifiable.
It is true that sometimes generating new ids is redundant and may be avoided. For example translation of license terms or list of contributors.
To help you decide, I put together three conditions which these examples have in common:
the list and items are static–they are not computed and do not change;
the items in the list have no ids;
the list is never reordered or filtered.
When all of them are met, you may safely use the index as a key."

20BBrown14: I'd argue that as a library, we can't guarantee that all of those conditions will be met. A consumer can definitively make that assumption for their use case, however.

20BBrown14: To provide a simple example of when one of these conditions owuld not be met, and a non-contrived example, would be when loading the read-only data from an endpoint. Initially the value of the fields are likely an empty string but then when the data is loaded it's changed.

Similarly, a form that displays read-only data based on the selection of another element in the UI.

SeanGroff: > To provide a simple example of when one of these conditions owuld not be met, and a non-contrived example, would be when loading the read-only data from an endpoint. Initially the value of the fields are likely an empty string but then when the data is loaded it's changed.
> 
> Similarly, a form that displays read-only data based on the selection of another element in the UI.

I'm happy to add `index` to the `type` so that the consumer can make the choice of whether or not to use `index` rather than the library.

The 2 real world examples shouldn't be affected by using `index` as the key. If the value of the props need to change in order to reflect new values in the form then we set `enableReinitialize` within the formik form.

20BBrown14: I think a happy medium here would be to allow the consumer to pass an optional key and have consumer provided keys take precedence over whatever the library decides to use. 
You're likely right in that we'd never see an issue but the ones that would crop out would be quite the doozies to try to debug. This way consumers could at least alleviate one potential issue. Plus, if they need to identify the elements in the DOM for whatever reason they can use their own keys instead.

Chris-Boyle (approved):

src/components/SQForm/SQFormReadOnly.tsx
SeanGroff: Developer is now able to specify React key's
cc: @20BBrown14

20BBrown14 (approved):

Issue #672, opened 2022-04-15
SQForm V8: `SQFormDatePicker` muiFieldProps does not take an `onAccept` prop


aharpt:
I am working on SQForm V8 migration on the AQE2 project.  I ran into this issue with the muiFieldProps in the. SQFormDatePicker component.  See below video.

https://www.loom.com/share/e362fd34b3a94a709dcf65f547af34d0

PR #673, opened 2022-04-15
Issue 672/fix date picker type

aharpt:
Fix hole in `BaseDatePickerProps`.

https://www.loom.com/share/9aa352088e9f4702af3a9db90a8c78b3

✅ Closes: #672

### DIFF
--- a/src/components/SQForm/SQFormDatePicker.tsx
+++ b/src/components/SQForm/SQFormDatePicker.tsx
@@ -1,13 +1,13 @@
-import React from 'react';
-import type {Moment} from 'moment';
+import type {BaseFieldProps} from '../../types';
+import {useForm} from './useForm';
+import {makeStyles, ClickAwayListener} from '@material-ui/core';
+import type {InputBaseComponentProps} from '@material-ui/core';
 import Grid from '@material-ui/core/Grid';
 import TextField from '@material-ui/core/TextField';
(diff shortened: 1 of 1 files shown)


SeanGroff:
Can we squash the commits (or whatever method works best for you) to get rid of the commit from your other PR?

aharpt:
> Can we squash the commits (or whatever method works best for you) to get rid of the commit from your other PR?

@SeanGroff The import commit actually has to do with an inadvertent import order change in `SQFormDatePicker.tsx` which I did not want to appear in the diff.

SeanGroff:
After looking through their source code `types` I think we can extend their `BasePickerProps` instead of manually adding one prop so this doesn't happen again with other holes in the current type.

https://github.com/mui/material-ui-pickers/blob/e875729b3492a4aedebec9f14589accc83475241/lib/src/typings/BasePicker.tsx

aharpt:
> After looking through their source code `types` I think we can extend their `BasePickerProps` instead of manually adding one prop so this doesn't happen again with other holes in the current type.
> 
> https://github.com/mui/material-ui-pickers/blob/e875729b3492a4aedebec9f14589accc83475241/lib/src/typings/BasePicker.tsx

@SeanGroff PR is updated to reflect this change.

hvilander (approved):

SeanGroff (approved):

Issue #677, opened 2022-05-02
SQFormDialog tertiary button enhancement


jamesgriffith9519:
The SQFormDialog's optional tertiary button is in need of an enhancement that will allow the user access to formik context. 

loom with more context: https://www.loom.com/share/9fb1df3fac334499aca7827675d49b0b

1. The text in the tertiary button is….long. Something like “Add Another” would make the button look less like a hot dog lol but ask UI/UX for their preference in wording

2. Create an issue to enhance the tertiary button in this file https://github.com/SelectQuoteLabs/SQForm/blob/master/src/components/SQFormDialog/SQFormDialogInner.tsx

3. We have access to everything we need from Formik Context as needed. See line 132

4. isDisabled={isTertiaryDisabled} could be updated to use the isValid? (Don’t remember the exact prop name) prop from formik context to be isDisabled={isTertiaryDisabled || !isValid} so we can leverage the form validation if we want to.

5. onClick={onTertiaryClick} We NEED to enhance this to give the consumer access to the formik context. Something like onClick={() => onTertiaryClick(formikContext)}

SeanGroff:
Thanks @jamesgriffith9519 !
Do you mind including the guidance I provided in Slack with this issue just so the future PR reviewers will have the full context of our Acceptance Criteria? 

Copy/Paste is totally fine lol

PR #678, opened 2022-05-02
feat: 🎸 update onTertiaryClick prop to return context

jamesgriffith9519:
✅ Closes: #677

https://www.loom.com/share/a4450b3223434ea3bfa72f62f0aecb0e

### DIFF
--- a/src/components/SQFormDialog/SQFormDialog.tsx
+++ b/src/components/SQFormDialog/SQFormDialog.tsx
@@ -57,7 +57,7 @@ export interface SQFormDialogProps<Values extends FormikValues> {
    * */
   validationSchema?: AnyObjectSchema;
   /** Callback function invoked when the user clicks the tertiary button */
-  onTertiaryClick?: (formikContext: FormikContextType<unknown>) => void;
+  onTertiaryClick?: (formikContext: FormikContextType<Values>) => void;
 }
 
 function SQFormDialog<Values extends FormikValues>({
--- a/src/components/SQFormDialog/SQFormDialog.tsx
+++ b/src/components/SQFormDialog/SQFormDialog.tsx
@@ -5,6 +5,7 @@ import {useInitialRequiredErrors} from '../../hooks/useInitialRequiredErrors';
 import type {FormikHelpers, FormikValues, FormikContextType} from 'formik';
 import type {DialogProps, GridProps} from '@material-ui/core';
 import type {AnyObjectSchema} from 'yup';
+import type {SQFormDialogTertiaryValue} from './types';
 
 export interface SQFormDialogProps<Values extends FormikValues> {
   /** The secondary button text (Button located on left side of Dialog) */
@@ -31,12 +32,14 @@ export interface SQFormDialogProps<Values extends FormikValues> {
   ) => void | Promise<unknown>;
   /** Determine if the secondary action button should be displayed (default: true) */
   showSecondaryButton?: boolean;
-  /** Whether to show the tertiary button. (Default: false) */
-  showTertiaryButton?: boolean;
+  /**
+   * determine the status of the tertiary button
+   * can be one of'HIDE_BUTTON'|'NO_VALIDATION'|'IS_DISABLED'|'IS_ENABLED'|'FORM_VALIDATION_ONLY'|'IS_DISABLED_AND_FORM_VALIDATION',
+   * this will determine if the button is rendered, as well as if the button is disabled and if it uses form validation.
+   */
+  tertiaryStatus: SQFormDialogTertiaryValue;
   /** The tertiary button text */
   tertiaryButtonText?: string;
-  /** Whether the tertiary button is disabled (Default: false) */
-  isTertiaryDisabled?: boolean;
   /** Whether to show save/submit button (default: true) */
   shouldDisplaySaveButton?: boolean;
   /** The primary button text (Button located on right side of Dialog) */
@@ -79,9 +82,8 @@ function SQFormDialog<Values extends FormikValues>({
   shouldRequireFieldUpdates = false,
   validationSchema,
   showSecondaryButton = true,
-  showTertiaryButton = false,
-  isTertiaryDisabled = false,
   onTertiaryClick,
+  tertiaryStatus = 'HIDE_BUTTON',
 }: SQFormDialogProps<Values>): React.ReactElement {
   const initialErrors = useInitialRequiredErrors(
     validationSchema,
@@ -111,9 +113,8 @@ function SQFormDialog<Values extends FormikValues>({
         title={title}
         muiGridProps={muiGridProps}
         showSecondaryButton={showSecondaryButton}
-        showTertiaryButton={showTertiaryButton}
+        tertiaryStatus={tertiaryStatus}
         tertiaryButtonText={tertiaryButtonText}
-        isTertiaryDisabled={isTertiaryDisabled}
         onTertiaryClick={onTertiaryClick}
       />
     </Formik>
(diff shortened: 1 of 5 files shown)


src/components/SQFormDialog/SQFormDialogInner.tsx
```diff
@@ -55,7 +56,7 @@ interface SQFormDialogInnerProps {
   /** Whether the tertiary button is disabled (Default: false) */
   isTertiaryDisabled?: boolean;
   /** Callback function invoked when the user clicks the tertiary button */
-  onTertiaryClick?: React.MouseEventHandler<HTMLButtonElement>;
+  onTertiaryClick?: (formikContext: FormikContextType<unknown>) => void;
```
SeanGroff, line 59: Instead of `unknown` I think this would be `Values`?

jamesgriffith9519: Would this mean that we would have to use a generic much like the SQFormDialog file?

SeanGroff: > Would this mean that we would have to use a generic much like the SQFormDialog file?

Yep!

src/components/SQFormDialog/SQFormDialog.tsx
```diff
@@ -57,7 +57,7 @@ export interface SQFormDialogProps<Values extends FormikValues> {
    * */
   validationSchema?: AnyObjectSchema;
   /** Callback function invoked when the user clicks the tertiary button */
-  onTertiaryClick?: React.MouseEventHandler<HTMLButtonElement>;
+  onTertiaryClick?: (formikContext: FormikContextType<Values>) => void;
```
20BBrown14, line 60: Don't forget to pass the generic from `SQFormDialog` to `SQFormDialogInner` as well

jamesgriffith9519:
> Looks, great. Thanks @jamesgriffith9519!

Thanks for the help on this one!

src/components/SQFormDialog/SQFormDialog.tsx
jamesgriffith9519: ahh good call and thanks for explaining this a bit more!

src/components/SQFormDialog/SQFormDialogInner.tsx
```diff
@@ -183,8 +184,8 @@ function SQFormDialogInner({
           <span style={{paddingRight: '20px'}}>
             <SQFormButton
               title={tertiaryButtonText}
-              isDisabled={isTertiaryDisabled}
-              onClick={onTertiaryClick}
+              isDisabled={isTertiaryDisabled || !formikContext.isValid}
```
20BBrown14, line 187: What if the consumer intends for this 3rd button to be seperate from formikContext for whatever reason? Doesn't seem unreasonable to me. Perhaps I'm over thinking it? But maybe the third button opens another informational dialog? A bit hard to think of something not super contrived, to be honest. What do we think here?

src/components/SQFormDialog/SQFormDialogInner.tsx
```diff
@@ -183,8 +184,8 @@ function SQFormDialogInner({
...
+              onClick={() => onTertiaryClick?.(formikContext)}
```
hvilander, line 188: Maybe this is how we want it...
if as a consumer don't want the form validitiy to be considered and just manage it through isTertiaryDisabled... then false may do things I dont want.
Would it be better to maybe do `isTertiaryDisabled ?? !formikContext.isVaild` ?

just a thought, I think the common use case is a submit the form and keep a modal open for another so maybe we don't care?

stories/SQFormDialog.stories.tsx
```diff
@@ -143,14 +144,23 @@ export const WithTertiaryButton: SQFormDialogStory = (args) => {
     </>
   );
 };
+
+const handleTertiaryClick = (
+  formikContext: FormikContextType<DefaultArgsValues>
+) => {
+  formikContext.resetForm();
+  console.log('formikContext', formikContext);
```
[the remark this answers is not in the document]
jamesgriffith9519: Oh i had left this in so for the storybook to demonstrate getting formikContext, but we can remove it if we think it would be better!

src/components/SQFormDialog/SQFormDialogInner.tsx
20BBrown14: Doing `isTertiaryDisabled ?? !formikContext.isVaild` makes it impossible to use the `isTertiaryDisabled` and the formikContext.isValid together. Most common use case is disabling the button when the modal is loading but we also want the button disabled with the form validation.

SeanGroff: Since the conditional is not a "one or the other" we need to break this out into a helper function. How about this?

```typescript
function getIsDisabled() {
  if (isTertiaryDisabled === true || !formikContext.isValid) {
    return true
    }
    
   return false;
}
```

20BBrown14:
```
if(noValidation) {
  return Boolean(isTertiaryDisabled)
}

if (isTertiaryDisabled || !formContext.isValid) {
  return true
}

return false;

```

SeanGroff: Hrmmm....i'm leaning towards introducing a breaking change. This dialog's props API is already big, the tertiary button is adding 4 or 5 props to it.

We can eliminate the `noValidation`,  `showTertiaryButton` & `isTertiaryDisabled` props with a `tertiaryStatus` prop.

The prop can be a TS Union to capture all the possible states of the Tertiary button. While this is more verbose, I think it guards against impossible states (if we were to pass isTertiaryDisabled=true and noValidation={true}) and gives the consumer full control over the validation and rendering with one prop.

```typescript
// new Prop
type TertiaryStatus = 'HIDE_BUTTON' | 'NO_VALIDATION' | 'IS_DISABLED' | 'IS_ENABLED' | 'FORM_VALIDATION_ONLY' | 'IS_DISABLED_AND_FORM_VALIDATION'
```

Then we can use the status to derive our values

```typescript
// pseudo codez
showTertiary={status !== 'HIDE_BUTTON}
isDisabled={getIsDisabled()}

function getIsDisabled() {
   switch (status) {
      case 'IS_DISABLED':
         return true
      case 'FORM_VALIDATION_ONLY':
         return !formContext.isValid
      case 'IS_DISABLED_AND_FORM_VALIDATION':
         return status === 'IS_DISABLED' || !formContext.isValid)
       default
         return false
   }
}
```

20BBrown14: Seems reasonable with a default value of HIDE_BUTTON. We could also perhaps export an enum or something with those values for ease of use.

SeanGroff: Agreed. 
Set the default value to hide, and export the Union type to the consumers

jamesgriffith9519: @SeanGroff I think I have adjusted the code to function as you and Branden have outlined above (and ill push it up soon for review) but the one thing I'm kind of confused about is the case in which the tertiaryStatus is 'IS_DISABLED_AND_FORM_VALIDATION' .

I interpreted this as a use case in which the user wants the button to use validation to determine if the button is disabled but also another boolean value such as when a query is loading, but wouldn't we need another prop for the user to put the additional boolean value? sorry for the questions!

SeanGroff: The naming from my pseudo code above probably isn't great so feel free to use better naming.

`IS_ENABLED` would be the equivalent to `isDisabled === false`
`IS_DISABLED` would be the equivalent to `isDisabled === true`
'NO_VALIDATION` would be the equivalent to `noValidation === true // do not run any validation of any kind`

jamesgriffith9519:
Here's a new loom video showing off the changes based on our conversion in the comments and a brief meeting with Sean https://www.loom.com/share/969671a10b134c74b60a49ea36989b23

20BBrown14 (approved):

SeanGroff:
Thank you @jamesgriffith9519 @20BBrown14 @hvilander for all the valuable feedback and discussion we had in order to arrive at this solution 🚀

PR #684, opened 2022-05-20
fix: 🐛 field array valication bug

jamesgriffith9519:
fix bug that was caused by our validation errorMessage being converted
to an array when we are expecting a string.

here's a loom with my findings and a bit of info about the PR https://www.loom.com/share/a475ddc9e30e44dfb331baa78b306531

✅ Closes: 563

### DIFF
--- a/src/components/SQForm/useForm.tsx
+++ b/src/components/SQForm/useForm.tsx
@@ -78,7 +78,9 @@ export function useForm<TValue, TChangeEvent>({
   const isDirty = !isEqual(meta.initialValue, meta.value);
   const hasValue = _getHasValue(meta);
   const isError = !!errorMessage;
-  const isRequired = errorMessage?.toLowerCase() === 'required';
+  const isRequired = Array.isArray(errorMessage)
+    ? errorMessage.toString().toLowerCase() === 'required'
+    : errorMessage?.toLowerCase() === 'required';
 
   const getFieldStatus = () => {
     if (isRequired && !hasValue && !isDirty && !isTouched) {


src/components/SQForm/useForm.tsx
```diff
@@ -78,7 +78,9 @@ export function useForm<TValue, TChangeEvent>({
   const isDirty = !isEqual(meta.initialValue, meta.value);
   const hasValue = _getHasValue(meta);
   const isError = !!errorMessage;
-  const isRequired = errorMessage?.toLowerCase() === 'required';
+  const isRequired = Array.isArray(errorMessage)
```
jamesgriffith9519, line 81: the loom shows .toString() but I actually switched to .join("") so that it cuts out the commas and returns us a regular string that is not comma-separated.

SeanGroff:
😿 
https://github.com/jaredpalmer/formik/issues/3178

SeanGroff:
> 😿 [jaredpalmer/formik#3178](https://github.com/jaredpalmer/formik/issues/3178)

@jamesgriffith9519 Since this bug lives upstream within Formik we'll have to handle the bug in our code.

Let's also improve our test coverage to test against this bug.

Here's my pseudo code suggestion for the test

```javascript
  it('displays an error message when validation fails', async () => {
    const sqFormProps = {
      initialValues: {
        friends: ['joe', 'jack', 'jim', 'jill'],
      },
      validationSchema: Yup.object({
        friends: Yup.array().required().min(5),
      }),
    };

    render(<SQFormInclusionList sqFormProps={sqFormProps} />);

    // get one of the unchecked checkbox elements
   // userEvent.click(theUncheckedCheckbox);

    // check to ensure submit button is properly enabled
    const submitButton = screen.getByRole('button', {name: /submit/i});
    await waitFor(() => {
      expect(submitButton).toBeEnabled();
    });

   // uncheck a checkbox to make the validation fail (less than 5 boxes checked)
   // userEvent.click(aCheckbox)

  // get error message element
  // expect(errorMessage).text.toBe('friends field must have at least 5 items')
  
  // finally, expect submit button to be disabled
    await waitFor(() => {
      expect(submitButton).toBeDisabled();
    });
  });
```

Then for our code solution, the errorMessage itself is the problem so let's use your same solution but to handle the issue at the root of the problem. We also want to return `undefined` if there is no error instead of an empty string in order to match Formiks behavior for the errors property.

Here's my suggestion but I haven't tested it very deeply

Screen Shot 2022-05-23 at 9 15 22 AM

SeanGroff:
> > 😿 [jaredpalmer/formik#3178](https://github.com/jaredpalmer/formik/issues/3178)
> 
> @jamesgriffith9519 Since this bug lives upstream within Formik we'll have to handle the bug in our code.
> 
> Let's also improve our test coverage to test against this bug.
> 
> Here's my pseudo code suggestion for the test
> 
> ```js
>   it('displays an error message when validation fails', async () => {
>     const sqFormProps = {
>       initialValues: {
>         friends: ['joe', 'jack', 'jim', 'jill'],
>       },
>       validationSchema: Yup.object({
>         friends: Yup.array().required().min(5),
>       }),
>     };
> 
>     render(<SQFormInclusionList sqFormProps={sqFormProps} />);
> 
>     // get one of the unchecked checkbox elements
>    // userEvent.click(theUncheckedCheckbox);
> 
>     // check to ensure submit button is properly enabled
>     const submitButton = screen.getByRole('button', {name: /submit/i});
>     await waitFor(() => {
>       expect(submitButton).toBeEnabled();
>     });
> 
>    // uncheck a checkbox to make the validation fail (less than 5 boxes checked)
>    // userEvent.click(aCheckbox)
> 
>   // get error message element
>   // expect(errorMessage).text.toBe('friends field must have at least 5 items')
>   
>   // finally, expect submit button to be disabled
>     await waitFor(() => {
>       expect(submitButton).toBeDisabled();
>     });
>   });
> ```
> 
> Then for our code solution, the errorMessage itself is the problem so let's use your same solution but to handle the issue at the root of the problem. We also want to return `undefined` if there is no error instead of an empty string in order to match Formiks behavior for the errors property.
> 
> Here's my suggestion but I haven't tested it very deeply
> 
> Screen Shot 2022-05-23 at 9 15 22 AM

After writing my test example I noticed the story doesn’t render an error message.

Formik comes with an ErrorMessage component but we don’t use it so let's try to add in the SQFormHelperText component to the InclusionList WithValidation story so we can test against it. 
I haven't looked closely at the story itself so hopefully this isn't to difficult to add

jamesgriffith9519:
> > > 😿 [jaredpalmer/formik#3178](https://github.com/jaredpalmer/formik/issues/3178)
> > 
> > 
> > @jamesgriffith9519 Since this bug lives upstream within Formik we'll have to handle the bug in our code.
> > Let's also improve our test coverage to test against this bug.
> > Here's my pseudo code suggestion for the test
> > ```js
> >   it('displays an error message when validation fails', async () => {
> >     const sqFormProps = {
> >       initialValues: {
> >         friends: ['joe', 'jack', 'jim', 'jill'],
> >       },
> >       validationSchema: Yup.object({
> >         friends: Yup.array().required().min(5),
> >       }),
> >     };
> > 
> >     render(<SQFormInclusionList sqFormProps={sqFormProps} />);
> > 
> >     // get one of the unchecked checkbox elements
> >    // userEvent.click(theUncheckedCheckbox);
> > 
> >     // check to ensure submit button is properly enabled
> >     const submitButton = screen.getByRole('button', {name: /submit/i});
> >     await waitFor(() => {
> >       expect(submitButton).toBeEnabled();
> >     });
> > 
> >    // uncheck a checkbox to make the validation fail (less than 5 boxes checked)
> >    // userEvent.click(aCheckbox)
> > 
> >   // get error message element
> >   // expect(errorMessage).text.toBe('friends field must have at least 5 items')
> >   
> >   // finally, expect submit button to be disabled
> >     await waitFor(() => {
> >       expect(submitButton).toBeDisabled();
> >     });
> >   });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then for our code solution, the errorMessage itself is the problem so let's use your same solution but to handle the issue at the root of the problem. We also want to return `undefined` if there is no error instead of an empty string in order to match Formiks behavior for the errors property.
> > Here's my suggestion but I haven't tested it very deeply
> > Screen Shot 2022-05-23 at 9 15 22 AM
> 
> After writing my test example I noticed the story doesn’t render an error message.
> 
> Formik comes with an ErrorMessage component but we don’t use it so let's try to add in the SQFormHelperText component to the InclusionList WithValidation story so we can test against it. I haven't looked closely at the story itself so hopefully, this isn't to difficult to add

Ok I think I have made the changes as you have described! I made a quick loom to demo them as well. The only thing i wasn't 100% sure on was the idea of adding the SQFormHelperText component to the story so we could test, i ended up adding it to the story below the list items https://www.loom.com/share/48d0763f48d24f05900d66be3d055fa3

stories/__tests__/SQFormInclusionList.stories.test.tsx
```diff
@@ -140,4 +140,95 @@ describe('SQFormInclusionList Tests', () => {
...
+    });
+
+    await waitFor(() => {
+      expect(submitButton).toBeDisabled();
+    });
+
+    // get one of the checked checkbox elements
+    const joe = screen.getByLabelText('Joe');
+
+    //click checkbox to make validation pass by adding a 5th item to the array
```
laurelbean, line 224: I think this comment should be saying something more like 'uncheck checkbox' and then I think we also want to add another expect for the submit button after this to ensure the submit button is still disabled.

stories/__tests__/SQFormInclusionList.stories.test.tsx
```diff
@@ -140,4 +140,97 @@ describe('SQFormInclusionList Tests', () => {
...
+    userEvent.click(joe);
+
+    await waitFor(() => {
+      const errorText = screen.getByText(
+        /friends field must have at least 5 items/i
+      );
+      expect(errorText).toBeInTheDocument();
+    });
+
+    const jane = screen.getByLabelText('Jane');
```
SeanGroff, line 188: On line 179 we uncheck a checkbox and then we expect the expected error message.
I think right here we can expect the Submit button to be disabled and end the test, right?

I'm not following what clicking Jane does for us.

stories/__tests__/SQFormInclusionList.stories.test.tsx
```diff
@@ -140,4 +140,97 @@ describe('SQFormInclusionList Tests', () => {
...
+      expect(errorText).toBeInTheDocument();
+    });
+
+    // finally, expect submit button to be disabled
+    await waitFor(() => {
+      expect(submitButton).toBeDisabled();
+    });
+  });
+
+  it("It should display the error message 'friends field must have at least 5 items' when our form initially starts in a invalid state and it should continue to display the same error message after unchecking one of the already checked checkboxes", async () => {
```
SeanGroff, line 205: The test description says `it` so we don't need to begin our description with `it` as well.

I agree with testing the initial state of the form but I don't think we need to test again that unchecking another item keeps us in an error state.

So we could shorten this description down to something like 

`should display the error message "friends field must have at least 5 items" when our form is initially loaded with less than 5 items checked`

stories/__tests__/SQFormInclusionList.stories.test.tsx
jamesgriffith9519: There is a specific situation in which the error would only happen if you unchecked two items below the validation limit and this was meant to test the user unchecking two checkboxes. I may be doing a poor job of explaining but I demo it in the loom for more context

SeanGroff: Yea I saw the loom but didn't connect the dots. I'm following what you're testing against now

SeanGroff (approved):

stories/__tests__/SQFormInclusionList.stories.test.tsx
jamesgriffith9519: ahh ok! I went ahead and pushed up a commit removing the extra "it" word and removing the additional test for unchecking another item keeping us in the error state.

laurelbean (approved):

aharpt (approved):

SeanGroff (approved):

Issue #691, opened 2022-06-13
Material-UI Icons import throws an error due to missing declaration file


aharpt:
Every place that we import material-ui/icons, we get the following error: 

![material-ui icons errors](https://user-images.githubusercontent.com/16596855/173420047-07f33912-5084-471a-8b48-b090fb78e01f.png)

Pertinent part of the error message: "Could not find a declaration file for module '@material-ui/icons'.:

PR #692, opened 2022-06-13
chore: 🤖 Add missing TypeScript declaration files

aharpt:
Add missing TypeScript declaration files

https://www.loom.com/share/7178559618a24c67ba9d2a5949224cbc

✅ Closes: #691

### DIFF
--- a/src/types/material-ui-icons.d.ts
+++ b/src/types/material-ui-icons.d.ts
@@ -1,0 +1,1 @@
+declare module '@material-ui/icons';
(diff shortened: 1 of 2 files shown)


SeanGroff:
@aharpt I don't recall running into this issue. That doesn't mean it's not an issue of course. I'd rather not stub out an empty declaration file if the typings do exist though.

Looking at the source code for v4 of MUI Icons, I see this script https://github.com/mui/material-ui/blob/v4.x/packages/material-ui-icons/scripts/create-typings.js

Do you mind investigating the types issue instead of silencing TypeScript with an empty module declaration?

aharpt:
It has been determined that this PR is not needed, so am closing it.  See comment on issue #692 .

PR #696, opened 2022-07-07
feat: 🎸 Add HelperText option to Dialog footer

aharpt:
Add HelperText option to SQFormDialog footer

Main Loom: https://www.loom.com/share/6e01befa8fbb40cda9e3f12565134a29
Loom detailing consuming breaking change: https://www.loom.com/share/19a4be8f78634879805ad10ec65d608d

✅ Closes: 695

### DIFF
--- a/src/components/SQForm/SQFormHelperText.tsx
+++ b/src/components/SQForm/SQFormHelperText.tsx
@@ -36,6 +36,7 @@ const helperStateMap = {
 
 export type SQFormHelperTextProps = {
   isFailedState?: boolean;
+  isValidState?: boolean;
   errorText?: string;
   failText?: string;
   validText?: string;
@@ -43,6 +44,7 @@ export type SQFormHelperTextProps = {
 
 function SQFormHelperText({
   isFailedState = false,
+  isValidState,
   errorText = 'There is an error in the form',
   failText = 'Cannot proceed',
   validText = 'All fields completed',
@@ -55,7 +57,7 @@ function SQFormHelperText({
     switch (true) {
       case isFailedState:
         return 'fail';
-      case isValid:
+      case isValidState ?? isValid:
         return 'valid';
       default:
         return 'error';
--- a/src/components/SQFormDialog/SQFormDialog.tsx
+++ b/src/components/SQFormDialog/SQFormDialog.tsx
@@ -63,6 +63,10 @@ export type SQFormDialogProps<Values extends FormikValues> = {
    * https://jaredpalmer.com/formik/docs/guides/validation#validationschema
    * */
   validationSchema?: AnyObjectSchema;
+  /** text to eventually get passed SQFormHelperText component */
+  helperText?: string;
+  /** helper text type to eventually get passed to SQFormHelperText component */
+  helperTextType?: 'fail' | 'error' | 'valid';
 };
 
 function SQFormDialog<Values extends FormikValues>({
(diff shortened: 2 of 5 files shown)


aharpt:
Need to figure out what is up with prettier... stand by.

Ariel-Schnur-SQ:
Awesome! This is exactly what I had in mind, you are correct about not wanting to block the footer. My only question is what happens when you put a really long text string in? I'm curious if it wraps around or gets weird. Probably not an issue though because you could make the modal bigger to accommodate.

aharpt:
Yeah @Ariel-Schnur-SQ it just wraps and looks kind of weird.  Essentially you would just have to fix this by making the dialog larger.  We are giving consumers a lot of options with three buttons, Helper Text so they/we will just have to use it wisely and/or increase the size of the modals.

20BBrown14 (changes requested):
I have some requests here, need to think it out/type it out but want to block merging before someone gets too happy :P even though I think my review is now the only one that can unblock the merge button

src/components/SQForm/SQFormHelperText.tsx
```diff
@@ -36,26 +35,27 @@ const helperStateMap = {
...
   const classes = useStyles();
-  const {isValid} = useFormikContext();
   const {fail, error, valid} = helperStateMap;
 
   const getHelperTextType = () => {
     switch (true) {
       case isFailedState:
         return 'fail';
-      case isValid:
+      case isValidState:
```
20BBrown14, line 58: I think we can still use `isValid` here to ensure all our current components still work the way they're supposed to.

Still accept `isValidState` in as a prop to `SQFormHelperText` but make it optional (`isValidState?: boolean`) with no default value so that it could be undefined. Then we can use `isValidState` if it exists and give it priority. If it doesn't exist, fall back to what we had before (`isValid`). 

This route gives consumers more control over the text if they want it, but if they don't it'll work the way that it has in the past.

20BBrown14 (changes requested):
Okie doke. Good work! One suggestion, though. I think we can do this without a breaking change. Also, the current places we are using `SQFormHelperText` may not work properly with your implementation, like on the SQFormScrollableCard. If you go to that story and fill out all the fields the error helper text is still there when it should be gone.

If you make the change(s) I suggested we shouldn't have any breaking change. You'll need to adjust your commit(s) to ensure we don't have any breaking changes and if you need help with that using Git let me know. Or just let me know if you have any questions about my suggestion.

Screen Shot 2022-07-07 at 11 22 26 AM

Ariel-Schnur-SQ (approved):

aharpt:
> Okie doke. Good work! One suggestion, though. I think we can do this without a breaking change. Also, the current places we are using `SQFormHelperText` may not work properly with your implementation, like on the SQFormScrollableCard. If you go to that story and fill out all the fields the error helper text is still there when it should be gone.
> 
> If you make the change(s) I suggested we shouldn't have any breaking change. You'll need to adjust your commit(s) to ensure we don't have any breaking changes and if you need help with that using Git let me know. Or just let me know if you have any questions about my suggestion.
> 
> Screen Shot 2022-07-07 at 11 22 26 AM

@20BBrown14 Everything should be updated.  Here is a recent loom: https://www.loom.com/share/3e9e417d4e8b4695936dffd1de566e76. Let me know if anything else needs to be looked at.

20BBrown14 (approved):
🚀 Nice job

Issue #649, opened 2022-03-22
Order TS imports


20BBrown14:
We should either find an automated solution to order TS orders or go through each file and do it manually.

The order should be

modules -> local imports -> types -> css files

_Originally posted by @SeanGroff in https://github.com/SelectQuoteLabs/SQForm/pull/648#discussion_r832566295_

SeanGroff:
```
import React from 'react'
import {debounce} from 'lodash'
import Card from '../components/Card'
import {getOptions} from '../utils'
import type Moment from 'moment'
import type {Foo} from 'react'
import 'thing.css'
```

> modules -> local imports -> types -> css files
Within each section ^ the following order applies
> import * -> import Default -> import {NamedExport}

PR #698, opened 2022-07-08
chore: 🤖 Order TS Imports

aharpt:
Ordered TypeScript imports

https://www.loom.com/share/37ab6db9e898477184aeaa8400dd7f02

✅ Closes: #649


laurelbean (approved):
I think we should always have the React import at the top, even with the * Yup ones. Maybe that's just from me always seeing it that way.

20BBrown14 (approved):

lucas-homer (approved):

PR #703, opened 2022-07-25
Issue 701/add MUI5 dependencies

aharpt:
Loom: https://www.loom.com/share/c5db46730d4c43f7ab6850f471a23ac7


20BBrown14:
Could you link the upgrade guide you followed?

aharpt:
> Could you link the upgrade guide you followed?

Yep. https://mui.com/material-ui/migration/migration-v4/

package.json
```diff
@@ -90,13 +90,19 @@
     "text-mask-addons": "^3.8.0"
   },
   "peerDependencies": {
+    "@emotion/react": "^11.9.3",
+    "@emotion/styled": "^11.9.3",
     "@material-ui/core": ">= 4 < 5",
     "@material-ui/icons": ">= 4 < 5",
     "@material-ui/lab": "^4.0.0-alpha.46",
     "@material-ui/pickers": "^4.0.0-alpha.12",
+    "@mui/icons-material": "^5.8.4",
```
20BBrown14, line 99: I see we have these ones added but still the old ones. Is that intended?

20BBrown14:
I'm trying to think of the best way to do this incrementally. If I recall, and correct me if I'm wrong, on SSC we moved all stories to an `old_stories` directory and created a ticket for each component. Whenever a ticket was picked up the story was moved to the stories fold and the component and/or story were converted away from makeStyles.

Do you think we should do that in this very first ticket? Interested in thoughts here @laurelbean @lucas-homer

package.json
aharpt: Yes.  I didn't end up removing all of the old `material-ui` imports from all the component files so keeping these old dependencies is necessary for now.  I looked into changing the imports on all component files, but that complicated some stuff with the theme we are importing from shared-components.  If you want to have more discussion around this, let me know, and we can get on a call or something so I can show exactly what I was seeing.

laurelbean:
I think moving the stories worked fairly well last time. It's a good indicator for what has and hasn't completed yet. 
I also think it should be in this PR. My idea is that the result of this PR will have everything needed so anyone can take a component and get started.

20BBrown14:
Agreed Laurel. An example component being converted is a good idea in my opinion as well but not entirely necessarily. If you could incorporate the changes mentioned above, Aaron, that'd be great. Definitely let us know if you have any question.s

aharpt:
> Agreed Laurel. An example component being converted is a good idea in my opinion as well but not entirely necessarily. If you could incorporate the changes mentioned above, Aaron, that'd be great. Definitely let us know if you have any question.s

@20BBrown14 @laurelbean Sounds good, will do.

package.json
laurelbean: Looks like SSC still has the older Mui packages listed as Peer Dependencies.

aharpt: @laurelbean Correct, it does.

aharpt:
@laurelbean @20BBrown14 Updated loom: https://www.loom.com/share/5bb85be99a3c424dabdfb446b77a49c9

20BBrown14 (approved):
🚀

package.json
```diff
@@ -90,13 +90,22 @@
...
+    "@mui/material": "^5.9.2",
+    "@mui/styles": "^5.9.2",
+    "ag-grid-community": "^25.2.1",
+    "ag-grid-react": "^25.2.0",
     "@selectquotelabs/sqhooks": ">= 2",
     "moment": "^2.29.1",
-    "react": ">= 16.8.0 < 18",
+    "react": ">= 17.0.0 < 18",
     "react-dom": ">= 16.8.0 < 18",
     "scplus-shared-components": ">= 7",
```
20BBrown14, line 110: This should be updated to >= 10 now

aharpt: Are you referring to scplus-shared-components?  We want the react versions to be no less that v17?

20BBrown14: Yep! Github probably being weird and putting my comment as showing on a different line.

laurelbean (approved):

20BBrown14 (approved):

PR #707, opened 2022-07-26
refactor: 💡 migrate SQFormButton to Mui-V5

jamesgriffith9519:
convert SQFormButton to mui-v5
This component did not actually have any styling that needed to be adjusted nor any imports, so i just moved its story file down to the new stories directory as well as created the new `_tests_  `directory and moved its test file into it. 

✅ Closes: #706


stories/SQFormButton.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import {Grid} from '@material-ui/core';
 import {SQFormButton as SQFormButtonComponent, SQFormTextField} from '../src';
-import {SQFormStoryWrapper} from './components/SQFormStoryWrapper';
-import {createDocsPage} from './utils/createDocsPage';
+import {SQFormStoryWrapper} from '../old_stories/components/SQFormStoryWrapper';
+import {createDocsPage} from '../old_stories/utils/createDocsPage';
```
aharpt, line 5: This brings up an interesting question that I was running into myself on the AsyncAutoComplete component.  Should we be importing these helpers from the `old_stories` folder or should we be moving these helper files themselves so the import stays the same?  My current thought is to do it like this and leave these helpers in the `old_stories` folder and move them at the end.  The reason for this is that migrating too many files at a time will cause chain reactions that will cause problems with the files still to be migrated.  I just ran into this issue with the AsyncAutocomplete component.  Definitely willing to discuss, just my two cents.

aharpt (approved):

aharpt:
It is worth noting that one of the `SQFormButton `stories contains a `SQFormTextField `in it where the color is currently slightly off.  However, I tested and verified that this will be fixed once the `SQFormTextField `component is converted to MUI5.

stories/SQFormButton.stories.tsx
laurelbean: I'm fine waiting to migrate the helper files. I'd just want to add a checkbox/line item for migrating any remaining `old_stories` files to our list so we don't forget.

laurelbean (approved):

stories/SQFormButton.stories.tsx
aharpt: @laurelbean Yeah, that's a good idea.

20BBrown14 (approved):

PR #712, opened 2022-07-28
refactor: 💡 SQFormTextarea to MUIv5

20BBrown14:
Migrate SQFormTextarea to MUIv5

BREAKING CHANGE: 🧨 Renamed rows and maxRows props on SQFormTextarea

✅ Closes: #710


package.json
```diff
@@ -98,16 +98,15 @@
...
     "@mui/styles": "^5.9.2",
+    "@mui/x-date-pickers": "^5.0.0-beta.2",
+    "@selectquotelabs/sqhooks": ">= 2",
     "ag-grid-community": "^25.2.1",
     "ag-grid-react": "^25.2.0",
-    "@selectquotelabs/sqhooks": ">= 2",
     "moment": "^2.29.1",
     "react": ">= 17.0.0 < 18",
     "react-dom": ">= 16.8.0 < 18",
-    "scplus-shared-components": ">= 10",
```
aharpt, line 110: This may have been an accidental removal, but don't we want this shared-components declaration?

20BBrown14: we do, good catch!

laurelbean (approved):

src/components/SQForm/SQFormTextarea.tsx
```diff
@@ -97,8 +100,8 @@ function SQFormTextarea({
         onChange={handleChange}
         onBlur={handleBlur}
         required={isFieldRequired}
-        minRows={rows}
-        maxRows={rowsMax}
+        minRows={minRows}
+        maxRows={maxRows}
         variant="outlined"
```
aharpt, line 105: Maybe not super-intuitive, but per the migration docs for the `TextField` component (https://mui.com/material-ui/migration/v5-component-changes/#textfield) the outlined variant is now the default variant so we can remove line 105.  Not absolutely necessary IMO just thought I would call it out.

aharpt (approved):

PR #736, opened 2022-08-03
Issue 735/Migrate SQForm MultiValue to MUI5

aharpt:
https://www.loom.com/share/36530b785ba446439e6906d8a36197c1


package-lock.json
```diff
@@ -120,6 +120,7 @@
         "moment": "^2.29.1",
         "react": ">= 17.0.0 < 18",
         "react-dom": ">= 16.8.0 < 18",
+        "scplus-shared-components": ">= 10",
```
aharpt, line 123: Somehow this got removed in a previous PR.  We need this, so am adding it back.

src/components/SQForm/SQFormMultiValue.tsx
```diff
@@ -269,50 +253,51 @@ function SQFormMultiValue({
       'props' in endAdornment &&
       endAdornment.props
     ) {
-      const newEndAdornment = {
+      newEndAdornment = {
         ...endAdornment,
         props: {
           ...endAdornment.props,
-          className: `${endAdornment.props.className} ${clearButtonClasses.endAdornment}`,
+          style: {top: 'auto', bottom: '8px'},
```
20BBrown14, line 260: If I'm remembering correctly, the main driver behind this was that the little "x" icon to clear in the input field would do weird things if there were a lot of options and caused the field to be multiline. Coudl you confirm that if you add like 20 options and the input fields has to expand vertically that the x button is behaving properly.

20BBrown14 (changes requested):

src/components/SQForm/SQFormMultiValue.tsx
aharpt: @20BBrown14 See loom for how it is currently behaving.  I believe this is what we want.

https://www.loom.com/share/c3133afc51b7438fb8f0c700ca1142d3

20BBrown14 (approved):

Issue #358, opened 2021-07-21
Add All Component Tests


laurelbean:
List of all remaining components that still need tests added:

- [ ] #265 
- [ ] #266 
- [ ] #267 
- [ ] #269 
- [ ] SQFormGuidedWorkflow
- [ ] SQFormInclusionList
- [ ] SQFormMaskedTextField
- [ ] #307 
- [ ] SQFormResetButtonWithConfirmation
- [ ] SQFormResetInitialValuesButton
- [ ] SQFormScrollableCard

PR #747, opened 2022-08-04
refactor: 💡 Migrate DatePicker to MUI5

aharpt:
`SQFormDatePicker`: https://www.loom.com/share/f5758d90e3104eaeac4a1e1f895e2c2f
`SQFormDatePickerWithDateFNS`: https://www.loom.com/share/147125be9eba44ccb80ea057a08c6853

✅ Closes: #745 #748

### DIFF
--- a/stories/__tests__/SQFormDatePicker.stories.test.tsx
+++ b/stories/__tests__/SQFormDatePicker.stories.test.tsx
@@ -1,9 +1,9 @@
 import React from 'react';
 import userEvent from '@testing-library/user-event';
-import MomentAdapter from '@material-ui/pickers/adapter/moment';
 import {composeStories} from '@storybook/testing-react';
 import {render, screen, within, waitFor} from '@testing-library/react';
-import {LocalizationProvider} from '@material-ui/pickers';
+import {LocalizationProvider} from '@mui/x-date-pickers';
+import {AdapterMoment} from '@mui/x-date-pickers/AdapterMoment';
 import * as stories from '../SQFormDatePicker.stories';
 import type {SQFormDatePickerProps} from 'components/SQForm/SQFormDatePicker';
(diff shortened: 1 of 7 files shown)


stories/__tests__/SQFormDatePicker.stories.test.tsx
```diff
@@ -96,8 +96,8 @@ describe('SQFormDatePicker Tests', () => {
     expect(calendarDialog).toBeInTheDocument();
     expect(calendarDialog).toBeVisible();
 
-    const dateOptions = within(calendarDialog).getAllByRole('cell');
-    const selectedDate = dateOptions[0];
+    const dateOptions = within(calendarDialog).getAllByRole('button');
+    const selectedDate = dateOptions[4];
```
aharpt, line 100: I had issues with this story not passing for me.  We needed to actually get all of the buttons within the `calendarDialog`.  The 1st of the month corresponds to the fifth button inside the `calendarDialog`.  A bit hacky, but the best I could figure out.  I can revisit if needed, but don't think that it is very likely we would be adding/removing buttons that would cause this hard-coded value not to work.

laurelbean: I think this will work for now. I already have a line item to go update the tests for this component once we upgrade to version 5.
https://github.com/SelectQuoteLabs/SQForm/issues/358

aharpt:
Note: This PR also migrates the `DatePickerWithDateFNS` component.  I had to do both components in one PR since the DatePickerWithDateFNS comonent relies on props from the regular date picker which was causing the tests to fail.  I have a separate loom for each component if people are interested

laurelbean (approved):

20BBrown14 (approved):

PR #753, opened 2022-08-05
refactor: 💡 Migrate DateTimePicker to MUI5

aharpt:
Please watch the loom as it highlights differences between the previous and latest `DateTimePicker`.

https://www.loom.com/share/77e4f85b79e045eea09389d5635d03b8

✅ Closes: #751

### DIFF
--- a/src/components/SQForm/SQFormDateTimePicker.tsx
+++ b/src/components/SQForm/SQFormDateTimePicker.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import {ClickAwayListener, Grid, TextField} from '@mui/material';
 import {DateTimePicker} from '@mui/x-date-pickers/DateTimePicker';
-import {MarkOptional} from 'ts-essentials';
 import {useForm} from './useForm';
+import type {MarkOptional} from 'ts-essentials';
 import type {Moment} from 'moment';
 import type {BaseDateTimePickerProps} from '@mui/x-date-pickers/DateTimePicker/shared';
 import type {BaseFieldProps} from '../../types';
@@ -63,7 +63,6 @@ function SQFormDateTimePicker({
     <ClickAwayListener onClickAway={handleClickAway}>
       <Grid item={true} sm={size}>
         <DateTimePicker
-          {...muiFieldProps}
           label={label}
           disabled={isDisabled}
           value={value}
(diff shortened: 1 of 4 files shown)


src/components/SQForm/SQFormDateTimePicker.tsx
```diff
@@ -1,15 +1,10 @@
...
-import {
-  ClickAwayListener,
-  Grid,
-  makeStyles,
-  TextField,
-} from '@material-ui/core';
-import {DateTimePicker} from '@material-ui/pickers';
+import {ClickAwayListener, Grid, TextField} from '@mui/material';
+import {DateTimePicker} from '@mui/x-date-pickers/DateTimePicker';
+import {MarkOptional} from 'ts-essentials';
```
20BBrown14, line 4: Should be using `import type`?

src/components/SQForm/SQFormDateTimePicker.tsx
```diff
@@ -77,8 +61,9 @@ function SQFormDateTimePicker({
 
   return (
     <ClickAwayListener onClickAway={handleClickAway}>
-      <Grid item sm={size}>
+      <Grid item={true} sm={size}>
         <DateTimePicker
+          {...muiFieldProps}
```
20BBrown14, line 66: I think we should spread this _after_ our props so that the consumer can override anything they want to. Likely not a huge concern but I think that'd be best unless someone has any reasons why we shouldn't.

20BBrown14 (approved):
Looks good with some minor comments. I wouldn't worry too much about the date time picker UI adjustments in MUIv5 since we usually have no problem just adhering to whatever it is that MUI is doing. 

Approved with 2 minor changes

src/components/SQForm/SQFormDateTimePicker.tsx
aharpt: Yes.  I'll fix this.

src/components/SQForm/SQFormDateTimePicker.tsx
aharpt: @20BBrown14 The new MUI5 type for `muiFieldProps` creates some issues with overriding other component props that caused me to have to move this to the top of the component.  After thinking about this further, I should be able to initialize `muiFieldProps` to have all of our default props which should allow us to fix this issue.  I'll include it in a follow-up commit.

PR #760, opened 2022-08-08
chore: 🤖 Fix installation of node_modules in SQFormDocs

20BBrown14:
Turns out you can't use `cd` in the pipeline but we can use the --prefix flag to install them without needing to be in that directory.


laurelbean (approved):
Nice find. Might be a good FE topic to go over some of the docker improvements you've done for SSC and here?

PR #771, opened 2022-08-15
refactor: 💡 Migrate OutcomeForm to MUI5

aharpt:
The only place we use the `OutcomeForm `component is the `SQFormGuidedWorkflow` component which has not been migrated yet (issue here: https://github.com/SelectQuoteLabs/SQForm/issues/766).  These changes are quite minimal, but if people have a preference to wait on this PR until that component has been completed, we can certainly do that.

✅ Closes: #770


laurelbean:
Since this component is part of SQFormGuidedWorkflow, I would have included it in the Guided Workflow issue. But it looks like there won't be any conflicting changes so it should be fine to go on it's own.

laurelbean (approved):

PR #774, opened 2022-08-15
refactor: 💡 Migrate TopSubHeader to MUI5

aharpt:
https://www.loom.com/share/46ecca4a99904de2a418044ba6a53656

✅ Closes: #772


caleboleary (approved):

laurelbean:
If TopSubHeader isn't being used anywhere, my opinion is to delete it instead of keeping it through this migration.

aharpt:
> If TopSubHeader isn't being used anywhere, my opinion is to delete it instead of keeping it through this migration.

@laurelbean Agreed.  My original thought was that we are letting our consumers use this component since it is exported from the file.  However, when I look into this more, we are only exporting it from the file, and not actually allowing it to be consumed.  I'll remove this component.

laurelbean (approved):

20BBrown14 (approved):

PR #775, opened 2022-08-16
Issue 773/Migrate ScrollableCard to MUI5

aharpt:
https://www.loom.com/share/44ec39e7e86a4acaa49c76b6236772d8

### DIFF
--- a/src/components/SQFormScrollableCard/SQFormScrollableCard.tsx
+++ b/src/components/SQFormScrollableCard/SQFormScrollableCard.tsx
@@ -1,26 +1,18 @@
 import React from 'react';
-import {
-  Card,
-  CardHeader,
-  CardContent,
-  CardActions,
-  Grid,
-  makeStyles,
-} from '@material-ui/core';
+import {Card, CardHeader, CardContent, CardActions, Grid} from '@mui/material';
 import {Formik, Form} from 'formik';
 import {useDebouncedCallback} from 'use-debounce';
 import SQFormButton from '../SQForm/SQFormButton';
 import SQFormHelperText from '../SQForm/SQFormHelperText';
 import {useInitialRequiredErrors} from '../../hooks/useInitialRequiredErrors';
 import {HEADER_HEIGHT} from '../../utils/constants';
-import type {CreateCSSProperties} from '@material-ui/core/styles/withStyles';
-import type {TypographyVariant, GridProps} from '@material-ui/core';
+import type {GridProps} from '@mui/material';
 import type {FormikHelpers, FormikValues} from 'formik';
 import type {AnyObjectSchema} from 'yup';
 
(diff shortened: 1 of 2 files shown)


src/components/SQFormScrollableCard/SQFormScrollableCard.tsx
```diff
@@ -74,58 +66,10 @@ export type SQFormScrollableCardProps<Values extends FormikValues> = {
   validationSchema?: AnyObjectSchema;
   /** Boolean used to determine if title/header is enabled or disabled */
   isHeaderDisabled?: boolean;
-  /** MUI Typography variant to be used for the title */
-  titleVariant?: TypographyVariant;
```
aharpt, line 78: This prop was not used anywhere in the component so I just deleted it.

src/components/SQFormScrollableCard/SQFormScrollableCard.tsx
```diff
@@ -1,26 +1,18 @@
...
-import type {TypographyVariant, GridProps} from '@material-ui/core';
+import type {GridProps} from '@mui/material';
 import type {FormikHelpers, FormikValues} from 'formik';
 import type {AnyObjectSchema} from 'yup';
 
 export type SQFormScrollableCardProps<Values extends FormikValues> = {
-  /** An object of css-in-js style properties to be passed and spread onto `classes.cardContent` */
-  cardContentStyles?: CreateCSSProperties;
+  /** An object of css-in-js style properties to be passed and spread onto the `CardContent` component */
+  cardContentStyles?: Record<string, unknown>;
```
laurelbean, line 15: I think we should use the `SxProps` type for these. 
`import { SxProps } from '@mui/material'`
or if that doesn't work, then try 
`import { SxProps } from '@material-ui/system';`

found here: https://github.com/mui/material-ui/issues/27564

20BBrown14, line 15: We should be slight more lenient with our typing here. I think we should crib exactly whatever the `sx` prop is. I think it also accepts an array? Is there a particular reason some variation of CSS PRoperties wouldn't work?

aharpt: Yeah that's a better idea.  I'll change that.

aharpt: @20BBrown14 I am working on trying to implement @laurelbean's suggestion.  I couldn't find the `CreateCSSProperties` type in MUI5 which is the reason for the change.

aharpt: @laurelbean @20BBrown14 Here is an updated loom outlining the solution I came to and why I went that direction.  Feel free to take a look if interested. https://www.loom.com/share/41cc6dc460994552ad3354c56c3c9690

PR #780, opened 2022-08-16
chore: 🤖 Update HelperText styles

20BBrown14:
Update HelperText styles to utilize the new MUIv5 theme from SSC

✅ Closes: #779

For all screenshots - New on left, old on right
Screen Shot 2022-08-16 at 1 00 16 PM
Screen Shot 2022-08-16 at 1 00 06 PM
Screen Shot 2022-08-16 at 12 58 03 PM
Screen Shot 2022-08-16 at 12 57 55 PM

### DIFF
--- a/.storybook/styles/AgGrid.css
+++ b/.storybook/styles/AgGrid.css
@@ -1,11 +1,17 @@
 /* ag-grid theme override css */
 .ag-header-cell {
   font-size: 12px;
-  background-color: var(--color-stone);
+  background-color: var(--color-rock);
   color: var(--color-white);
   width: auto;
   font-family: 'Open Sans';
   font-weight: 400;
+  padding-left: 10px !important;
+  padding-right: 10px !important;
+}
+
+.ag-header-cell::after {
+  border-right: 1px solid #a4a6a8 !important;
 }
 
 .ag-theme-balham .ag-row-hover {
@@ -73,12 +79,14 @@
 }
 
 .ag-header {
-  background-color: var(--color-stone) !important;
+  background-color: var(--color-rock) !important;
 }
 
 .columnList__hyperlink {
-  color: var(--color-teal);
+  color: var(--color-darkTeal);
   cursor: pointer;
+  font-size: 14px;
+  font-weight: 600;
 }
 
 .columnList__number {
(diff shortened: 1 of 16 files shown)


.storybook/styles/AgGrid.css
```diff
@@ -1,11 +1,17 @@
 /* ag-grid theme override css */
```
20BBrown14, line 1: All the changes to these .css files are just changes that I just copy pasted from the latest version of SSC

src/components/SQForm/useForm.tsx
```diff
@@ -128,24 +135,26 @@ export function useForm<TValue, TChangeEvent>({
     if (isFieldError) {
       return (
         <>
-          <WarningIcon color="error" style={SPACE_STYLE} />
-          {errorMessage}
+          <WarningIcon color="error" sx={ICON_STYLES} />
+          <Typography sx={(theme: Theme) => theme.typography.helper}>
```
20BBrown14, line 139: We weren't originally using the theme or a Typography component for these helper text messages. So added that here and made sure we were using the correct styling

stories/__tests__/SQFormAsyncAutocomplete.stories.test.tsx
```diff
@@ -167,12 +167,14 @@ describe('SQFormAsyncAutocomplete Tests', () => {
...
-      render(<SQFormAsyncAutocompleteWithValidation size="auto" />);
+      const {container} = render(
+        <SQFormAsyncAutocompleteWithValidation size="auto" />
+      );
 
       await waitFor(() => {
-        const requiredText = screen.getByText(/required/i);
-        expect(requiredText).toBeVisible();
-        expect(requiredText).toHaveClass('Mui-required');
+        const helperText = container.querySelector('.MuiFormHelperText-root');
```
20BBrown14, line 175: All of these changes in these tests are going to be similar. Considering we're using the typography now the parent of the "text element" is what actually has the class we're looking for. So that's what we're getting here.

src/types/@mui/material/styles.d.ts
```diff
@@ -0,0 +1,9 @@
+import type {CSSProperties} from '@mui/styles';
```
20BBrown14, line 1: We need to declare this module to tell MUI what custom properties we have in our Theme. This might be something we try to expect directly from SSC when we convert it to TS?

aharpt (approved):

Issue #762, opened 2022-08-10
ResetForm() doesn't re-validate yup schema


hvilander:
https://github.com/jaredpalmer/formik/issues/3512

A potential issue with formik, that we may want a work around too. 
Resetting a form may not fire off yup revalidation.

You can see an example in this loom: https://www.loom.com/share/fce9ad5d04a14522bdcc29c2b02fc62e

(Assuming I have not hit my limit and deleted it)

PR #781, opened 2022-08-16
fix: 🐛 Fix resetForm not re-validating

aharpt:
Fix resetForm() not re-validating.  Fix suggested towards the bottom of the linked Github issue. https://github.com/jaredpalmer/formik/issues/3512.

https://www.loom.com/share/d302bc139c284c9e89e0764da6fbf4ba

Updated Loom: https://www.loom.com/share/a5ba4a97030748a0962cadfd409b1cc5

✅ Closes: #762


src/components/SQForm/SQFormButton.tsx
```diff
@@ -3,34 +3,38 @@ import {RoundedButton} from 'scplus-shared-components';
...
+      shouldRequireFieldUpdates,
+      onClick,
+      buttonType: type,
+    }
+  );
 
   const getClickHandler = (event: React.MouseEvent<HTMLButtonElement>) => {
     if (isResetButton) {
-      return handleReset;
+      return initialValues ? () => setValues(initialValues) : handleReset;
```
20BBrown14, line 37: Hm. This is def weird. Let's add a comment here with the relevant information about why we're doing this so we don't forget. A link to an issue  would be great so that we can see when/if it's fixed.

src/components/SQForm/SQFormButton.tsx
```diff
@@ -3,34 +3,41 @@ import {RoundedButton} from 'scplus-shared-components';
 import {useFormButton, BUTTON_TYPES} from './useFormButton';
 import type {ButtonType} from './useFormButton';
 
-export type SQFormButtonProps = {
+export type SQFormButtonProps<Values = void> = {
```
laurelbean, line 6: I think we need to use something like `unknown` for this default value. `void` is for functions, but you're using Values as a type for an object.

aharpt: @laurelbean Sounds good.  I'll look into updating this.

laurelbean:
`initialValues` is a value that you can get from `useFormikContext` which can be gotten from `useFormButton`.
So this can be simplified into something more like:
```
const {. . ., initialValues, setValues} = useFormButton({. . .});

const formReset = (event: React.MouseEvent<HTMLButtonElement>) => {
    handleReset(event);  //pass through event
    setValues(initialValues); //not sure if this should go 1st or 2nd
  };

  const getClickHandler = (event: React.MouseEvent<HTMLButtonElement>) => {
    if (isResetButton) {
      return formReset;
    } 
    . . .
  };
  ```
  
  I haven't tested it, so might need some adjustments.
  
  This doesn't solve the helper text issue though. You might want to look into trying adding in `validateForm` or `setErrors` to get the form in the right state for the helper text.

laurelbean:
Another option we could potentially pursue would be downgrading the Formik version to one that doesn't have this reset issue. As long as it doesn't re-introduce other big issues.

aharpt:
@laurelbean @20BBrown14 I am hoping that I have finally gotten to the bottom of the issue and found a decent solution.  Here is an updated loom: https://www.loom.com/share/a5ba4a97030748a0962cadfd409b1cc5

cc: @lucas-homer @hvilander

hvilander (approved):

hvilander:
Great work Aaron!!

laurelbean (approved):

20BBrown14 (approved):

lucas-homer (approved):

jamesgriffith9519 (approved):
This looks similar to how we addressed the issue locally in the AQE2 project and from what I can tell it seems to work well!

laurelbean:
Unless someone needs this soon, I think this could wait until after our Mui 5 update to try and avoid merge conflicts. If it's needed sooner though I have no issues pushing it.

hvilander:
> Unless someone needs this soon, I think this could wait until after our Mui 5 update to try and avoid merge conflicts. If it's needed sooner though I have no issues pushing it.

Agreed, at least I don't think this is immediately urgent. Where I bumped into it is long from being put into production.

PR #783, opened 2022-08-17
Update CODEOWNERS

laurelbean:
This will add Lucas and myself as approving reviewers on PRs.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @SeanGroff @codyarose @20BBrown14
+* @20BBrown14 @lucas-homer @laurelbean


aharpt (approved):

20BBrown14 (approved):

20BBrown14:
Will this commit cause any weird deployments?

laurelbean:
@20BBrown14 I haven't done changes to `.github` files before, so I'm not 100% sure, but I wouldn't think so? I'm not changing any code. There is potential for it to kick off a new build, but it would just be building the exact same as what's already out there.

20BBrown14:
> @20BBrown14 I haven't done changes to `.github` files before, so I'm not 100% sure, but I wouldn't think so? I'm not changing any code. There is potential for it to kick off a new build, but it would just be building the exact same as what's already out there.

Makes sense to me. I wonder if we should make sure this is a chore commit or similar just in case so we don't bump the version for no reason.

laurelbean:
Yeah, probably safer. I'll close this and push a chore commit instead.

Issue #808, opened 2022-10-04
Migrate onChange validation to occur onBlur by default for inputs


samasastudio:
Ticket tracking issue: https://selectquote.atlassian.net/browse/SSR-149

Overall would like to see inputs validate inputs onBlur so that validation errors/warnings do not throw while the user is typing/making changes.

Loom!  https://loom.com/share/0627d64b2ac347cbb395ac25edd30157

samasastudio:
After discussions and thoughts in both our frontend chapter meetings and UX team, I'm deciding to keep the current behavior as default while making validation onBlur and onChange as options that can be set to true or false via new `validateOnBlur` and `validateOnChange` props.

PR #809, opened 2022-10-05
SSR 149 validateOnBlur and validateOnChange options

samasastudio:
This adds the feature to be able to control if we are validating onChange or onBlur for both the SQForm and SQFormDialog components.  By default both are set to true (as they have behaved previously), and each can be deactivated by setting to false.

closes issue [#808 ](https://github.com/SelectQuoteLabs/SQForm/issues/808)

Ticket [SSR-149](https://selectquote.atlassian.net/browse/SSR-149?atlOrigin=eyJpIjoiNTc4MDcwYjk1OTBjNGQ1ZDk2YzdlMWJlNzY4YTc4ZTEiLCJwIjoiaiJ9)

[Loom video demoing new feature and stories](https://www.loom.com/share/2d032be41aee4e0f9f5afebde7bb0224)

### DIFF
--- a/src/components/SQForm/SQForm.tsx
+++ b/src/components/SQForm/SQForm.tsx
@@ -37,6 +37,10 @@ export type SQFormProps<Values extends FormikValues> = {
    * https://jaredpalmer.com/formik/docs/guides/validation#validationschema
    * */
   validationSchema?: AnyObjectSchema;
+  // Set if validation should happen on blur or not.  By default is set to true.
+  validateOnBlur?: boolean;
+  // Set if validation should happen on change.  By defautl is set to true.
+  validateOnChange?: boolean;
 };
 
 function SQForm<Values extends FormikValues>({
@@ -46,6 +50,8 @@ function SQForm<Values extends FormikValues>({
   muiGridProps = {},
   onSubmit,
   validationSchema,
+  validateOnBlur,
+  validateOnChange,
 }: SQFormProps<Values>): JSX.Element {
   const initialErrors = useInitialRequiredErrors(
     validationSchema,
@@ -76,6 +82,8 @@ function SQForm<Values extends FormikValues>({
       onReset={handleReset}
       validationSchema={validationSchema}
       validateOnMount={true}
+      validateOnBlur={validateOnBlur}
+      validateOnChange={validateOnChange}
     >
       {(_props) => {
         return (
--- a/src/components/SQForm/SQForm.tsx
+++ b/src/components/SQForm/SQForm.tsx
@@ -37,9 +37,9 @@ export type SQFormProps<Values extends FormikValues> = {
    * https://jaredpalmer.com/formik/docs/guides/validation#validationschema
    * */
   validationSchema?: AnyObjectSchema;
-  // Set if validation should happen on blur or not.  By default is set to true.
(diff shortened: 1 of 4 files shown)


src/components/SQForm/SQForm.tsx
```diff
@@ -46,6 +50,8 @@ function SQForm<Values extends FormikValues>({
   muiGridProps = {},
   onSubmit,
   validationSchema,
+  validateOnBlur,
+  validateOnChange,
```
aharpt, line 54: I think both of these props should be `true` by default so this PR does not introduce any breaking changes.

src/components/SQFormDialog/SQFormDialog.tsx
```diff
@@ -96,6 +100,8 @@ function SQFormDialog<Values extends FormikValues>({
   helperText,
   helperTextType = 'error',
   throwAlertOnCancel = true,
+  validateOnBlur,
+  validateOnChange,
```
aharpt, line 104: Same here about setting to `true` by default.

src/components/SQForm/SQForm.tsx
samasastudio: Yeah they are already set to true at the Formik level, so this doesn't contain any breaking changes.  If we want to add a belt to our suspenders code wise to for readability and an extra layer of reinforcement then I'm ok with that.  Let me know!

src/components/SQFormDialog/SQFormDialog.tsx
samasastudio: same as above

aharpt (approved):

src/components/SQForm/SQForm.tsx
samasastudio: Absolutely, thanks for asking!

src/components/SQForm/SQForm.tsx
```diff
@@ -37,6 +37,10 @@ export type SQFormProps<Values extends FormikValues> = {
    * https://jaredpalmer.com/formik/docs/guides/validation#validationschema
    * */
   validationSchema?: AnyObjectSchema;
+  // Set if validation should happen on blur or not.  By default is set to true.
+  validateOnBlur?: boolean;
+  // Set if validation should happen on change.  By defautl is set to true.
```
20BBrown14, line 42: nit: typo default

Also, could we mention that it's set to `true` by default in Formik and not in our code so it doesn't get confusing if someone is reading this documentation when we're clearly not setting it to default true on our own.

20BBrown14 (approved):

src/components/SQForm/SQForm.tsx
samasastudio: great point and good catch on the typo!  will make this change now.

laurelbean (approved):
As far as the tests go, I believe they were commented out because the Autocomplete component spams the logs with a warning. Since you're not using an Autocomplete, it would be nice if you could add a test case for your stories, but I won't require it since as you said the validation tests are commented currently.

src/components/SQForm/SQForm.tsx
samasastudio: pushed 👍

Issue #813, opened 2022-11-04
Zero value not showing in read only field


hvilander:
If you want to display the number zero as a read only value it will not show in SQFormReadOnlyField because of this OR. https://github.com/SelectQuoteLabs/SQForm/blob/master/src/components/SQForm/SQFormReadOnlyField.tsx#L32

Zero could be a real value and should not be discarded.

PR #814, opened 2022-11-04
fix: 🐛 ReadOnlyField didn't accept zero

aharpt:
Fix Issue where ReadOnlyField Didn't accept Zero

https://www.loom.com/share/9335149a27be411db6326a44bcc3b413

✅ Closes: #813

### DIFF
--- a/src/components/SQForm/SQFormReadOnlyField.tsx
+++ b/src/components/SQForm/SQFormReadOnlyField.tsx
@@ -23,13 +23,15 @@ function SQFormReadOnlyField({
     formikField: {field},
   } = useForm({name});
 
+  const getFieldValue = field.value || field.value === 0 ? field.value : '- -';
+
   return (
     <Grid item sm={size}>
       <TextField
         id={toKebabCase(name)}
         label={label}
         name={name}
-        value={field.value || '- -'}
+        value={getFieldValue}
         fullWidth={true}
         InputLabelProps={{shrink: true}}
         InputProps={{


src/components/SQForm/SQFormReadOnlyField.tsx
```diff
@@ -23,13 +23,15 @@ function SQFormReadOnlyField({
     formikField: {field},
   } = useForm({name});
 
+  const getFieldValue = field.value || field.value === 0 ? field.value : '- -';
```
hvilander, line 26: I was kinda thinking it should just be checking for null or undefined...
`getFieldValue = field.value ?? '- -'`

Maybe I am missing something?

aharpt: @hvilander Agreed.  We would want to allow for things like false or the empty string.  I will update this.

aharpt: @hvilander Ok, have to go back to the drawing board a bit.  We can't do it that way because we NEED to render the '- -' when we have an empty string (which is falsy).  This is breaking our tests that are expecting to see a `- -` when we don't provide a value.  So we need to render the '- -' for `null`, `undefined`, and `''`.  Here is probably the nicest way to do it (although requires use of the loose equality operator which we may not want):

`field.value == null || field.value === '' ? '- -' : field.value`

This works since  `undefined == null`. Let me know if you can think of a better way to accomplish this

20BBrown14: What about 

`field.value !== null && field.value !== undefined && field.value.toString() ? field.value || '- -'`

aharpt: @20BBrown14 I like it, I think that should work.

hvilander: +1 to branden's solution. 
I am afraid of double equals and don't trust it.

samasastudio (approved):

src/components/SQForm/SQFormReadOnlyField.tsx
samasastudio: @aharpt throwing `!field.value && !Number.isInteger(field.value) ? '- -' : field.value` into the ring as well (i think it would work?), but this looks great and no change needed.

hvilander (approved):

Issue #828, opened 2023-01-23
Update to latest SSC version


laurelbean:
After the TypeScript update is complete, update the version here.

20BBrown14:
Work in progress branch: https://github.com/SelectQuoteLabs/SQForm/tree/issue/828 (also addresses #829 )

Issue #829, opened 2023-01-23
Update ag-grid versions to match SSC


laurelbean:
Match `ag-grid-community` and `ag-grid-react` versions to same version SSC uses

PR #833, opened 2023-02-02
feat: 🎸 Upgrade to SCPlus-Shared-Components@12

20BBrown14:
Upgraded SCPlus-Shared-Components to version 12 and fixed build, and type errors. No breaking changes

I tested these changes in an ancillary app and didn't notice any issues.
Pulling this down and testing it in a real project isn't really possible either since there are none on Mui v5 afaik.
So long as there are no breaking changes we should be good. I am not aware of any in this PR.

✅ Closes: #828, #829


src/components/SQForm/SQFormAutocomplete.tsx
```diff
@@ -72,12 +72,13 @@ const OuterElementType = React.forwardRef<HTMLDivElement>(
   (props: OuterElementTypeProps, ref) => {
     const outerProps = React.useContext(OuterElementContext);
     return (
-      <div
+      <Box
+        component="div"
```
laurelbean, line 76: Box defaults to div

src/components/SQFormDialog/SQFormDialog.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import {Formik} from 'formik';
-import SQFormDialogInner from './SQFormDialogInner';
+import SQFormDialogInner, {SQFormDialogInnerProps} from './SQFormDialogInner';
```
laurelbean, line 3: I think our convention is to split this into 2 import statements since of them is a type.

laurelbean (approved):

src/components/SQFormDialog/SQFormDialog.tsx
hvilander: agreed

hvilander (approved):

aharpt (approved):

Issue #832, opened 2023-01-26
SQFormDatePicker throws error on use


laurelbean:
After upgrading to the latest version of SQForm, any forms that use date pickers are breaking with the error:
`MUI: Can not find utils in context. It looks like you forgot to wrap your component in LocalizationProvider, or pass dateAdapter prop directly.`

Loom: https://www.loom.com/share/5c1d81dd1aaf4a9f8fd5695d5d012338

laurelbean:
This should be solved by PR #833 
Some combination of updating `@mui/x-date-pickers` and removing the need for `--legacy-peer-deps` seems to have solved this.

PR #835, opened 2023-02-07
Fixed datepicker issue

20BBrown14:
Fixed the datepicker issue that we were seeing in consumer apps when trying to render it. The error had a message like `MUI Utils not in context`.

This PR effectively removes the need to use the `--legacy-peer-deps` flag when installing node modules. This required updating a bunch of packages. I haven't found any issues as a result of these upgrades which I tested in an ancillary app and in sc2-senior.

I'm still not 100% what was causing the issue but mismatch MUI dep versions, or a broken dep tree due to the the `--legacy-peer-dep` flag could have been the culprit.

This should release v10.1.1

Closes: #832


laurelbean (approved):
Thanks for taking this on!

src/components/SQForm/SQFormReadOnlyField.tsx
```diff
@@ -42,7 +42,6 @@ function SQFormReadOnlyField({
         InputProps={{
           ...InputProps,
           readOnly: true,
-          disableUnderline: true,
```
hvilander, line 45: I assume this is a new default?

package.json
```diff
@@ -62,12 +62,10 @@
   "lint-staged": {
     "*.{js,ts,tsx}": [
       "prettier --write",
-      "eslint --fix",
-      "git add"
```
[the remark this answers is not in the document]
20BBrown14: The latest version of the `lint-staged` package automatically does a `git add` after each command so no need to have it in our command list

src/components/SQForm/SQFormReadOnlyField.tsx
20BBrown14: I believe this prop is no longer a prop after upgrading. Which is why it was removed.

hvilander (approved):

package.json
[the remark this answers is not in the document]
hvilander: :party parrot:

src/components/SQForm/SQFormReadOnlyField.tsx
hvilander: fair

hvilander (approved):

samasastudio (approved):

aharpt (approved):

Issue #836, opened 2023-02-07
update Autocomplete styles


samasastudio:
Update the current implementation of the Autocomplete with added specifications. Here is the current implementation of this SQForm:  [Webpack App](https://master--5f4431386ea00a00220d495c.chromatic.com/?path=/story/components-sqformautocomplete--default)  (this is an example link)

Design reference: [SC3 DESIGN LIBRARY](https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=1518%3A11678) 
Typography: [SC3 DESIGN LIBRARY](https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=340%3A9946) 
Color: [SC3 DESIGN LIBRARY](https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=340%3A9646)

Default:
Label:  Text Style: Label 12 px Weight: 400
Value: Text Style: Value 14px Weight: 600 Semi-bold

Focused:
Label: Text Style: Label 12 px Weight: 400
Value: Text Style: Value 14px Weight: 600 Semi-bold

With Validation:
Label: Text Style: Label 12 px Weight: 400
Value: Text Style: Value 14px Weight: 600 Semi-bold
Line: #C67000 Warning 
Helper Text/icon: #C67000 Warning   #3C7107 Text/Success 
Icon Size: 12 x 11 Warning  9 x 11 Success badge

samasastudio:
Will be resolved by this PR: https://github.com/SelectQuoteLabs/SQForm/pull/805

samasastudio:
Reopening as updates were only applied to AsyncAutocomplete component and need to be expanded to vanilla Autocomplete.

PR #841, opened 2023-02-10
style: 💄 update SQFormAutocomplete styles

samasastudio:
Updated sizing and weight for label and value text. Switched success and warning colors in validation to new 'textWarningGYellow' and 'textSuccessGreen' styles.

Update the current implementation of the Autocomplete with added specifications. Here is the current implementation of this SQForm: [Webpack App](https://master--5f4431386ea00a00220d495c.chromatic.com/?path=/story/components-sqformautocomplete--default) (this is an example link)

Design reference: [SC3 DESIGN LIBRARY](https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=1518%3A11678)
Typography: [SC3 DESIGN LIBRARY](https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=340%3A9946)
Color: [SC3 DESIGN LIBRARY](https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=340%3A9646)

Default:
Label: Text Style: Label 12 px Weight: 400

Focused:
Label: Text Style: Label 12 px Weight: 400

With Validation:
Label: Text Style: Label 12 px Weight: 400
Line: #C67000 Warning
Helper Text/icon: #C67000 Warning #3C7107 Text/Success
Icon Size: 12 x 11 Warning 9 x 11 Success badge

✅ Closes: #836

Loom Demo: [here](https://www.loom.com/share/c5c74522ce0a405db62ec1d2beecf679)

### DIFF
--- a/src/components/SQForm/SQFormAutocomplete.tsx
+++ b/src/components/SQForm/SQFormAutocomplete.tsx
@@ -319,6 +319,22 @@ function SQFormAutocomplete({
           // This makes sure the autocomplete options list width is never less than the input field width
           minWidth: `${Number(baseWidth ?? 0) + LISTBOX_PADDING}px !important`,
         },
+        '& .MuiInputBase-root.Mui-error:before': {
+          borderColor: 'var(--color-textWarningYellow)',
+        },
+        '& .Mui-error': {
+          color: 'var(--color-textWarningYellow)',
+        },
+        '& .MuiFormHelperText-root > .MuiSvgIcon-colorError': {
+          color: 'var(--color-textWarningYellow)',
+        },
+        '& .MuiFormHelperText-root > .MuiSvgIcon-root:not(.MuiSvgIcon-colorDisabled):not(.MuiSvgIcon-colorError)':
+          {
+            color: 'var(--color-textSuccessGreen)',
+          },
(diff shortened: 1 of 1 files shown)


20BBrown14:
So do we plan on making updates to other components? I'm just a little concerned about changing CSS for some fields and not others. If we eventually want changes for all components we could update the theme.

I guess I'm basically just asking what the main driver behind this ticket is,.

laurelbean:
> I guess I'm basically just asking what the main driver behind this ticket is,.

This and most of the SSR tickets right now are to get all the components in line with the SC3 Design Library. Carlos and I have been splitting it out into components to make sure all the aspects are checked. A lot of the colors would be the same across components so looking into updating the theme is probably a good idea.

samasastudio:
> A lot of the colors would be the same across components so looking into updating the theme is probably a good idea.

Would it be better to pass theme elements instead of making reference to the `root.css` file variables, and then hold off on merging this until the theme in `scplus-shared-components` is updated overall?  Once we can test and confirm that the styles have applied correctly, we could then merge?

Issue #448, opened 2021-09-21
SQFormScrollableCard -- if `isHeaderDisabled` is set to true, `title` should not be required


lucas-homer:
At the moment, `title` prop is required, but if `isHeaderDisabled` is set to `true` we don't do anything with it, so let's not make it required.

https://www.loom.com/share/8b30454e4beb4555945ab083482ce29d

aharpt:
@laurelbean @20BBrown14 @hvilander Took a look at this issue, and it appears that this work is already done: https://www.loom.com/share/f2fe4aad3cae4fed9523d1d8d02f6a47.

laurelbean:
So from what I'm seeing now, `title` is not required. However, I think want it to be required unless you choose `isHeaderDisabled`. So I think we could still improve things by implementing something similar to what you did with the Accordion component.
https://bitbucket.org/SelectQuote/scplus-shared-components/pull-requests/282#Lsrc/components/Accordion/Accordion.tsxT5

aharpt:
@laurelbean Sounds good, I'll take a look at implementing that.

PR #846, opened 2023-02-15
refactor: 💡 Title prop is conditionally required

aharpt:
https://www.loom.com/share/0a2063f2cb46419a804c87fb90bbbb93

✅ Closes: #448


src/components/SQFormScrollableCard/SQFormScrollableCard.tsx
```diff
@@ -56,15 +70,11 @@ export type SQFormScrollableCardProps<Values extends FormikValues> = {
...
   SubHeaderComponent?: React.ReactNode;
-  /** The Title for the Header component */
-  title?: string;
   /**
    * Yup validation schema shape
    * https://jaredpalmer.com/formik/docs/guides/validation#validationschema
    * */
   validationSchema?: AnyObjectSchema;
-  /** Boolean used to determine if title/header is enabled or disabled */
-  isHeaderDisabled?: boolean;
```
hvilander, line 67: Checking my understanding, does this make isHeaderDisabled a required prop?

hvilander:
Another question, maybe a naive one.
Why not have title as optional, if title is not provided disable header? (I am probably missing some use case)

src/components/SQFormScrollableCard/SQFormScrollableCard.tsx
aharpt: @hvilander That's a great callout and totally my bad!  Thanks for calling this out.

aharpt:
> Another question, maybe a naive one. Why not have title as optional, if title is not provided disable header? (I am probably missing some use case)

@hvilander That's a totally fair question.  To be honest, not really seeing a reason this couldn't work...  I know Branden created an issue to make the title prop either required or have a default value.  @20BBrown14 I cannot find a reason why Harold's suggestion wouldn't work.  Do you know of any use-cases that would go against this?

20BBrown14:
> > Another question, maybe a naive one. Why not have title as optional, if title is not provided disable header? (I am probably missing some use case)
> 
> @hvilander That's a totally fair question. To be honest, not really seeing a reason this couldn't work... I know Branden created an issue to make the title prop either required or have a default value. @20BBrown14 I cannot find a reason why Harold's suggestion wouldn't work. Do you know of any use-cases that would go against this?

Makes sense to me. I could _maybe_ see a case where someone doesn't want a title to display but they want the space in the UI to still be as if the title was there. However, in that case, I think they could provide a `' '` to not have a title but still keep the space in the UI? But that's probably an edge case.

Do we think removing the `isHeaderDisabled` prop and changing logic based on the value of `title` is a breaking change?
I'm thinking about it and maybe? The only way it'd be a breaking change is if `title` is a truthy value, and they're still supplying `isHeaderDisabled={true}` which is supposed to be an unsupported config but our type definitions, as of now, allows it. and if we changed it, on the next deploy this component would be providing a title/header in the UI when it wasn't supposed to be.

Thoughts?

hvilander:
Technically it _might_ be a breaking change, as that would be unsupported I am fine considering it a non-breaking change.

20BBrown14:
> Technically it _might_ be a breaking change, as that case is clearly unsupported I am fine considering it a non-breaking change.

I can live with that. And to be fair <sup>to be fair</sup> nobody is on the latest SQForm version anyway, and by the time they are this is an issue they'd fix during migration and not in prod.

laurelbean:
I think conditionally rendering title only if it exists would be cleaner. Having too many props can get confusing. If we go that route, please add to the `title` prop comment about the conditional render.

stories/__tests__/SQFormScrollableCard.stories.test.tsx
```diff
@@ -137,14 +137,6 @@ describe('SQFormScrollableCard', () => {
       expect(screen.getByText(/Testing Title/i)).toBeInTheDocument();
     });
 
-    it('Should not render specified header title when disabled', () => {
```
laurelbean, line 140: Instead of removing this, can you modify it to use `title` instead?

src/components/SQFormScrollableCard/SQFormScrollableCard.tsx
```diff
@@ -46,6 +46,8 @@ export type SQFormScrollableCardProps<Values extends FormikValues> = {
     values: Values,
     formikHelpers: FormikHelpers<Values>
   ) => void | Promise<unknown>;
+  /** The Title for the Header component. If you do not want the header to show, leave undefined or blank (e.g. '') */
+  title?: string;
```
laurelbean, line 50: I'd like to keep the props alphabetical.

aharpt:
@hvilander @20BBrown14 @laurelbean I have removed the `isHeaderDisabled` prop and believe everything discussed is complete.  Let me know if there are any remaining issues.

stories/__tests__/SQFormScrollableCard.stories.test.tsx
```diff
@@ -138,11 +138,9 @@ describe('SQFormScrollableCard', () => {
     });
 
     it('Should not render specified header title when disabled', () => {
-      render(<Default title="Testing Title" isHeaderDisabled={true} />);
+      render(<Default title={undefined} />);
```
laurelbean, line 141: I think instead of explicitly passing in `undefined` most people will just not pass it in, so can we leave it out or make it empty string like the comment says?

aharpt: Agreed.  I believe that I do need to pass the `title` here in order to override the default value given to the `Default` component, but I agree that passing in an empty string makes more sense.  Will update this.

stories/__tests__/SQFormScrollableCard.stories.test.tsx
```diff
@@ -138,11 +138,9 @@ describe('SQFormScrollableCard', () => {
...
 
-      expect(() => screen.getByText(/Testing Title/i)).toThrow(
-        'Unable to find an element'
-      );
+      expect(screen.queryByTestId(/title/i)).toBeNull();
```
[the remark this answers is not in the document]
aharpt: Makes, sense.  I will update this.

laurelbean (approved):

hvilander (approved):

20BBrown14 (approved):

PR #856, opened 2023-02-23
SSR-163: SQForms: Masked Read Only Field - Update component

asim-sqfe:
Update the current implementation of the Masked Read Only Field with added specifications. Here is the current implementation of this SQForm: https://master--5f4431386ea00a00220d495c.chromatic.com/?path=/story/components-sqformmaskedreadonlyfield--default  (this is an example link)

Design reference: https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=1518%3A11678 
Typography: https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=340%3A9946 
Color: https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=340%3A9646 

Default:

Label:  Text Style: Label 12 px Weight: 400

Loom Demo: [here](https://www.loom.com/share/309a383180244435b6b4e3e2327c9f2el)

### DIFF
--- a/src/components/fields/SQFormMasked/SQFormMaskedReadOnlyField.tsx
+++ b/src/components/fields/SQFormMasked/SQFormMaskedReadOnlyField.tsx
@@ -54,6 +54,9 @@ function SQFormMaskedReadOnlyField({
         ...InputProps,
         placeholder,
         inputComponent: TextFieldMask,
+        sx: {
+          fontWeight: 600,
+        },
       }}
       inputProps={{
         ...inputProps,


malikfarhan183:
@asim-sqfe I think we need to use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) so that it can automatically determine a semantic version bump.

laurelbean:
> @asim-sqfe I think we need to use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) so that it can automatically determine a semantic version bump.

Yes, you can use `npm run commit` to help you create the commit.

malikfarhan183 (approved):

laurelbean (approved):

Issue #861, opened 2023-03-01
Disabled dropdown causes layout shifts


hvilander:
<img width="334" alt="image" src="https://user-images.githubusercontent.com/2816261/222165082-f8e5f8b5-3f43-4a78-9df0-70bb442416db.png">
vs
<img width="345" alt="Screenshot 2023-03-01 at 08 12 07" src="https://user-images.githubusercontent.com/2816261/222165120-80f030ff-43d7-42a5-9c50-f375bce05701.png">

Ideally we know how big the helper text is and should add that amount of padding when the field is disabled. 

I believe this would make for nicer forms when we are dynamically disabling dropdown fields.

PR #870, opened 2023-03-20
fix: 🐛 Fix dropdown layout shift on disable

aharpt:
https://www.loom.com/share/1a01edc7694d44fbbeb10a8ef95afa6d

✅ Closes: #861

### DIFF
--- a/src/components/fields/SQFormDropdown/SQFormDropdown.tsx
+++ b/src/components/fields/SQFormDropdown/SQFormDropdown.tsx
@@ -125,7 +125,7 @@ function SQFormDropdown({
   };
 
   return (
-    <Grid item sm={size}>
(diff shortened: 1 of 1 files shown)


laurelbean:
I agree, I think always having that space held for the Helper text will make things easier in the long run.

aharpt:
> I would like to make a suggestion. Instead of using a seemingly random bottom padding, we could just utilize an empty `FormHelperText`
> 
> ```
> <FormHelperText>{!isDisabled && displayHelperText ? HelperTextComponent : null}</FormHelperText>
> ```
> 
> In my opinion this is more robust and will continue to work if the FormHelperText component, or it's contents, are changed in the future.
> 
> Thoughts?

@20BBrown14 Agreed that is a better way to do this.  Have updated the PR accordingly.

malikfarhan183 (approved):

hvilander (approved):

PR #875, opened 2023-03-28
874 sqforms readonlyfield update component

Mudassar-Raza:
Update the current implementation of the ReadOnlyField with added specifications. Here is the current implementation of this SQForm: [Webpack App](https://master--5f4431386ea00a00220d495c.chromatic.com/?path=/story/components-sqformreadonlyfield--default)    (this is an example link)

Design reference: [SC3 DESIGN LIBRARY](https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=1518%3A11678) 
Typography: [SC3 DESIGN LIBRARY](https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=340%3A9946) 
Color: [SC3 DESIGN LIBRARY](https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-SYSTEM?node-id=340%3A9646) 

Default:
Label:  Text Style: Label 12 px Weight: 400

Note: Underline in the read-only needs to be removed

Loom Link: [Loom](https://www.loom.com/share/1aa834f6e013457393ec0a24c623feb0)

closes: #874

### DIFF
--- a/src/components/fields/SQFormReadOnlyField/SQFormReadOnlyField.tsx
+++ b/src/components/fields/SQFormReadOnlyField/SQFormReadOnlyField.tsx
@@ -11,11 +11,12 @@ export type SQFormReadOnlyFieldProps = BaseFieldProps & {
   inputProps?: TextFieldProps['inputProps'];
 };
 
+const readOnlyInputProps = {
+  disableUnderline: true,
+  readOnly: true,
+};
+
 const styles = {
-  '& .MuiInput-root:before, & .MuiInput-root:after, & .MuiInput-underline:hover:not(.Mui-disabled):not(.Mui-multiline):before':
-    {
-      borderBottom: '0px',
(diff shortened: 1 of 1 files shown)


laurelbean (approved):

aharpt (approved):

malikfarhan183 (approved):

hvilander:
In your loom it looks like the label briefly uses a teal styling like it is actionable. Is this correct? Is this how it should work? Looking at the Figma it looks like the value could be clickable to copy, but the label should never look actionable.

Mudassar-Raza:
Yes, this seems right. The label should not look actionable.

Issue #876, opened 2023-04-07
SQFormDatePicker Tests Update


aharpt:

aharpt:
@laurelbean I have made a few adjustments, but just curious if you had anything specific in mind in terms of things you wanted to see changed on the DatePicker tests?

laurelbean:
There were a few test cases I couldn't get to work because for some reason Testing Library was interpreting the text box as a button. See my comment here: https://github.com/SelectQuoteLabs/SQForm/issues/265#issuecomment-880045256

PR #879, opened 2023-04-12
test: 💍 Update DatePicker tests

aharpt:
Changes in this PR are mainly due to VSCode react-testing-library intellisense suggestions.  There is still an ongoing "issue" with the datepicker tests where the input and calendar button seem to be treated as one input with the role of `textbox`.  Due to the fact that this seems to be controlled by MUI, it may not be doable to alter this on our end.  See discussion on the issue if interested https://github.com/SelectQuoteLabs/SQForm/issues/876.

Screen Shot 2023-04-12 at 10 50 19 AM

✅ Closes: #876

### DIFF
--- a/stories/__tests__/SQFormDatePicker.stories.test.tsx
+++ b/stories/__tests__/SQFormDatePicker.stories.test.tsx
@@ -20,7 +20,7 @@ const renderDatePicker = (
   >
 ) => {
   return render(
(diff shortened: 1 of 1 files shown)


laurelbean:
It would be interesting to see if using a plain MUI date component also has the same issue with Testing Library.

laurelbean (approved):

PR #884, opened 2023-04-28
feat: 🎸 Add prop for SQForm fields to allow 'validation' as inf

malikfarhan183:
**Jira Description:**
We need to implement design and business requirements that require informational text to a SQForm field, similar to Validation.

Any place there is validation we need the ability for informational text, including to SQFormMultiselect, since there is currently no mechanism to override what the form engine is doing for validation text generation.

This is needed for the following input types:

image
 

Figma Fields examples: [SC3 DESIGN LIBRARY](https://www.figma.com/file/DsJr2cFZqMYkOX6VjcuhqW/SC3-DESIGN-LIBRARY?node-id=1518%3A11678&t=IWwH5R2yzR5UNulh-0)

**Loom Video:** https://www.loom.com/share/1f883c4f27e342099f40eefab4fbb130

**Other Notes :** I have added the “informationalText” prop for both SQFormDropdown and SQFormTextField. Now if we will send this prop to these fields we will get the helper text under the field. let me know if we need to update the component.

**Design Review :** Approved by UI/UX

✅ Closes: SSR-14

### DIFF
--- a/src/components/fields/SQFormDropdown/SQFormDropdown.tsx
+++ b/src/components/fields/SQFormDropdown/SQFormDropdown.tsx
@@ -23,6 +23,8 @@ export type SQFormDropdownProps = BaseFieldProps & {
   children: SQFormOption[];
   /** Whether to display empty option - - in options */
   displayEmpty?: boolean;
+  /** Used to display informational Text under the input field same like warning text */
+  informationalText?: string;
   /** Disabled property to disable the input if true */
   isDisabled?: boolean;
   /** Whether or not to show the helper text */
@@ -62,6 +64,7 @@ function SQFormDropdown({
   onChange,
   size = 'auto',
   muiFieldProps = {},
+  informationalText = '',
 }: SQFormDropdownProps): React.ReactElement {
   const {
     formikField: {field},
@@ -71,6 +74,7 @@ function SQFormDropdown({
     name,
     onBlur,
     onChange,
+    informationalText,
   });
   const labelID = label.toLowerCase();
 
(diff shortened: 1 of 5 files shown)


stories/SQFormDropdown.stories.tsx
```diff
@@ -70,7 +70,11 @@ const Template: DropdownStoryType = (args) => {
         validationSchema={schema}
         {...sqFormProps}
       >
-        <SQFormDropdownComponent {...dropdownProps} size={getSizeProp(size)} />
+        <SQFormDropdownComponent
```
aharpt, line 73: @malikfarhan183 Looks really good.  Instead of replacing the current story, I think we should add a new story with the helper text.  That way, we would have an example of both a dropdown with the informational text, and an example without it.

stories/SQFormTextField.stories.tsx
```diff
@@ -45,7 +45,11 @@ const Template: SQFormTextFieldStoryType = (args) => {
       initialValues={{[defaultArgs.name]: ''}}
       {...SQFormProps}
     >
-      <SQFormTextFieldComponent {...rest} size={getSizeProp(size)} />
+      <SQFormTextFieldComponent
```
aharpt, line 48: Same as above.  I think adding a new story instead of changing the existing one might be preferable.

aharpt (approved):

stories/SQFormDropdown.stories.tsx
malikfarhan183: @aharpt Thanks for your great suggestion. I have created a new story for this new helper text prop. 
image

stories/SQFormTextField.stories.tsx
malikfarhan183: @aharpt 
image

aharpt:
Looks good. Thanks @malikfarhan183!

laurelbean (approved):

src/components/fields/SQFormDropdown/SQFormDropdown.tsx
```diff
@@ -162,7 +166,18 @@ function SQFormDropdown({
...
         {!isDisabled && displayHelperText && (
-          <FormHelperText>{HelperTextComponent}</FormHelperText>
+          <FormHelperText
+            sx={{
+              '&.MuiFormHelperText-root': {
+                color: 'var(--color-slate)',
+              },
+              '&.Mui-focused': {
+                color: 'var(--color-textInfoBlue)',
+              },
```
hvilander, line 176: why does the text field need the orange color added to error and dropdown does not?

PR #901, opened 2023-06-27
feat: 🎸 SQFormTransferTool

hvilander:
Creates a new component SQFormTransferTool. It is an opinionated, data driven modal to aid in call transfers

loom with some extra explanation: https://www.loom.com/share/58f48fff0e304f789506373aa752fb0b?sid=290e878f-3823-4ff3-976c-bc5a5861d406

### DIFF
--- a/src/components/SQFormTransferTool/SQFormTransferProductPanels.tsx
+++ b/src/components/SQFormTransferTool/SQFormTransferProductPanels.tsx
@@ -10,7 +10,7 @@ export type SQFormTransferProductPanelsProps = {
 
 function getPanel(
   transferProducts: Array<TransferProduct>
-): AccordionPanelType[] | null {
+): AccordionPanelType[] {
   if (!transferProducts?.length) {
     return [];
   }
--- a/src/components/SQFormTransferTool/SQFormTransferProductPanels.tsx
+++ b/src/components/SQFormTransferTool/SQFormTransferProductPanels.tsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import {Box} from '@mui/material';
+import {Accordion} from 'scplus-shared-components';
+import type {AccordionPanelType} from 'scplus-shared-components';
+import type {TransferProduct} from './types';
+
+export type SQFormTransferProductPanelsProps = {
+  transferProducts: TransferProduct[];
+};
+
+function getPanel(
+  transferProducts: Array<TransferProduct>
+): AccordionPanelType[] | null {
+  if (!transferProducts?.length) {
+    return [];
+  }
+
+  return transferProducts.map(({productID, productDisplayName, enabled}) => {
+    return {
+      body: <>TO BE IMPLEMENTED SC3-1809 (accordions) and SC3-1810 (steps)</>,
+      name: `${productID}`,
+      title: productDisplayName,
+      isDisabled: !enabled,
+    };
+  });
+}
+
+export default function SQFormTransferProductPanels({
+  transferProducts,
+}: SQFormTransferProductPanelsProps): React.ReactElement | null {
+  const panels = getPanel(transferProducts) ?? [];
+  if (!panels.length) {
+    return null;
+    // TODO ??
+    // Is this valid? DO we want some sort of loading indicator?
+    // or will be not opening the modal until the data has loaded?
+  }
+
+  return (
+    <Box
+      sx={{
+        minWidth: '600px',
+        padding: '16px',
+        paddingBottom: '0px',
+      }}
+    >
+      <Accordion accordionPanels={panels} />
+    </Box>
+  );
+}
--- a/src/components/SQFormTransferTool/SQFormTransferProductPanels.tsx
+++ b/src/components/SQFormTransferTool/SQFormTransferProductPanels.tsx
@@ -8,9 +8,7 @@ export type SQFormTransferProductPanelsProps = {
   transferProducts: TransferProduct[];
 };
 
-function getPanel(
-  transferProducts: Array<TransferProduct>
-): AccordionPanelType[] {
+function getPanel(transferProducts: TransferProduct[]): AccordionPanelType[] {
   if (!transferProducts?.length) {
     return [];
   }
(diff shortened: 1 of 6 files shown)


hvilander:
I can see the build error and will correct it first thing tomorrow.

src/components/SQFormTransferTool/SQFormTransferProductPanels.tsx
```diff
@@ -0,0 +1,50 @@
...
+import type {AccordionPanelType} from 'scplus-shared-components';
+import type {TransferProduct} from './types';
+
+export type SQFormTransferProductPanelsProps = {
+  transferProducts: TransferProduct[];
+};
+
+function getPanel(
+  transferProducts: Array<TransferProduct>
+): AccordionPanelType[] | null {
```
20BBrown14, line 13: nit: It doesn't look like this function can actually return null.

20BBrown14 (approved):

20BBrown14:
Is it possible to remove the footer while loading?

20BBrown14:
I think I'd like to see a bit more padding between the panels inside the modal

src/components/SQFormTransferTool/SQFormTransferProductPanels.tsx
```diff
@@ -0,0 +1,50 @@
...
+      name: `${productID}`,
+      title: productDisplayName,
+      isDisabled: !enabled,
+    };
+  });
+}
+
+export default function SQFormTransferProductPanels({
+  transferProducts,
+}: SQFormTransferProductPanelsProps): React.ReactElement | null {
```
20BBrown14, line 30: Don't quote me, but I think `React.ReactElement` includes `null` in its type

src/components/SQFormTransferTool/SQFormTransferProductPanels.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const panels = getPanel(transferProducts) ?? [];
+  if (!panels.length) {
+    return null;
```
20BBrown14, line 33: Is this possible? An edge case? Or an indication of an error? If this is unexpected but should still be handled maybe we can throw a `console.warn` in here?

src/components/SQFormTransferTool/SQFormTransferTool.tsx
```diff
@@ -0,0 +1,54 @@
...
+}: SQFormTransferToolProps<Values>): React.ReactElement {
+  // TODO Validation: this will be internal and the consumer need not worry about it. It should be part of SC3-1810 processing each step
+  // TODO Initial Values: like validation this will be hidden from the consumer as they should not need it. SC3-1810 steps
+  // TODO Transfer: transfers will be handled via consumer supplied callback. SC3-1811
+
+  return (
+    <SQFormDialog
+      isOpen={isOpen}
+      onSave={onSave}
+      onClose={noop}
```
20BBrown14, line 39: is this noop intended to be permanent here?

src/components/SQFormTransferTool/SQFormTransferTool.tsx
```diff
@@ -0,0 +1,54 @@
...
+      title="Standard Transfer Modal"
+      initialValues={initialValues}
+      shouldDisplaySaveButton={false}
+      cancelButtonText="save and close"
+      showSecondaryButton={!isLoading}
```
20BBrown14, line 44: Should we disable the button on loading instead of removing it entirely?

src/components/SQFormTransferTool/SQFormTransferTool.tsx
hvilander: Probably....
Until something requires consuming the MUI 5 dialog directly I think I would prefer to use SQForm Dialog. Which requires an onClose Prop. The only use cases for exiting the dialog will require 'saving'. 

That being said, I am definitely using the button's weirdly here. My 'save' action is on the secondary to match the style of a comp. 

So as I get a bit further I am waiting to see if using SQFormDialog will really work for this or not. 
Since it may not, I am kinda playing fast and loose with things.

All that being said I have talked myself into adding a TODO comment here to re-evaluate after I get a bit further into the requirements.

src/components/SQFormTransferTool/SQFormTransferTool.tsx
hvilander: 1. there is not a disable on the secondary button right now for SQFormDialog
2. see the above comment about my shakey status of its usage.

src/components/SQFormTransferTool/SQFormTransferTool.tsx
laurelbean: If we're using the buttons backwards, that seems like something we would want to take a look at. I think it's going to heavily depend on how it's consumed once it gets a little further along.

hvilander:
> I think I'd like to see a bit more padding between the panels inside the modal

vertical padding? This is what the shared components provide. This would be a change to ssc.

src/components/SQFormTransferTool/SQFormTransferProductPanels.tsx
hvilander: The TODO comment below this, is to help circle back to this very question. The requirements have not yet defined if we really want a loading indicator or what do to if we fail to load. 

I will resolve this once I can get better requirements.

src/components/SQFormTransferTool/SQFormTransferTool.tsx
hvilander: Like above there are TODO's all over this given the current requirements. There is one on line 34. I agree this will depend on how its consumed, hence the todos. I am not a huge fan of littering the component with them, but I don't have a ton of choice yet.

src/components/SQFormTransferTool/SQFormTransferTool.tsx
```diff
@@ -0,0 +1,55 @@
...
+  // TODO Initial Values: like validation this will be hidden from the consumer as they should not need it. SC3-1810 steps
+  // TODO Transfer: transfers will be handled via consumer supplied callback. SC3-1811
+  // TODO the onSave, onClose functions seem backward, after adding some more of the required functionality re-evaluate this
+
+  return (
+    <SQFormDialog
+      isOpen={isOpen}
+      onSave={onSave}
+      onClose={noop}
+      title="Standard Transfer Modal"
```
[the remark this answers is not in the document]
hvilander: adding this.

laurelbean (approved):
A great start! Thanks for the TODOs to help keep it organized.

20BBrown14:
> > I think I'd like to see a bit more padding between the panels inside the modal
> 
> vertical padding? This is what the shared components provide. This would be a change to ssc.

oof

PR #904, opened 2023-07-10
feat: 🎸 Virtualized Multiselect

20BBrown14:
Added the ability for the SQFormMultiselect to be virtualized. This GREATLY improves performance for implementations with a lot of options. Kept the virtualization optional to avoid breaking changes within v9.

This will look to release v9.8.0, I'm fairly certain. I will then merge this feature into master as well.

Loom: https://www.loom.com/share/20e25bd99e0148d88a4c5bcdd545331b

### DIFF
--- a/src/components/SQForm/SQFormMultiSelect.tsx
+++ b/src/components/SQForm/SQFormMultiSelect.tsx
@@ -8,11 +8,12 @@ import {
   FormHelperText,
   Grid,
   Checkbox,
-  ListItemText,
   Tooltip,
   makeStyles,
+  ListItemText,
 } from '@material-ui/core';
 import {useFormikContext} from 'formik';
+import {VariableSizeList} from 'react-window';
 import {EMPTY_LABEL} from '../../utils/constants';
 import {useForm} from './useForm';
 import {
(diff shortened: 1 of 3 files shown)


src/components/SQForm/SQFormMultiSelect.tsx
```diff
@@ -21,18 +22,16 @@ import {
   getUndefinedValueWarning,
 } from '../../utils/consoleWarnings';
 import type {TooltipProps, SelectProps} from '@material-ui/core';
+import type {ListChildComponentProps} from 'react-window';
 import type {BaseFieldProps, SQFormOption} from '../../types';
 
-export type SQFormMultiSelectProps = BaseFieldProps & {
+export type SQFormMultiSelectProps<TVirtualized = boolean> = BaseFieldProps & {
```
laurelbean, line 28: Do we need this generic type since it seems like the prop should always be a boolean?

src/components/SQForm/SQFormMultiSelect.tsx
```diff
@@ -211,6 +236,75 @@ function SQFormMultiSelect({
...
+      <MenuItem
+        key={`${name}_${option.value}`}
+        value={option.value}
+        style={{
+          ...style,
+          paddingTop: 0,
+          paddingBottom: 0,
+        }}
+        onClick={(event) => {
+          console.log({event});
```
laurelbean, line 252: Leftover log?

src/components/SQForm/SQFormMultiSelect.tsx
20BBrown14: Ya, it's a boolean. But I'm using the value of the boolean to determine what type of event will be sent to the onChange handler. It's mainly to keep TS happy. Consumers should never have to supply the generic.

laurelbean (approved):

hvilander (approved):

PR #909, opened 2023-07-13
feat: 🎸 Deal with validation - SC3-1810 Part 2

hvilander:
Given that every field is dropdown with provided options, every field is already valid

So I have removed todo's regarding validation as I feel this is the best path. 
I explain my thinking in the [Loom](https://www.loom.com/share/c1a9de55529f43a5b3a6ac77462d5ab4?sid=98c05987-82b0-4c8a-b5ee-eec50f513d81)

I also make a case for the second storybook script launcher I have added in package.json

I have also done a touch of clean up to remove some console errors for duplicate keys.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "nuke-everything": "rimraf ./dist && rimraf ./node_modules && rm -f package-lock.json",
     "preview": "npm run build && npm run build-storybook && cp -r .out/ dist/ && vite preview",
     "storybook": "start-storybook -p 6006",
+    "storybook.dev": "ESLINT_NO_DEV_ERRORS=true npm run storybook",
     "test": "jest --watch",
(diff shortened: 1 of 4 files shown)


20BBrown14 (approved):

package.json
```diff
@@ -32,6 +32,7 @@
     "nuke-everything": "rimraf ./dist && rimraf ./node_modules && rm -f package-lock.json",
     "preview": "npm run build && npm run build-storybook && cp -r .out/ dist/ && vite preview",
     "storybook": "start-storybook -p 6006",
+    "storybook.dev": "ESLINT_NO_DEV_ERRORS=true npm run storybook",
```
laurelbean, line 35: Nice, I'm definitely using this next time.

laurelbean (approved):

PR #910, opened 2023-07-17
SC3-1811 transfer step

hvilander:
The loom tells almost all https://www.loom.com/share/3bbf5a8effff4f4a98d5ad6cea23ba7c?sid=800f3080-1a3e-41bd-a1ea-20e1c7365337

Additionally: 
1. the spacing issue I mention is resolved on epic/version-11, it is an issue with dropdown that is fixed.
2. The data contract on transfer is a bit shaky. Please review it a bit extra. Since this is going to be just between the consumer and this component we have some flex on how to make it work. There was not a strict requirement provided.

Here is the screen shot of the data being logged out to the console on transfer click: 
image

and on save / close (more work for this on the next PR)
image

### DIFF
--- a/src/components/SQFormTransferTool/AccordionBody.tsx
+++ b/src/components/SQFormTransferTool/AccordionBody.tsx
@@ -6,27 +6,81 @@ import {
   TextButton,
 } from 'scplus-shared-components';
 import Step from './Step';
-import type {OnTransfer, TransferProduct} from './types';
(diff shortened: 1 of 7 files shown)


laurelbean (approved):

20BBrown14 (approved):

laurelbean:
I think what you have for the call back data is good. It has everything that I know of that would be needed.

